PR #43, opened 2017-12-06
Fixes issue #42

### LINKED ISSUE #42: register signal event not supported?
Hi all,
I have troubles registering a change detection event. The following code 

```import nidaqmx.task
import nidaqmx.system

def getDaqDcPowerVoltage(task_handle=0, every_n_samples_event_type=0, number_of_samples=0, callback_data=0):
   result = None
   print('this is the digital change detection callback function.')
   print(callback_data)                                            
   return 0                                                        
   
DaqMxSystem = nidaqmx.system.System.local()                         
task = nidaqmx.Task()                                               
task.di_channels.add_di_chan('PXI1Slot2/port0/line0')               
task.timing.cfg_samp_clk_timing(1000)                               
task.Buffer = 10                                                    
task.output_buf_size = 10                                           
task.read_all_avail_samp = True                                     
task.register_signal_event(nidaqmx.constants.Signal.CHANGE_DETECTION_EVENT, getDaqDcPowerVoltage(callback_data = 0))
task.start()
```

ends in this error:

     cfunc = lib_importer.daqlib.DAQmxRegisterSignalEvent
     AttributeError: 'DaqLibImporter' object has no attribute 'daqlib'

I saw, that the DaqLibImporter class in _lib.py is truely missing the daqlib property. Is this something that is not supported in the current nidaqmx-python implementation or am I doing something wrong?

BR Andreas

hillbilly3456avk:
instead of using daqlib property for DAQmxRegisterSignalEvent use windll. Fixes #42 

Sorry for the unfinished unittest - i didn't really copy how you reference the DAQmx Devices, but at least it should point the line ...

### DIFF
--- a/nidaqmx/tests/test_events.py
+++ b/nidaqmx/tests/test_events.py
@@ -21,7 +21,7 @@ class TestEvents(object):
 
         samples_chunk = 100
         sample_rate = 5000
-
+    
         with nidaqmx.Task() as task:
             task.ai_channels.add_ai_voltage_chan(
                 x_series_device.ai_physical_chans[0].name)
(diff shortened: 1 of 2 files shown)


nidaqmx/tests/test_events.py
```diff
@@ -76,3 +76,66 @@ def callback(task_handle, every_n_samples_event_type,
             task.stop()
 
             assert non_local_var['samples_read'] == 0
+    
+    @pytest.mark.parametrize('seed', [generate_random_seed()])
+    def test_signal_event(self, x_series_device, seed):
```
brianlee1200: Since this unit test doesn't actually test the feature in question, I would revert this for now. I can merge your changes to correct the library used for the `DAQmxRegisterSignalEvent` function, though, thanks!

hillbilly3456avk: thats fine for me - actually I was just too lazy to do the Unittest right.

rares-pop:
This got fixed in:
https://github.com/ni/nidaqmx-python/commit/db9cf5b3b5a12393548e2a619dbdbb265e5fb473

PR #45, opened 2017-12-12
Bug in InStream.readinto()

samuelgarcia:
Ensure number_of_samples_per_channel is int (and not numpy.int)
Otherwise ctype is buggy in _read_raw.

### DIFF
--- a/nidaqmx/_task_modules/in_stream.py
+++ b/nidaqmx/_task_modules/in_stream.py
@@ -2001,7 +2001,7 @@ class InStream(object):
                 number_of_channels * channels_to_read.ai_raw_samp_size / 8))
 
         samples_read, _ = _read_raw(
-            self._handle, numpy_array, number_of_samples_per_channel,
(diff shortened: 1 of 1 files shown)


campagnola:
Here's an exception I encountered that would be fixed by this PR:

```
  File "our_code\ni_daqmx.py", line 232, in run
    nb_sample = aitask.in_stream.readinto(raw_data_flat)
  File "C:\Anaconda3\lib\site-packages\nidaqmx\_task_modules\in_stream.py", line 1992, in readinto
    self.timeout)
  File "C:\Anaconda3\lib\site-packages\nidaqmx\_task_modules\read_functions.py", line 583, in _read_raw
    ctypes.byref(number_of_bytes_per_sample), None)
ctypes.ArgumentError: argument 2: <class 'TypeError'>: wrong type
```

rares-pop:
Thank you for the contribution!

samuelgarcia:
cool!!

PR #75, opened 2019-02-12
Fix for Linux where DAQmxGetSysNIDAQUpdateVersion is not available

jsalort:
I have started using this module on Linux (with AI channels), and this is the only change that I add to do to make it work. Apparently, DAQmx on Linux does not have `DAQmxGetSysNIDAQUpdateVersion`. It seems to be enough to just fallback to `DAQmxGetSysNIDAQMinorVersion` in that case (although I am not sure that it means the same thing exactly).

### DIFF
--- a/nidaqmx/system/system.py
+++ b/nidaqmx/system/system.py
@@ -150,7 +150,7 @@ class System(object):
         try:
             cfunc = lib_importer.windll.DAQmxGetSysNIDAQUpdateVersion
         except DaqFunctionNotSupportedError:
-            cfunc = 0
+            return 0
         if cfunc.argtypes is None:
             with cfunc.arglock:
                 if cfunc.argtypes is None:
--- a/nidaqmx/system/system.py
+++ b/nidaqmx/system/system.py
@@ -150,7 +150,7 @@ class System(object):
         try:
             cfunc = lib_importer.windll.DAQmxGetSysNIDAQUpdateVersion
         except DaqFunctionNotSupportedError:
-            cfunc = lib_importer.windll.DAQmxGetSysNIDAQMinorVersion
+            cfunc = 0
         if cfunc.argtypes is None:
             with cfunc.arglock:
                 if cfunc.argtypes is None:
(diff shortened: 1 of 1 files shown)


rares-pop:
@jsalort - what version of DAQmx did you use?

jsalort:
I am using NI-DAQmx 8.0.2f0 on OpenSUSE 11.4 on a PXI-1031 with PXI-8102 chassis.

rares-pop:
Oh, that's fairly old.

nidaqmx/system/system.py
```diff
@@ -147,7 +147,10 @@ def _update_version(self):
         """
         val = ctypes.c_uint()
 
-        cfunc = lib_importer.windll.DAQmxGetSysNIDAQUpdateVersion
+        try:
+            cfunc = lib_importer.windll.DAQmxGetSysNIDAQUpdateVersion
+        except DaqFunctionNotSupportedError:
+            cfunc = lib_importer.windll.DAQmxGetSysNIDAQMinorVersion
```
rares-pop, line 153: can we just return 0 here?

it's strange to have the update part of the version in sync with the minor, i.e. 8.0.0, 8.1.1, 8.2.2 (without actual 8.2.0 or 1). 

I'm thinking if there's no update, either have the DriverVersion return major.minor, or major.minor.0, and I guess it's safer for comparisons and so on to use a 0 there.

jsalort: Agreed. I have updated to PR.

nidaqmx/system/system.py
```diff
@@ -147,7 +147,10 @@ def _update_version(self):
         """
         val = ctypes.c_uint()
 
-        cfunc = lib_importer.windll.DAQmxGetSysNIDAQUpdateVersion
+        try:
+            cfunc = lib_importer.windll.DAQmxGetSysNIDAQUpdateVersion
+        except DaqFunctionNotSupportedError:
+            cfunc = 0
```
rares-pop, line 153: I'd expect this to be:
`return 0`

jsalort: Right. Sorry about that.

rares-pop:
Awesome! Thank you very much for the contribution!

Issue #80, opened 2019-05-13
Tests don't run with pytest 4.0+


charterchap:
For example see test_read_write.py:365
```python    
    fixture_dev = x_series_device()
    max_port_width = max([d.do_port_width for d in fixture_dev.do_ports])

    @pytest.mark.skipif(
        len([d.do_port_width <= 16 for d in x_series_device().do_ports]) < 2,
        reason='task.read() accepts max of 32 bits for digital uint reads.')
```
will give an error: `AttributeError: 'function' object has no attribute 'do_ports'`
See pytest docs: https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly

zhindes:
We recently updated all dependencies to their latest version and resolved issues.

PR #81, opened 2019-05-13
Partial Fix Example for pytest Fixture Direct Calls

charterchap:
See issue #80 This is not a complete fix and I don't actually recommend it gets merged.

### DIFF
--- a/nidaqmx/tests/fixtures.py
+++ b/nidaqmx/tests/fixtures.py
@@ -11,8 +11,7 @@ class NoFixtureDetectedError(Error):
     pass
 
 
-@pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


nidaqmx/tests/fixtures.py
```diff
@@ -23,12 +22,17 @@ def x_series_device():
                 len(device.do_lines) >= 8 and
                 (len(device.di_lines) == len(device.do_lines)) and
                 len(device.ci_physical_chans) >= 4):
+            print(device)
```
rares-pop, line 25: is this a debugging left-over?

zhindes:
This PR has been inactive for a while, closing.

PR #95, opened 2020-05-11
fix: task.__del__ to use _saved_name

rares-pop:
My understanding is that when this is getting garbage collected, the python environment is shutting down and importing the _lib fails when trying to get the 'name' property. Basically the sys.meta_path is empty by the time is trying to load the daqmx dll.

Using the already saved name will prevent loading the dll on __del__.

Signed-off-by: Rares POP <rares.pop@ni.com>

### DIFF
--- a/nidaqmx/task.py
+++ b/nidaqmx/task.py
@@ -102,7 +102,7 @@ class Task(object):
             warnings.warn(
                 'Task of name "{0}" was not explicitly closed before it was '
                 'destructed. Resources on the task device may still be '
-                'reserved.'.format(self.name), DaqResourceWarning)
+                'reserved.'.format(self._saved_name), DaqResourceWarning)
 
     def __enter__(self):
         return self


rares-pop:
@spalatofhi - can you take a look at this?

spalatofhi:
I haven't tested. I think this should fix the issue, because `_saved_name` is set inside `__init__` (indirectly). A better fix, perhaps, would be to avoid the import inside `check_for_error`.

rares-pop:
Strangely enough, the _saved_name is there for 'close()' - but I don't understand yet the reason if any. For __del__ it works just nice.

It would be nice to cache the _lib and use that everywhere, as opposed to load it each time. But that is the pattern that is used right now and I don't want to break it in one instance. Maybe a larger refactor can fix that.

I did test this on Windows 10 64bit (py 3.6.3) and LinuxRT (py3.5), and works as expected.

This sort of thing didn't get triggered by the tests though, due to how pytest works and the fact that this wasn't the only test.

spalatofhi:
Hi,
The import in `check_for_error` seems to be a workaround to avoid a circular import. The "proper" fix is probably to restructure the error modules. The import is still performed exactly once, so the `_lib` is thus always the same instance.

As of python 3.8 there is now a builtin for caching properties:
https://docs.python.org/3/library/functools.html#functools.cached_property
Similar tools for memoization have been around for longer.

Thanks for fixing this.

rares-pop:
@spalatofhi - I did create this - https://github.com/ni/nidaqmx-python/issues/99

I'm part of the new team that is going to own this repository, and until we are figuring out the level of investment (the time we have allocated to work on it), I'm tackling things I believe I can make progress on without a huge ramp-up.

Thank you for the contributions you are making and helping us get this straight!

PR #104, opened 2020-08-29
Fix warning regarding ABC import from collections

### LINKED ISSUE #73: DeprecationWarning in Python 3.7
In 3.7 a deprecation warning which will become an error in 3.8 appears:

    from collections import Sequence
    c:\python37\lib\site-packages\nidaqmx\system\_collections\physical_channel_collection.py:8: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

tirkarthi:
Fixes #73

### DIFF
--- a/nidaqmx/_task_modules/channel_collection.py
+++ b/nidaqmx/_task_modules/channel_collection.py
@@ -5,7 +5,11 @@ from __future__ import unicode_literals
 
 import ctypes
 import six
-from collections import Sequence
+
+if six.PY2:
+    from collections import Sequence
+else:
+    from collectionsa.abc import Sequence
 
 from nidaqmx._lib import lib_importer, ctypes_byte_str
 from nidaqmx._task_modules.channels.channel import Channel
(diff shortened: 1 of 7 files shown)


nidaqmx/_task_modules/channel_collection.py
```diff
@@ -5,7 +5,11 @@
 
 import ctypes
 import six
-from collections import Sequence
+
+if six.PY2:
```
rares-pop, line 9: We are removing support for python 2.7, so please simplify those imports to only:

`from collections.abc import Sequence`

tirkarthi: Thanks, the setup.py still had 2.7 classifier at https://github.com/ni/nidaqmx-python/blob/0058a029cca6e6115bac1a3fcdc0c46c3e5e60e9/setup.py#L64 due to which I added the check. I will use your suggestion

PR #127, opened 2021-08-29
Fix #126 (`Task.write` does not handle certain inputs as expected)

### LINKED ISSUE #126: `Task.write` does not handle certain inputs as expected
Hi!

Running this code, I expect to write the signal to line 2, but instead I get a strange error message about mismatching number of channels (even though there is no mismatch):

```python
import nidaqmx

taskLen = 100
signalLine2 = [True] * taskLen
signals = [signalLine2]

with nidaqmx.Task('mytask') as task:
    task.do_channels.add_do_chan('Dev1/port0/line2')
    task.write(signals, auto_start=True)
    task.wait_until_done()

print('done')
```

```
nidaqmx.errors.DaqError: Write cannot be performed, because the number of channels in the data does not match the number of channels in the task.

When writing, supply data for all channels in the task. Alternatively, modify the task to contain the same number of channels as the data written.

Number of Channels in Task: 1
Number of Channels in Data: 1

Task Name: mytask
```

I'm aware that it's possible to avoid this by simply using `signalLine2` as the input instead of the 2-dimensional `signals` variable, but sometimes you can end up in this situation when creating a varying number of signals in a loop.

----

Similarly, an error is raised when mixing Python lists and numpy arrays in the input:

```python
import nidaqmx
import numpy as np

taskLen = 100
signalLine2 = np.ones(taskLen, dtype=bool)
signalLine3 = np.ones(taskLen, dtype=bool)
signals = [signalLine2, signalLine3]

with nidaqmx.Task('mytask') as task:
    task.do_channels.add_do_chan('Dev1/port0/line2')
    task.do_channels.add_do_chan('Dev1/port0/line3')
    task.write(signals, auto_start=True)
    task.wait_until_done()

print('done')
```

```
nidaqmx.errors.DaqError: Write failed, because this write method only accepts boolean samples when there is one digital line per channel in a task.

Requested sample type: <class 'numpy.ndarray'>

Task Name: mytask
```

stafak:
Fixes #126.

This PR changes the `write` method in the following ways:

- Lists are converted to numpy arrays in order to fix the error that is raised when mixing Python lists and numpy arrays, and also to simplify the code.
- An extra conditional fixes the strange "Number of Channels in Task: 1, Number of Channels in Data: 1" error (see https://github.com/ni/nidaqmx-python/pull/127/files#diff-52340ddf18f49fd7571bdf82b12b90c54ee8b0b8018f3ef93fa7ad13e2ad1bfcL1185-R1184).

### DIFF
--- a/nidaqmx/task.py
+++ b/nidaqmx/task.py
@@ -1170,20 +1170,18 @@ class Task(object):
         number_of_channels = len(channels_to_write.channel_names)
         write_chan_type = channels_to_write.chan_type
 
+        if isinstance(data, list):
+            data = numpy.asarray(data)
+
         element = None
         if number_of_channels == 1:
-            if isinstance(data, list):
-                if isinstance(data[0], list):
-                    self._raise_invalid_write_num_chans_error(
-                        number_of_channels, len(data))
-
-                number_of_samples_per_channel = len(data)
-                element = data[0]
-
-            elif isinstance(data, numpy.ndarray):
+            if isinstance(data, numpy.ndarray):
                 if len(data.shape) == 2:
-                    self._raise_invalid_write_num_chans_error(
-                        number_of_channels, data.shape[0])
+                    if data.shape[0] == 1:
+                        data = data[0]
+                    else:
+                        self._raise_invalid_write_num_chans_error(
+                            number_of_channels, data.shape[0])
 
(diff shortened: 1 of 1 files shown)


zhindes:
Closing due to inactivity

PR #134, opened 2022-01-21
Prepare nidaqmx-python for the first release in 5 years

### LINKED ISSUE #36: no version information available
There used to be a version.py file in nidaqmx, which seems to have been removed.  This was useful in that I could check nidaqmx.version and ensure that the minimum revision was installed for my application. 

Or, please provide a nidaqmx.__version__ like most python packages do.

The fact that older versions had a version.py file means that upgrading will leave the old file there, thus confusing the situation.

zhindes:
_(Note: these commits will be preserved and not squashed when the PR is completed.)_

# Major Changes

* Switched to [poetry](https://python-poetry.org/) build system.
* Updated supported Python versions to 3.7, 3.8, 3.9, and 3.10
* Updated to latest dependencies
    * Note: This required replacing `numpy.bool` with `bool` native type to resolve warnings. These are synonymous now.

# Resolved Issues

* Fixed `test_many_sample_pulse_ticks` test which was previously disabled due to a DAQmx bug that was fixed.
* Added a bridge device to test MAX config to enable `test_list_of_floats_property` to run
* Closes #36

# Testing

Tests pass 100% via tox:
```
  py37: commands succeeded
  py38: commands succeeded
  py39: commands succeeded
  py310: commands succeeded
  congratulations :)
```

Ran readthedocs.io process against a preview branch.


pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+# Documentation, must be in main dependencies (but optional) list for
+# readthedocs integration.
+# See https://github.com/readthedocs/readthedocs.org/issues/4912.
+Sphinx = {version="^4.4", optional=true}
+sphinx_rtd_theme = {version="^1.0", optional=true}
+
+[tool.poetry.extras]
+docs = ["sphinx", "sphinx_rtd_theme"]
+
+[tool.poetry.dev-dependencies]
```
maxxboehme, line 50: You should add tox as a dev dependency since we are using it.

maxxboehme (changes requested):

CONTRIBUTING.md
```diff
@@ -0,0 +1,135 @@
...
+
+```sh
+$ poetry run pytest
+```
+
+To run the unit tests in all Python interpreters supported by **nidaqmx**, run the following
+commands in the root of the distribution:
+
+```sh
+$ pip install tox
```
maxxboehme, line 60: If you add tox to the dev tools I believe you should be able to call `poetry run tox`

zhindes: Yeah, that works (if I add `tox` as a dev dependency, like you described above). It is kind of weird though because then `tox` creates its own dev environments seeded from within the `poetry` dev environment. That isn't a big deal, though, and it makes it a bit more straight-forward. So, yeah, I'll do that :)

maxxboehme: Yeah, it is weird interaction but it is nice to have the tools used listed on the same spot for the most part.

maxxboehme (approved):

PR #169, opened 2022-05-07
Relax numpy version requirement

zariiii9003:
Pip will find the correct numpy version. There is no need to require numpy 1.22 for everyone. There are still many packages that are incompatible with 1.22 (e.g. numba).


zariiii9003:
@zhindes done

PR #180, opened 2022-06-24
Use ndarray.size instead of numpy.prod

zariiii9003:
The numpy.prod() call seems quite expensive:
image

The `_calculate_num_samps_per_chan()` call could be skipped entirely if the user requested to read all available data (=-1), right?

### DIFF
--- a/nidaqmx/_task_modules/read_functions.py
+++ b/nidaqmx/_task_modules/read_functions.py
@@ -26,7 +26,7 @@ def _read_analog_f_64(
 
     error_code = cfunc(
         task_handle, num_samps_per_chan, timeout, fill_mode.value,
-        read_array, numpy.prod(read_array.shape),
+        read_array, read_array.size,
         ctypes.byref(samps_per_chan_read), None)
     check_for_error(error_code, samps_per_chan_read=samps_per_chan_read.value)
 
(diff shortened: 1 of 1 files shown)


zhindes:
> The _calculate_num_samps_per_chan() call could be skipped entirely if the user requested to read all available data (=-1), right?

You mean in the StreamReaders? If so, no, `_calculate_num_samps_per_chan` converts the -1 to a real value.

zhindes (approved):
Good fix, thanks!

zariiii9003:
> You mean in the StreamReaders? If so, no, `_calculate_num_samps_per_chan` converts the -1 to a real value.

`_read_analog_f_64()` calls `DAQmxReadAnalogF64` which accepts -1 to read all values. So unless `_verify_array_shape` is True, that seems unnecessary. 
You could save ~30% time by checking _verify_array_shape first:
```python3
        if self._verify_array_shape:
            number_of_samples_per_channel = (
                self._task._calculate_num_samps_per_chan(
                    number_of_samples_per_channel))
    
            self._verify_array(data, number_of_samples_per_channel, True, True)

        return _read_analog_f_64(
            self._handle, data, number_of_samples_per_channel,
            timeout)
```

zhindes:
Oh, right, great point. I was going to recommend that we move `_calculate_num_samps_per_chan` into `_verify_array`, but that wouldn't be quite right unless we made `number_of_samples_per_channel` and out-param. We want to use a consistent snapshot of the `number_of_samples_per_channel` to read for both `_verify_array` and `_read_analog_f_64`.

Our [stream tests](https://github.com/ni/nidaqmx-python/blob/master/nidaqmx/tests/test_stream_analog_readers_writers.py) don't seem to cover anything useful for -1, so I'd want to include some new tests, some pseudocode:
```py
for verify_array_shape in [True, False]:
   for acq_mode in [Finite, Continuous, On Demand]:
      # setup task
      # if continuous, wait for 0.1s
      if verify_array_shape:
         # allocate not enough data
         expect throw:
            read(-1)

      # allocate plenty of data
      read(-1)
      # validate we got as much data as we wanted (all samples, ~0.1s worth, or 1 sample depending on acq_mode)
      # validate good data
```

Issue #162, opened 2022-03-09
Example for synchronizing tasks / read and write operations


fhchl:
It would be great to have some examples for how synchronize tasks. 

For example, I am have connected the output 0 of my USB-4431 to the input 0 and run

```python
import numpy as np
import nidaqmx as ni
from nidaqmx.constants import TaskMode

def query_devices():
    local_system = ni.system.System.local()
    driver_version = local_system.driver_version

    print('DAQmx {0}.{1}.{2}'.format(driver_version.major_version, driver_version.minor_version,
                                    driver_version.update_version))

    for device in local_system.devices:
        print('Device Name: {0}, Product Category: {1}, Product Type: {2}'.format(
            device.name, device.product_category, device.product_type))


def playrec(
    outdata, sr=96000, input_mapping=['Dev2/ai0'],
    output_mapping=['Dev2/ao0']
):
    nsamples = outdata.shape[0]
    with ni.Task() as read_task, ni.Task() as write_task:
        for o in output_mapping:
            aochan = write_task.ao_channels.add_ao_voltage_chan(o)
            aochan.ao_max = 3.5   # output range of USB-4431
            aochan.ao_min = -3.5
        for i in input_mapping:
            # FIXME: doesn't support multiple output mappings yet.
            aichan = read_task.ai_channels.add_ai_voltage_chan(i)
            aichan.ai_min = -10
            aichan.ai_max = 10

        for task in (read_task, write_task):
            task.timing.cfg_samp_clk_timing(rate=sr, source='OnboardClock', samps_per_chan=nsamples)
        
        # commiting task reduces lag between read and write
        read_task.control(TaskMode.TASK_COMMIT)
        write_task.control(TaskMode.TASK_COMMIT)

        write_task.write(outdata, auto_start=False)
        write_task.start()
        indata = np.asarray(read_task.read(nsamples)).T
        
    return indata

if __name__ == "__main__":
    query_devices()

    sr = 96000
    duration = 1
    t = np.linspace(0, duration, sr*duration, endpoint=False)
    sig = 1 * np.sin(1000*2*np.pi*t)

    indata = playrec(sig)
    import matplotlib.pyplot as plt
    plt.plot(t, indata)
    plt.show()
```

Looking at the plot, I can clearly see that there is around 5 ms of lag between the start of the write and the start of the read task on the device. 

If I run a similar example in Matlab, there is no such lag:

```matlab
duration = 10;
sr = 96000;         % NI 4431

[...]

PR #184, opened 2022-08-14
Add playrec.py example

fhchl:
As discussed in #162.

Gotta quickly test this Monday with a device to see if there didn't sneak any mistake in.

### DIFF
--- a/nidaqmx_examples/playrec.py
+++ b/nidaqmx_examples/playrec.py
@@ -1,0 +1,83 @@
+"""Simultaneous read and write with NI USB-4431 or similar device."""
+
+import numpy as np
+import nidaqmx as ni
+from nidaqmx.constants import WAIT_INFINITELY
+
+
+def query_devices():
+  local = ni.system.System.local()
+  driver = local.driver_version
+
+  print(f'DAQmx {0}.{1}.{2}'.format(
+    driver.major_version, driver.minor_version, driver.update_version))
+
+  for device in local.devices:
+    print('Device Name: {0}, Product Category: {1}, Product Type: {2}'.format(
+      device.name, device.product_category, device.product_type))
+
+
+def playrec(
+  data, samplerate=96000, input_mapping=['Dev2/ai0', 'Dev2/ai1', 'Dev2/ai2'],
+  output_mapping=['Dev2/ao0']
+):
+  """Simultaneous playback and recording though NI device.
+
+  Parameters:
+  -----------
+  data: array_like, shape (nsamples, len(output_mapping))
+    Data to be send to output channels.  
+  samplerate: int
+    Samplerate
+  input_mapping: list of str
+    Input device channels
+  output_mapping: list of str
+    Output device channels
+
+  Returns
+  -------
+  ndarray, shape (nsamples, len(input_mapping))
+    Recorded data
+
+  """
+  max_out_range = 3.5 # output range of USB-4431
+  max_in_range = 10   # input range of USB-4431
+  max_outdata = np.max(np.abs(outdata))
+  if max_outdata > max_out_range:
+    raise ValueError(
+      f"outdata amplitude ({max_outdata:.2f}) larger than allowed range"
+      f"(+-{max_out_range:.2f}).")
+
+  outdata = np.asarray(outdata).T
+  nsamples = outdata.shape[1]
+
+  with ni.Task() as read_task, ni.Task() as write_task:
+    for o in output_mapping:
+      aochan = write_task.ao_channels.add_ao_voltage_chan(o)
+      aochan.ao_max = max_out_range
+      aochan.ao_min = -max_out_range
+    for i in input_mapping:
+      aichan = read_task.ai_channels.add_ai_voltage_chan(i)
+      aichan.ai_min = -max_in_range
+      aichan.ai_max = max_in_range
+
+    for task in (read_task, write_task):
+      task.timing.cfg_samp_clk_timing(rate=samplerate, source='OnboardClock',
+                                      samps_per_chan=nsamples)
+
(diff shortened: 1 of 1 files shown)


zhindes (approved):
Looks great. Let me know when you've done your testing, and I can merge.

nidaqmx_examples/playrec.py
```diff
@@ -0,0 +1,83 @@
+"""Simultaneous read and write with NI USB-4431 or similar device."""
+
+import numpy as np
+import nidaqmx as ni
+from nidaqmx.constants import WAIT_INFINITELY
+
+
+def query_devices():
```
zhindes, line 8: This function isn't used.

nidaqmx_examples/playrec.py
```diff
@@ -0,0 +1,83 @@
...
+    driver.major_version, driver.minor_version, driver.update_version))
+
+  for device in local.devices:
+    print('Device Name: {0}, Product Category: {1}, Product Type: {2}'.format(
+      device.name, device.product_category, device.product_type))
+
+
+def playrec(
+  data, samplerate=96000, input_mapping=['Dev2/ai0', 'Dev2/ai1', 'Dev2/ai2'],
+  output_mapping=['Dev2/ao0']
```
zhindes, line 22: default parameters for these are too application-specific. can you remove them and pass them in __main__?

nidaqmx_examples/playrec.py
```diff
@@ -0,0 +1,83 @@
...
+    write_task.start()  # write_task doesn't start at read_task's start_trigger
+                        # without this
+    indata = read_task.read(nsamples, timeout=WAIT_INFINITELY) # do not time out
+                                                               # for long inputs
+
+  return np.asarray(indata).T
+
+
+if __name__ == '__main__':
+  # excite through one output and record at three inputs
```
zhindes, line 81: this comment will make more sense when you remove the default params and pass them in

nidaqmx_examples/playrec.py
```diff
@@ -0,0 +1,83 @@
...
+    Output device channels
+
+  Returns
+  -------
+  ndarray, shape (nsamples, len(input_mapping))
+    Recorded data
+
+  """
+  max_out_range = 3.5 # output range of USB-4431
+  max_in_range = 10   # input range of USB-4431
```
zhindes, line 44: you might be able to query these from `system.device.ai_voltage_rngs` and `system.device.ao_voltage_rngs`

nidaqmx_examples/playrec.py
```diff
@@ -0,0 +1,83 @@
...
+      f"(+-{max_out_range:.2f}).")
+
+  outdata = np.asarray(outdata).T
+  nsamples = outdata.shape[1]
+
+  with ni.Task() as read_task, ni.Task() as write_task:
+    for o in output_mapping:
+      aochan = write_task.ao_channels.add_ao_voltage_chan(o)
+      aochan.ao_max = max_out_range
+      aochan.ao_min = -max_out_range
```
zhindes, line 58: these are input params on `add_ao_voltage_chan`: `min_val`, `max_val`

nidaqmx_examples/playrec.py
```diff
@@ -0,0 +1,83 @@
...
+    for i in input_mapping:
+      aichan = read_task.ai_channels.add_ai_voltage_chan(i)
+      aichan.ai_min = -max_in_range
+      aichan.ai_max = max_in_range
```
zhindes, line 62: these are input params on `add_ai_voltage_chan`: `min_val`, `max_val`

nidaqmx_examples/playrec.py
```diff
@@ -0,0 +1,83 @@
...
+                                      samps_per_chan=nsamples)
+
+    # trigger write_task as soon as read_task starts
+    write_task.triggers.start_trigger.cfg_dig_edge_start_trig(
+            read_task.triggers.start_trigger.term)
+    write_task.write(outdata, auto_start=False)
+    write_task.start()  # write_task doesn't start at read_task's start_trigger
+                        # without this
+    indata = read_task.read(nsamples, timeout=WAIT_INFINITELY) # do not time out
+                                                               # for long inputs
```
zhindes, line 75: you switched to trailing comments here. do you mind moving them to their own line like the rest of the file for consistency?

zhindes (changes requested):

nidaqmx_examples/playrec.py
fhchl: Changed example slightly such that this function is actually useful here.

nidaqmx_examples/playrec.py
fhchl: Good idea.

nidaqmx_examples/playrec.py
fhchl: Agreed.

nidaqmx_examples/playrec.py
fhchl: Good idea.

nidaqmx_examples/playrec.py
fhchl: I was wondering why these are not automatically set when passing a device to `add_ai_voltage_chan`. Seems like they must be passed explicitly.

nidaqmx_examples/playrec.py
fhchl: see above

nidaqmx_examples/playrec.py
fhchl: Yes.

fhchl:
Tested the current example which works fine and throws the value error when the output data is to large.

zhindes (approved):

PR #208, opened 2023-02-15
[RFC] DAQmx code generator for attributes

Sakthi-SM:
* This PR adds a code generator for DAQmx classes and it focuses on generating the `ai_channel.py` class. This PR will be complete once the scrapigen metadata update is complete [scrapigen update](https://github.com/ni/grpc-device-scrapigen/pull/108)
* Modified the folder structure to include the code generator and the generated files.
* Added  `attribute.py` and `script_info.py` to the metadata. Currently the `attribute.py` contains only sample meta data necessary for generating some attributes in `ai_channel.py`
* Added the `mako` templates necessary for generating attributes.
* The metadata will be read and converted to the structure defined in the `attributes.py` and reused in the `mako` files.


src/codegen/__main__.py
```diff
@@ -0,0 +1,70 @@
...
+        0: logging.DEBUG,
+        1: logging.INFO,
+        2: logging.WARNING,
+        3: logging.ERROR,
+        4: logging.CRITICAL,
+    }[verbosity]
+
+    return args
+
+def main(argv=None):
```
Sakthi-SM, line 55: Users can run the generator by passing the output directory path in the args. 
Example : `python -m codegen --dest outputPath`

src/codegen/metadata/attributes.py
```diff
@@ -0,0 +1,130 @@
+attributes = {
```
Sakthi-SM, line 1: This is just a sample metadata which contains the required data for generating some attributes present in `ai_channel.py`

src/codegen/utilities/attribute_helpers.py
```diff
@@ -0,0 +1,15 @@
+import logging
+from collections import namedtuple
+from codegen.utilities.properties.attribute import attribute
+
+_logger = logging.getLogger(__name__)
+_logger.addHandler(logging.NullHandler())
+
+def get_attributes(metadata, className):
+    attributes_metadata = []
```
Sakthi-SM, line 9: This method converts the metadata to a list of `attribute` objects that are part of the specified class.

src/codegen/generator.py
```diff
@@ -0,0 +1,44 @@
...
+from mako.lookup import TemplateLookup
+from mako.template import Template
+
+import codegen.metadata as metadata
+
+_logger = logging.getLogger(__name__)
+_logger.addHandler(logging.NullHandler())
+
+def _get_metadata():
+    return metadata.metadata
```
keer1111, line 16: if possible avoid metadata.metadata

src/codegen/metadata/script_info.py
```diff
@@ -0,0 +1,8 @@
+script_info = {
```
keer1111, line 1: consider having the whole moduleinfo instead of having only the necessary key values

src/codegen/templates/property_deleter_template.py.mako
```diff
@@ -0,0 +1,37 @@
+<%def name="script_property_deleter(attribute)">\
+<%
+        from codegen.utilities.text_wrappers import wrap
+    %>\
+    @${attribute.name}.deleter
+    def ${attribute.name}(self):
+<%
+        argtypes = []
+        for handle_parameter in attribute.handle_parameters:
+            if handle_parameter.ctypes_data_type == 'ctypes.c_char_p':
```
keer1111, line 10: look for simplified notations if possible

src/codegen/templates/nidaqmx/_task_modules/channels/ai_channel.py.mako
```diff
@@ -0,0 +1,31 @@
...
+    """
+    __slots__ = []
+
+    def __repr__(self):
+        return 'AIChannel(name={0})'.format(self._name)
+
+<%namespace name="property_template" file="/property_template.py.mako"/>\
+%for attribute in attributes:
+${property_template.script_property(attribute)}\
+%endfor
```
keer1111, line 31: skipping templates review since it is reused from code geb

src/codegen/templates/nidaqmx/_task_modules/channels/ai_channel.py.mako
```diff
@@ -0,0 +1,31 @@
+<%
```
keer1111, line 1: consider removing daqmx nesting in the folder

src/codegen/utilities/attribute_helpers.py
```diff
@@ -0,0 +1,15 @@
+import logging
+from collections import namedtuple
+from codegen.utilities.properties.attribute import attribute
+
+_logger = logging.getLogger(__name__)
+_logger.addHandler(logging.NullHandler())
+
+def get_attributes(metadata, className):
```
keer1111, line 8:
```suggestion
def get_attributes(metadata, class_name):
```
since it is python use snakecase

src/codegen/utilities/attribute_helpers.py
```diff
@@ -0,0 +1,15 @@
...
+    attributes_metadata = []
+    for group_Name, attributes in metadata["attributes"].items():
```
keer1111, line 10:
```suggestion
    for group_name, attributes in metadata["attributes"].items():
```

src/codegen/utilities/properties/attribute.py
```diff
@@ -0,0 +1,281 @@
+class attribute():
```
keer1111, line 1:
```suggestion
class Attribute():
```
Adopt [coding conventions](https://github.com/ni/python-styleguide/blob/main/docs/Coding-Conventions.md)

src/codegen/utilities/text_wrappers.py
```diff
@@ -0,0 +1,45 @@
...
+
+def docstring_wrap(initial_indent, subsequent_indent):
+    """
+    Returns custom Mako filter function that wraps docstring text.
+
+    Returning another function from within this function is a trick used to
+    enable Mako filter functions to accept arguments.
+    """
+
+    def d(text):
```
keer1111, line 40: d?

keer1111 (changes requested):

generator/nidaqmx_python_generator/attribute_helpers.py
```diff
@@ -0,0 +1,15 @@
+import logging
```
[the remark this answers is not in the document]
Sakthi-SM: Yes it is a utility

src/codegen/utilities/attribute_helpers.py
Sakthi-SM: Updated

src/codegen/utilities/properties/attribute.py
Sakthi-SM: Updated

src/codegen/utilities/text_wrappers.py
Sakthi-SM: Renamed the functions appropriately

generated/_task_modules/channels/ai_channel.py
```diff
@@ -0,0 +1,202 @@
+# Do not edit this file; it was automatically generated.
```
Sakthi-SM, line 1: This file is generated using the sample metadata which contains only few of the attributes present in this class.

generated/_task_modules/channels/ai_channel.py
```diff
@@ -0,0 +1,202 @@
...
+    __slots__ = []
+
+    def __repr__(self):
+        return 'AIChannel(name={0})'.format(self._name)
+
+
+    @property
+    def ai_atten(self):
+        """
+        Specifies the amount of attenuation to use.
```
zhindes, line 28: The existing code indicates the type returned. I don't think that is a Python convention, but it does seem useful.

generated/_task_modules/channels/ai_channel.py
```diff
@@ -0,0 +1,202 @@
...
+
+        error_code = cfunc(
+            self._handle, self._name)
+        check_for_error(error_code)
+
+
+    @property
+    def ai_raw_samp_justification(self):
+        """
+        Indicates the justification of a raw sample from the device.
```
zhindes, line 77: Especially here, the enum type returned would be useful to know.

src/codegen/properties/attribute.py
```diff
@@ -0,0 +1,281 @@
...
+        self._ctypes_data_type = attribute_metadata["ctypes_data_type"]
+        self._python_data_type = attribute_metadata["python_data_type"]
+        self._description = attribute_metadata["description"]
+        self._has_explicit_read_buffer_size = attribute_metadata["has_explicit_read_buffer_size"]
+        self._bitfield_enum = attribute_metadata["bitfield_enum"]
+        self._is_list = attribute_metadata["is_list"]
+        self._calling_convention = attribute_metadata["calling_convention"]
+        self._c_function_name = attribute_metadata["c_function_name"]
+        if("is_object" in attribute_metadata): 
+            self._is_object = attribute_metadata["is_object"]
```
zhindes, line 19: I would expect `_is_object` to be set to False in the else case. You could do this:

```python
self._is_object = attribute_metadata.get("is_object", False)
```

src/codegen/properties/attribute.py
```diff
@@ -0,0 +1,281 @@
...
+        if("read_buffer_size" in attribute_metadata):
+            self._read_buffer_size = attribute_metadata["read_buffer_size"]
```
zhindes, line 21: I see you handle this explicitly in the getter, but again using `dict.get` could simplify your code.

```python
self._read_buffer_size = attribute_metadata.get("read_buffer_size")
```

The default is None if you don't specify.

src/codegen/properties/attribute.py
```diff
@@ -0,0 +1,281 @@
...
+        self._python_class_name = attribute_metadata["python_class_name"]
+        self._handle_parameters = []
+        self._object_constructor_params = []
+        if("handle_parameters" in attribute_metadata):
```
zhindes, line 25: I'll stop commenting on these, but all of them could be simplified by using `dict.get`

src/codegen/templates/property_deleter_template.py.mako
zhindes: I agree. You could move this logic into the `attribute.py` class to make these templates more generic.

src/codegen/templates/property_deleter_template.py.mako
```diff
@@ -0,0 +1,37 @@
...
+                argtypes.append('ctypes_byte_str')
+            else:
+                argtypes.append(handle_parameter.ctypes_data_type)
+    %>\
+    %if len(attribute.c_function_name) < 33:
```
zhindes, line 15: A comment here about why we care about length would be nice. It looks like we line-wrap if its too long? Alternatively, just always line-wrap to simply this code?

src/codegen/templates/property_deleter_template.py.mako
```diff
@@ -0,0 +1,37 @@
...
+        cfunc = lib_importer.${'windll' if attribute.calling_convention=='StdCall' else 'cdll'}.DAQmxReset${attribute.c_function_name}
+    %else:
+        cfunc = (lib_importer.${'windll' if attribute.calling_convention=='StdCall' else 'cdll'}.
```
zhindes, line 18: I would make an accessor in the attribute class for this snippet called `get_lib_importer_type` or something:

```python
${'windll' if attribute.calling_convention=='StdCall' else 'cdll'}
```

zhindes: With that, you might not even need `calling_convention` at all anymore!

src/codegen/templates/property_deleter_template.py.mako
```diff
@@ -0,0 +1,37 @@
...
+                 DAQmxReset${attribute.c_function_name})
+    %endif
+        if cfunc.argtypes is None:
+            with cfunc.arglock:
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        ${', '.join(argtypes) | wrap(24, 24)}]
```
zhindes, line 25: LOVE these `wrap` and `docstring_wrap` helpers. I wish the parameters were more documenting, so I could understand what they mean when looking at a template. Can you make them named parameters here?

```python
wrap(init_indent=24, subsequent_indent=24)
```

src/codegen/utilities/text_wrappers.py
```diff
@@ -0,0 +1,45 @@
...
+
+ An extra character is subtracted from the line width of the code text wrapper
+ to allow for closing brackets, since the Mako template will be prettier
+ without lines like [${variable + ']' | wrap(4. 8)}
+"""
+wrapper = textwrap.TextWrapper(width=78, break_long_words=False)
+docstring_wrapper = textwrap.TextWrapper(width=72, break_long_words=False)
+
+
+def wrap(initial_indent, subsequent_indent):
```
zhindes, line 16: `subsequent_indent` seems to often be equal to `initial_indent`. You could make it `None` by default. And then add code here that sets it to initial_indent if it is `None`. This could cleanup callers quite a bit!

src/codegen/templates/property_deleter_template.py.mako
zhindes: Actually, I gave a recommendation down below about how to eliminate `subsequent_indent` in these calls if they're equal. In that case a single parameter to wrap doesn't need to be named. It is very self-documenting!

src/codegen/templates/property_getter_template.py.mako
```diff
@@ -0,0 +1,165 @@
...
+\
+    %if len(attribute.c_function_name) < 33:
+        cfunc = lib_importer.${'windll' if attribute.calling_convention=='StdCall' else 'cdll'}.DAQmxGet${attribute.c_function_name}
+    %else:
+        cfunc = (lib_importer.${'windll' if attribute.calling_convention=='StdCall' else 'cdll'}.
+                 DAQmxGet${attribute.c_function_name})
+    %endif
+\
+## Create argument ctypes types list.
+    %if attribute.calling_convention == 'StdCall':
```
zhindes, line 30: There doesn't appear to be an else. So maybe just strip all the logic around checking the `calling_convention` and just assert that `calling_convention == 'StdCall'` in the attribute class?

src/codegen/templates/property_getter_template.py.mako
```diff
@@ -0,0 +1,165 @@
...
+                    attribute.bitfield_enum is None):
+                argtypes.append("wrapped_ndpointer(dtype={0}, flags=('C','W'))"
+                                .format(attribute.ctypes_data_type))
+            elif attribute.ctypes_data_type == 'ctypes.c_char_p':
+                argtypes.append(attribute.ctypes_data_type)
+            else:
+                argtypes.append('ctypes.POINTER({0})'.format(attribute.ctypes_data_type))
+
+            if attribute.has_explicit_read_buffer_size:
+                argtypes.append('ctypes.c_uint')
```
zhindes, line 49: Same comments here from above. Moving this logic to the `attribute` class could clean this up a bit!

zhindes (commented):
Nice work! I really like what you've done so far. My comments are pretty minor cleanup.

generated/_task_modules/channels/ai_channel.py
bkeryan: BTW, Python has multiple docstring conventions. nidaqmx-python seems to use Google style docstrings, and including the return type for property getters seems to be a standard part of that convention: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

src/codegen/generator.py
```diff
@@ -0,0 +1,43 @@
...
+    with open(output_path, "w+", newline="") as f:
+        f.write(template.render(data=metadata))
+
+
+def generate(args):
+    _logger.info(f"Generating files into {args.dest}")
+
+    os.makedirs(args.dest, exist_ok=True)
+
+    metadata = _get_metadata()
```
bkeryan, line 40: This variable shadows a module alias. Consider not aliasing.

src/codegen/properties/attribute.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def access(self):
+        """
+        Specifies if the attribute is read/write.
+        Supported Values
+        - `read`: The attribute is readable.(get)
+        - `write`: The attribute is writeable or can be set with new values.(set)
+        - `read-write`: The attribute can be both be read and written.(get and set)
+        This is used to decide the generation of getters and setters of a property representing the attribute.
```
bkeryan, line 58: The rest of the code in this repo seems to use Google style docstrings, but these docstrings are not formatted that way.

See: https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

> A docstring should be organized as a summary line (one physical line not exceeding 80 characters) terminated by a period, question mark, or exclamation point. When writing more (encouraged), this must be followed by a blank line, followed by the rest of the docstring starting at the same cursor position as the first quote of the first line.

Also see: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html


`ni-python-styleguide lint` probably warns about this.

src/codegen/properties/attribute.py
```diff
@@ -0,0 +1,281 @@
...
+        The name of the python class this attribute belongs to.
+        This is used to determine which attribute goes to which class when generating the python code.
+        """
+        return self._python_class_name
+
+    @property
+    def is_object(self):
+        """
+        'True': The attribute represents a known object.
+        'False' : The attribute represents the `type` value as usual.
```
bkeryan, line 151: I don't know if doc tools like Sphinx will parse this correctly. https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html uses "type: description" for properties, not "value: description\nvalue2: description".

`ni-python-styleguide lint` probably warns about this.

src/codegen/properties/attribute.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def ctypes_data_type(self):
+        """
+        The type of the attribute as per the ctypes definition in python.
+        This is used to provide the type of the attribute when making c function calls in python.
+        """
+        return self._ctypes_data_type
+
+class Parameter():
```
bkeryan, line 245: Why is `Parameter` in the `attribute` module?

src/codegen/properties/attribute.py
```diff
@@ -0,0 +1,281 @@
+class Attribute():
```
bkeryan, line 1: What do the parentheses signify? In Python 2, `class Attribute:` and `class Attribute(object):` had subtly different meanings, but AFAIK in Python 3 you can just write `class Attribute:`.

generated/_task_modules/channels/ai_channel.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+    @property
+    def ai_bridge_poly_forward_coeff(self):
+        """
+        Specifies an list of coefficients for the polynomial that
+            converts electrical values to physical values. Each element
+            of the list corresponds to a term othe equation. For
+            example, if index three of the list is 9,the fourth term of
+            the equation is 9x^3.
```
bkeryan, line 103: Part of this description is cut off.
- "term of the equation" changed to "term othe equation".
- "is 9, the" is missing a space.

image

generated/_task_modules/channels/ai_channel.py
```diff
@@ -0,0 +1,202 @@
...
+                temp_size = 0
+            elif size_or_code > 0 and temp_size == 0:
+                # Buffer size obtained, use to retrieve data.
+                temp_size = size_or_code
+            else:
+                break
+
+        check_for_error(size_or_code)
+
+        return str._factory(self._handle, val.value.decode('ascii'))
```
bkeryan, line 201: This should return `Channel._factory(...)` not `str._factory(...)`.

image

generated/_task_modules/channels/ai_channel.py
bkeryan: Another reason to add the type: generating exactly the same code makes it easier to diff the output. Once we're generating the same output, then we can start making changes to the output.

generated/_task_modules/channels/ai_channel.py
```diff
@@ -0,0 +1,202 @@
...
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        lib_importer.task_handle, ctypes_byte_str,
+                        ctypes.POINTER(ctypes.c_int)]
+
+        error_code = cfunc(
+            self._handle, self._name, ctypes.byref(val))
+        check_for_error(error_code)
+
+        return int(val.value)
```
bkeryan, line 93: This should return `DataJustification(val.value)` not `int(val.value)`.

image

generated/_task_modules/channels/ai_channel.py
bkeryan: When you use the full metadata, does this output attributes in the same order as before? If not, this is going to be hard to validate without fixing the order.

bkeryan, line 1: Why did you rename the `nidaqmx_examples` directory in this PR? This does not seem related to the code generator.

Also, the `nidaqmx_examples` directory has an `__init__.py` file, making it a package. We should consider whether we really want to change the package name.

(This isn't related to ai_channel.py but GitHub doesn't seem to let you attach issues to renamed files.)

src/codegen/templates/property_getter_template.py.mako
bkeryan: The DAQmx C API should only use cdecl for varargs functions, but it looks like a few other functions use it, possibly by mistake. I don't think any attribute-specific functions use it, which is probably why there was no else case in this template.

src/codegen/__main__.py
```diff
@@ -0,0 +1,70 @@
...
+
+import codegen.generator as generator
+
+_logger = logging.getLogger(__name__)
+_logger.addHandler(logging.NullHandler())
+
+def _parse_args(argv):
+    import argparse
+
+    parser = argparse.ArgumentParser()
```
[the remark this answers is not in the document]
mshafer-NI: > We have a module, niclickutils, which single-sources the verbose/quiet stuff below, but it's currently only available for internal projects.

I want to open source it, but have a lot of other Python stuff on my plate right now :D

Either way, new code should be using [`click` per our approved technologies](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/35648/Command-Line-Interface-Parser)

src/codegen/properties/attribute.py
Sakthi-SM: As suggested, used `dict.get` to get the value out of a dictionary and updated all instances where it has been used.

src/codegen/utilities/attribute_helpers.py
```diff
@@ -0,0 +1,15 @@
...
+from codegen.properties.attribute import Attribute
+
+_logger = logging.getLogger(__name__)
+_logger.addHandler(logging.NullHandler())
+
+def get_attributes(metadata, class_name):
+    attributes_metadata = []
+    for group_name, attributes in metadata["attributes"].items():
+        for id, attribute_data in attributes.items():
+            if(attribute_data["python_class_name"] == class_name):
```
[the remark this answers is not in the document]
Sakthi-SM: Currently, the pipeline doesn't have any actions set up and  adding the necessary validations will be taken care  in [Add build actions to the NI DAQmx pipeline](https://dev.azure.com/ni/DevCentral/_workitems/edit/2305330) in a different PR.
Removed the parentheses in all the instances of if-statements.

generated/_task_modules/channels/ai_channel.py
Sakthi-SM: Updated the getter template to return `object_type` instead of `python_data_type`, if the attribute has `object_type`

generated/_task_modules/channels/ai_channel.py
Sakthi-SM: Updated the getter template to return the specified `enum` instead of `python_data_type`.

generated/_task_modules/channels/ai_channel.py
Sakthi-SM: Yes Brad, the order of the generated attributes remains similar to the current code. We'll make sure that there is no discrepancy there.

Sakthi-SM: Since the repository is named as nidaqmx-python, we felt that using nidaqmx in internal folder would be redundant. As you pointed out since we are using it as package, I have reverted this change.

src/codegen/properties/attribute.py
Sakthi-SM: As a general notation, If an object is callable it means we will use the round parentheses to make a call to that object. But as you have mentioned, Python 3 removes need for such a notation for python class. Since it doesn't make any difference, I'll remove the parentheses.

src/codegen/properties/attribute.py
Sakthi-SM: Updated the description based on the google style guide

src/codegen/properties/attribute.py
Sakthi-SM: Moved `Parameter` to a separate module

src/codegen/properties/attribute.py
Sakthi-SM: Updated the description as per the style guide.

src/codegen/templates/property_deleter_template.py.mako
Sakthi-SM: Added a helper function in `attribute` to return the import type. The mako template will make use of that helper function to determine if it is windll/ cdll call.

generated/_task_modules/channels/ai_channel.py
Sakthi-SM: Since the metadata was handwritten, the typos in metadata led to this difference. I have updated the metadata accordingly.

generated/_task_modules/channels/ai_channel.py
Sakthi-SM: I have added helpers methods to specify the return type as part of the documentation. As of now for enums/classes, only the name of the object will be added in the description not the entire path. For example, the expected docstring is `class: nidaqmx.system.scale.Scale: documentation` but currently documentation will look like `class: Scale: documentation`. We are yet to add the part of dynamically adding the import path of enums/classes, this will be added in consecutive PRs.

src/codegen/properties/attribute.py
bkeryan: A class definition is not a function call. The parentheses specify a list of base classes. I guess this list is allowed to be empty, at least in Python 3, but I've never seen code that specified an empty list before.

src/codegen/properties/attribute.py
```diff
@@ -15,33 +17,31 @@ def __init__(self, id, attribute_metadata):
...
+        if"object_constructor_params" in attribute_metadata:
             for name, parameter_data in attribute_metadata["object_constructor_params"].items():
                 self._object_constructor_params.append(Parameter(name, parameter_data))
-        if("has_explicit_write_buffer_size" in attribute_metadata):
-            self._has_explicit_write_buffer_size = attribute_metadata["has_explicit_write_buffer_size"]
-        if("enum" in attribute_metadata):
+        self._has_explicit_write_buffer_size = attribute_metadata.get("has_explicit_write_buffer_size", False)
+        if "enum" in attribute_metadata:
             self._enum = attribute_metadata["enum"]
             self._is_enum = True
```
zhindes, line 35: I would expect this to be `False` in the else case here.

src/codegen/properties/attribute.py
```diff
@@ -15,33 +17,31 @@ def __init__(self, id, attribute_metadata):
...
         self._object_constructor_params = []
-        if("handle_parameters" in attribute_metadata):
+        if "handle_parameters" in attribute_metadata:
             for name, parameter_data in attribute_metadata["handle_parameters"].items():
                 self._handle_parameters.append(Parameter(name, parameter_data))
-        if("object_has_factory" in attribute_metadata):
-            self._object_has_factory = attribute_metadata["object_has_factory"]
-        if("object_constructor_params" in attribute_metadata):
+        self._object_has_factory = attribute_metadata.get("object_has_factory", False)
+        if"object_constructor_params" in attribute_metadata:
```
zhindes, line 29: odd spacing

src/codegen/properties/attribute.py
```diff
@@ -15,33 +17,31 @@ def __init__(self, id, attribute_metadata):
...
-            self._is_object = attribute_metadata["is_object"]
-        if("read_buffer_size" in attribute_metadata):
-            self._read_buffer_size = attribute_metadata["read_buffer_size"]
+        self._is_object = attribute_metadata.get("is_object", False)
+        self._read_buffer_size = attribute_metadata.get("read_buffer_size")
         self._python_class_name = attribute_metadata["python_class_name"]
         self._handle_parameters = []
         self._object_constructor_params = []
-        if("handle_parameters" in attribute_metadata):
+        if "handle_parameters" in attribute_metadata:
```
zhindes, line 25: set `_handle_parameters` to an empty list or None in the else case

src/codegen/properties/attribute.py
```diff
@@ -15,33 +17,31 @@ def __init__(self, id, attribute_metadata):
...
             for name, parameter_data in attribute_metadata["handle_parameters"].items():
                 self._handle_parameters.append(Parameter(name, parameter_data))
```
bkeryan, line 27: [style] Consider using list comprehensions with get()

```
self._handle_parameters = [
    Parameter(name, parameter_data)
    for name, parameter_data in attribute_metadata.get("handle_parameters", []).items()
]
```

src/codegen/properties/attribute.py
```diff
@@ -221,61 +223,76 @@ def read_buffer_size(self):
...
-        The cvi name of the parameter.
-        This is kept for the gRPC client implementation.
-        """
-        return self._cvi_name
\ No newline at end of file
+        return argtypes
+    
+    def get_return_type(self):
+        if self.is_enum and not self.is_list:
+            return 'class: {0}'.format(self.enum)
```
bkeryan, line 279: >  For example, the expected docstring is class: nidaqmx.system.scale.Scale: documentation

Take a closer look. It uses `:class:X` not `class: X`.

image

src/codegen/properties/attribute.py
```diff
@@ -97,23 +99,24 @@ def has_explicit_write_buffer_size(self):
...
+        
         If the value is `True` then the `_factory` method is used for instantiation of the object.
         """
         return self._object_has_factory
 
     @property
     def is_enum(self):
         """
-        Represents if the attribute is an enum or not.
+        booL: Represents if the attribute is an enum or not.
```
bkeryan, line 111: booL -> bool

src/codegen/metadata/attributes.py
```diff
@@ -0,0 +1,130 @@
...
+                    "cvi_name": "taskHandle"
+                },
+                'channel':{
+                    "accessor": "self._name", 
+                    "ctypes_data_type": "ctypes.c_char_p", 
+                    "cvi_name": "channel"
+                }
+            },
+            'is_object': False,
+            'python_class_name':'AI_Channel'
```
bkeryan, line 30: How much of this metadata will no longer be used if we switch to the generic attribute functions as proposed in https://github.com/ni/nidaqmx-python/pull/210 ?

src/codegen/__main__.py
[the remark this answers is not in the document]
Sakthi-SM: Thanks for the suggestion. I have replaced the argument parser with click.

src/codegen/properties/attribute.py
Sakthi-SM: Thanks for pointing this out. Any particular reason why we are following this format for class/enum objects alone?

src/codegen/metadata/attributes.py
subash-suresh: I see the `has_explicit_read_buffer_size` property may get obsolete once we do the implementation in #210, but other than that most of the other properties would still be needed for the property documentation and return type modifications after we make the c method call.

src/codegen/properties/attribute.py
```diff
@@ -22,9 +22,10 @@ def __init__(self, id, attribute_metadata):
         self._python_class_name = attribute_metadata["python_class_name"]
         self._handle_parameters = []
         self._object_constructor_params = []
-        if "handle_parameters" in attribute_metadata:
-            for name, parameter_data in attribute_metadata["handle_parameters"].items():
-                self._handle_parameters.append(Parameter(name, parameter_data))
+        self._handle_parameters = [
+        Parameter(name, parameter_data)
+        for name, parameter_data in attribute_metadata.get("handle_parameters", []).items()
+        ]
```
zhindes, line 28: This formatting looks a bit off. I don't think we have [black](https://pypi.org/project/black/) integrated into the build here, but it would probably make something prettier.

zhindes (commented):
Overall, I think this is great. I have a hard time doing a "final" review without seeing the final generated diff. I made a comment on that file with an approach that I think will make that more useful.

src/codegen/metadata/attributes.py
bkeryan: The generic attribute functions take an explicit read buffer size when getting a string or array.

Example:
``` 
DAQmxErrChk (DAQmxGetChanAttribute(taskHandle,chan,DAQmx_PhysicalChanName,chan,256));
```

generated/_task_modules/channels/ai_channel.py
```diff
@@ -0,0 +1,203 @@
+# Do not edit this file; it was automatically generated.
```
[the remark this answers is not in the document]
bkeryan: I like this idea.

At the end of the day, I want to see the codegen diffs so that I can verify that there are no unexpected changes.

src/codegen/properties/attribute.py
bkeryan: It's the format Sphinx uses for cross-references: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-class

The format is really `:py:class:` but Sphinx lets us omit `:py` because it's the default domain: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-default-domain

src/codegen/properties/attribute.py
```diff
@@ -276,13 +277,13 @@ def get_argument_types(self):
...
+            return ':class: {0}'.format(self.enum)
         elif self.is_object and not self.is_list:
-            return 'class: {0}'.format(self.object_type)
+            return ':class: {0}'.format(self.object_type)
         elif self.is_enum and self.is_list:
-            return 'class: list({0})'.format(self.enum)
+            return ':class: list({0})'.format(self.enum)
         elif self.is_object and self.is_list:
-            return 'class: list({0})'.format(self.object_type)
+            return ':class: list({0})'.format(self.object_type)
```
bkeryan, line 286: These have an extra space after `:class:` and are missing the backticks around the type. See https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects

image

src/codegen/generator.py
```diff
@@ -32,12 +32,12 @@ def _generate_file(metadata, template_file_name, output_path):
...
 
-    os.makedirs(args.dest, exist_ok=True)
+    os.makedirs(dest, exist_ok=True)
 
     codegen_metadata = _get_metadata()
 
     for info in codegen_metadata["script_info"]["modules"]:
-        _generate_file(codegen_metadata, info["templateFile"], args.dest / info["relativeOutputPath"])
\ No newline at end of file
+        _generate_file(codegen_metadata, info["templateFile"], dest + "\\" + info["relativeOutputPath"])
```
bkeryan, line 43: Don't change `dest` to a string and hardcode the Windows path separator. Keep using pathlib so this script works on Linux too.

src/codegen/__main__.py
```diff
@@ -1,67 +1,44 @@
...
         2: logging.WARNING,
         3: logging.ERROR,
         4: logging.CRITICAL,
     }[verbosity]
 
-    return args
+    return logging_level
+
+@click.command()
+@click.option('--dest', type=str)
```
bkeryan, line 30: `type=pathlib.Path`

generated/_task_modules/channels/ai_channel.py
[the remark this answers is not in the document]
Sakthi-SM: As per the suggestion, I have broken down this RFC PR into two PRs with the first PR updating the structure and copying the existing daqmx file [PR1](https://github.com/ni/nidaqmx-python/pull/211). Second PR will have the necessary items for generating `ai_channel.py`.[PR2](https://github.com/ni/nidaqmx-python/pull/212). This RFC will be abandoned.

src/codegen/properties/attribute.py
Sakthi-SM: As per the latest update in scrapigen, this logic will be removed in the subsequent PRs.

Sakthi-SM:
Discarding this PR as it has been broken down into subsequent smaller PRs.
1. [Update Folder Structure](https://github.com/ni/nidaqmx-python/pull/211)
2. [Code generator](https://github.com/ni/nidaqmx-python/pull/212)

PR #210, opened 2023-02-17
[RFC] Prototype for adding gRPC support for daqmx

subash-suresh:
-This PR is a prototype on a couple of daqmx apis, based on the implementation defined in
 [Pull Request 440269](https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/440269): HLD for Adding gRPC support for nidaqmx python drivers.
 
In this implementation,
- the nidaqmx.proto and the required gRPC stubs for the same has been added.
- Two new classes were created, `libraray_interpreter` for the c library calls and `grpc_interpreter` for the gRPC calls.
- These interpreters contain a very small collection of methods for the prototype, it just contains methods to create and close a task instance and a function to create a ai channel. Additionally, it contains a few APIs to read channel and task attributes.
- The `task.py` file has been updated to create and inject interpreter objects into its subclasses.


zhindes:
I like it, no concerns with the direction!

nidaqmx/task.py
```diff
@@ -2,6 +2,8 @@
 import numpy
 import six
 import warnings
+import nidaqmx._grpc_stub_interpreter as _grpc_stub_interpreter
```
bkeryan, line 5: This eagerly imports gRPC modules.

nimi-python lazily imports gRPC modules, so that non-gRPC applications will still work without installing `grpcio`:
https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/session.py#L6866

```
        if grpc_options:
            import nidcpower._grpc_stub_interpreter as _grpc_stub_interpreter
            interpreter = _grpc_stub_interpreter.GrpcStubInterpreter(grpc_options)
        else:
            interpreter = _library_interpreter.LibraryInterpreter(encoding='windows-1251')
```

It doesn't declare a hard dependency on `grpcio`: https://github.com/ni/nimi-python/blob/master/generated/nidcpower/setup.py

nidaqmx/_library_interpreter.py
```diff
@@ -0,0 +1,241 @@
...
+            with cfunc.arglock:
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        lib_importer.task_handle]
+
+        error_code = cfunc(
+            self._task_handle)
+        check_for_error(error_code)
+    
+    def add_ai_voltage_chan(
```
bkeryan, line 70: I think it would make more sense for the interpreter class method names to follow the C API, like `clear_task` and `create_ai_voltage_chan`.

Rationale: many of the Python method names, such as `close`, only make sense in the context of the class where they are defined.

nidaqmx/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,394 @@
...
+        self._system_handle = value
+
+    def get_system_handle(self):
+        return self._system_handle
+
+    def set_scale_handle(self, value=session_grpc_types.Session()):
+        self._scale_handle = value
+
+    def get_scale_handle(self):
+        return self._scale_handle
```
bkeryan, line 37: Scales do not have handles. They only have names. I think the scale class (not the interpreter) should store the scale name.

```
/******************************************************/
/***              Scale Configurations              ***/
/******************************************************/


int32 __CFUNC     DAQmxCreateLinScale            (const char name[], float64 slope, float64 yIntercept, int32 preScaledUnits, const char scaledUnits[]);
int32 __CFUNC     DAQmxCreateMapScale            (const char name[], float64 prescaledMin, float64 prescaledMax, float64 scaledMin, float64 scaledMax, int32 preScaledUnits, const char scaledUnits[]);
int32 __CFUNC     DAQmxCreatePolynomialScale     (const char name[], const float64 forwardCoeffs[], uInt32 numForwardCoeffsIn, const float64 reverseCoeffs[], uInt32 numReverseCoeffsIn, int32 preScaledUnits, const char scaledUnits[]);
int32 __CFUNC     DAQmxCreateTableScale          (const char name[], const float64 prescaledVals[], uInt32 numPrescaledValsIn, const float64 scaledVals[], uInt32 numScaledValsIn, int32 preScaledUnits, const char scaledUnits[]);
int32 __CFUNC     DAQmxCalculateReversePolyCoeff (const float64 forwardCoeffs[], uInt32 numForwardCoeffsIn, float64 minValX, float64 maxValX, int32 numPointsToCompute, int32 reversePolyOrder, float64 reverseCoeffs[]);

int32 __CFUNC_C   DAQmxGetScaleAttribute         (const char scaleName[], int32 attribute, void *value, ...);
int32 __CFUNC_C   DAQmxSetScaleAttribute         (const char scaleName[], int32 attribute, ...);
```

https://github.com/ni/grpc-device/blob/main/generated/nidaqmx/nidaqmx.proto#L6195
```
message CreateLinScaleRequest {
  string name = 1;
  double slope = 2;
  double y_intercept = 3;
  oneof pre_scaled_units_enum {
    UnitsPreScaled pre_scaled_units = 4;
    int32 pre_scaled_units_raw = 5;
  }
  string scaled_units = 6;
}

message CreateLinScaleResponse {
  int32 status = 1;
}
```

nidaqmx/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,394 @@
...
+        self._task_handle = value
+
+    def get_task_handle(self):
+        return self._task_handle
+    
+    def set_system_handle(self, value=session_grpc_types.Session()):
+        self._system_handle = value
+
+    def get_system_handle(self):
+        return self._system_handle
```
bkeryan, line 31: System properties & functions do not use handles. They use device/physchan/terminal names or no names at all. I think the Device and PhysicalChannel classes (not the interpreter class) should store the device/physchan names.

```
/******************************************************/
/***                Signal Routing                  ***/
/******************************************************/


int32 __CFUNC     DAQmxConnectTerms              (const char sourceTerminal[], const char destinationTerminal[], int32 signalModifiers);
int32 __CFUNC     DAQmxDisconnectTerms           (const char sourceTerminal[], const char destinationTerminal[]);
int32 __CFUNC     DAQmxTristateOutputTerm        (const char outputTerminal[]);


/******************************************************/
/***                Device Control                  ***/
/******************************************************/


int32 __CFUNC     DAQmxResetDevice               (const char deviceName[]);

int32 __CFUNC     DAQmxSelfTestDevice            (const char deviceName[]);

int32 __CFUNC_C   DAQmxGetDeviceAttribute        (const char deviceName[], int32 attribute, void *value, ...);
```
 
```
message ConnectTermsRequest {
  string source_terminal = 1;
  string destination_terminal = 2;
  oneof signal_modifiers_enum {
    InvertPolarity signal_modifiers = 3;
    int32 signal_modifiers_raw = 4;
  }
}

message ConnectTermsResponse {
  int32 status = 1;
}
```

```
message GetDeviceAttributeBoolRequest {
  string device_name = 1;
  oneof attribute_enum {
    DeviceBoolAttribute attribute = 2;
    int32 attribute_raw = 3;
  }
}

message GetDeviceAttributeBoolResponse {
  int32 status = 1;
  bool value = 2;
}
```

nidaqmx/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,394 @@
...
+        self.session_initialization_behavior = grpc_options.initialization_behavior
+        self.set_task_handle()
+        self.set_system_handle()
+        self.set_scale_handle()
+
+    def set_task_handle(self, value=session_grpc_types.Session()):
+        self._task_handle = value
+
+    def get_task_handle(self):
+        return self._task_handle
```
bkeryan, line 25: This means that each task will have a separate interpreter (like nimi-python). I guess that's okay, as long as system/scale objects can use an interpreter that does not have a task handle.

nidaqmx/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,394 @@
...
+import nidaqmx.nidaqmx_pb2 as grpc_types
+import nidaqmx.nidaqmx_pb2_grpc as nidaqmx_grpc
+import nidaqmx.session_pb2 as session_grpc_types
+
+class GrpcStubInterpreter(object):
+    '''Interpreter for interacting with a gRPC Stub class'''
+
+    def __init__(self, grpc_options):
+        self._grpc_options = grpc_options
+        self._lock = threading.RLock()
```
bkeryan, line 14: nimi-python uses this to implement the IVI lock/unlock functions. DAQmx does not have an equivalent user-accessible lock, so delete this.

nidaqmx/_library_interpreter.py
bkeryan: FYI, _grpc_interpreter.py has `clear_task` not `close`

nidaqmx/grpc_session_options.py
```diff
@@ -0,0 +1,71 @@
+# -*- coding: utf-8 -*-
+# This file was generated
+
+from enum import IntEnum
+
+
+# This constant specifies the gRPC package and service used by this API.
+# Customers can pass this value to the MeasurementLink discovery service to resolve the server instance that provides this interface.
+GRPC_SERVICE_INTERFACE_NAME = 'nidcpower_grpc.NiDCPower'
```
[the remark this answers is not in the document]
subash-suresh: I actually copied and reused the grpc options class from dcpower, these values will change for the daqmx implementation.

nidaqmx/utils.py
```diff
@@ -13,7 +15,16 @@
     "the colon. Colons are not allowed within the names of the individual "
     "objects.")
 
+def select_interpreter(grpc_options = None, interpreter = None):
```
bkeryan, line 18: This is a public submodule containing public utils. Either add a leading underscore to select_interpreter to denote that it is private, or move it to a private module.

nidaqmx/system/system.py
```diff
@@ -34,13 +37,24 @@ class System(object):
...
     def local():
         """
         nidaqmx.system.system.System: Represents the local DAQmx system.
         """
-        return System()
+        interpreter = _library_interpreter.LibraryInterpreter(encoding='windows-1251')
+        return System(interpreter=interpreter)
+    
+    @staticmethod
+    def remote(grpc_options):
```
bkeryan, line 52: Would it make sense for grpc_options to be keyword-only here? I guess not, because it's currently not optional.

nidaqmx/system/system.py
```diff
@@ -34,13 +37,24 @@ class System(object):
...
+    def __init__(self, interpreter = None):
+        self._interpreter = interpreter
 
     @staticmethod
     def local():
         """
         nidaqmx.system.system.System: Represents the local DAQmx system.
         """
-        return System()
+        interpreter = _library_interpreter.LibraryInterpreter(encoding='windows-1251')
```
bkeryan, line 48: Where did you copy windows-1251 from? windows-1251 is the encoding for languages like Russian and Ukrainian that use Cyrillic script. windows-1252 is Latin 1, which is the encoding for English, French, and German (among others).

`nicaiu.dll` uses ANSI/MBCS (windows-1252, windows-932, etc.). Both `nicai_utf8.dll` and gRPC use UTF-8.

We should avoid hardcoding a specific encoding, if possible. However, modern Linux systems use UTF-8 by default, so if we are using `nicaiu.dll`, it may be necessary to explicitly specify a MBCS encoding for the library interpreter.

nidaqmx/tests/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+import nidaqmx
+from nidaqmx.grpc_session_options import GrpcSessionOptions
+import grpc
+
+@pytest.fixture(scope="module")
+def grpc_session_creation_kwargs():
+        channel = grpc.insecure_channel(f"localhost:31763")
+        grpc_options = GrpcSessionOptions(
+            grpc_channel = channel,
+            session_name = "NIDAQmxSession",
```
bkeryan, line 11: Some tests create multiple tasks. By specifying a hardcoded session name, they will end up sharing a task on the grpc-device server, which is incorrect.

nimi-python's tests leave the session name blank, so the grpc-device server will assign each task a unique session name. I think we should do that here.

nidaqmx/tests/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+        grpc_options = GrpcSessionOptions(channel, "")
+        return {'grpc_options': grpc_options}
+
+@pytest.fixture(scope="module")
+def library_session_creation_kwargs():
+    return {}
+
+@pytest.fixture(params = ("library_session_creation_kwargs", "grpc_session_creation_kwargs"), scope="module")
+def session_creation_kwargs(request):
+    return request.getfixturevalue(request.param)
```
bkeryan, line 22: Good, if this works as expected, I think this will be lower touch than the approach that nimi-python uses (of subclassing the test class for library vs. grpc)

nidaqmx/tests/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+    yield nidaqmx.Task(**init_args,**session_creation_kwargs)
+
+@pytest.fixture(scope="function")
+def system(session_creation_kwargs):
+    if session_creation_kwargs == {}:
+        return nidaqmx.System.local()
+    else:
+        return nidaqmx.System.remote(**session_creation_kwargs)
+
+# In this prototype i am doing it for generic scale object creation, but this can be replicated for the different scale object creations as well.
```
bkeryan, line 45: Yes, they have different purposes. Scale.create_lin_scale() etc. actually create a DAQmx scale, not just a Python object representing a scale that already exists.

You don't have to mock up the others in this PR; this comment is enough.

nidaqmx/tests/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.fixture(scope="function")
+def scale(request, session_creation_kwargs):
```
bkeryan, line 47: Bad news: scales don't have much test coverage. Scale.create_*_scale doesn't seem to have any test coverage.

```
PS C:\dev\nidaqmx-python> rg -i scale tests
tests\test_write_exceptions.py
52:            writer = nidaqmx.stream_writers.AnalogUnscaledWriter(write_task.out_stream, auto_start=False)
99:            writer = nidaqmx.stream_writers.AnalogUnscaledWriter(write_task.out_stream, auto_start=False)

tests\test_system_collections.py
12:from nidaqmx.system._collections.persisted_scale_collection import (
13:    PersistedScaleCollection)
35:    def test_persisted_scale_collection_property(self):
38:        scales = system.scales
39:        assert isinstance(scales, PersistedScaleCollection)
40:        assert isinstance(scales, collections.abc.Sequence)
42:        if len(scales) > 0:
43:            assert isinstance(scales[0], nidaqmx.system.storage.PersistedScale)
46:            assert isinstance(scales[0].author, six.string_types)

tests\test_stream_power_readers.py
23:    These tests use simulated TestScale PPS device(s), TS-15200.
102:    These tests use simulated TestScale PPS device(s), TS-15200.
196:    These tests use simulated TestScale PPS device(s), TS-15200.

tests\test_read_write.py
606:    These tests use simulated TestScale PPS device(s), TS-15200.

tests\test_channel_creation.py
37:                max_val=20.0, units=VoltageUnits.FROM_CUSTOM_SCALE,
38:                custom_scale_name="double_gain_scale")
46:                    VoltageUnits.FROM_CUSTOM_SCALE)
47:            assert (ai_channel.ai_custom_scale.name ==
48:                    "double_gain_scale")
86:    #             max_val=1.0, units=VoltageUnits.volts, custom_scale_name="")
93:    #         assert ai_channel.ai_custom_scale.name == ""
198:                current_excit_val=0.005, custom_scale_name="")
226:                custom_scale_name="")
257:                custom_scale_name="")

tests\max_config\nidaqmxMaxConfig.ini
5:[DAQmxScale double_gain_scale]
8:PreScaledUnits = Volts
9:ScaledUnits =
10:ScaleType = Linear
12:[DAQmxScale no_scaling_scale]
15:PreScaledUnits = Volts
16:ScaledUnits =
17:ScaleType = Linear
19:[DAQmxScale polynomial_scale]
22:PreScaledUnits = Volts
23:ScaledUnits =
24:ScaleType = Polynomial

tests\fixtures.py
85:        if device.dev_is_simulated and device.product_category == ProductCategory.TEST_SCALE_MODULE and UsageTypeAI.POWER in device.ai_meas_types:
89:        "Could not detect a device that meets the requirements to be a TestScale simulated power device. "
101:        if device.dev_is_simulated and device.product_category == ProductCategory.TEST_SCALE_MODULE and UsageTypeAI.VOLTAGE in device.ai_meas_types:
105:        "Could not detect a device that meets the requirements to be a TestScale simulated voltage device. "
118:        if device.dev_is_simulated and device.product_category == ProductCategory.TEST_SCALE_MODULE and UsageTypeAI.POWER in device.ai_meas_types:
124:        "Could not detect two or more devices that meets the requirements to be a TestScale simulated power "
```

Cc: @zhindes

nidaqmx/system/system.py
subash-suresh: Sure Brad, thanks for pointing that out, i reused the implementation from nimi python and this value is from there.

nidaqmx/_library_interpreter.py
```diff
@@ -0,0 +1,241 @@
...
+                        ctypes.c_int, ctypes.c_char_p]
+
+        error_code = cfunc(
+            self._task_handle, channel_name, attribute_id, attribute_value)
+        check_for_error(error_code)
+
+    def reset_channel_attribute(
+        self, channel_name, attribute_id):
+
+        cfunc = lib_importer.windll.DAQmxReSetChanAttribute
```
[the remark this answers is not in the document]
subash-suresh: updated

nidaqmx/system/system.py
bkeryan: There is an open issue about that: https://github.com/ni/nimi-python/issues/1820

nidaqmx/tests/conftest.py
zhindes: Oh joy...

nidaqmx/tests/conftest.py
subash-suresh: Thanks for confirming this.

nidaqmx/tests/conftest.py
subash-suresh: Yes ,during my testing this approached worked perfectly fine.

nidaqmx/system/system.py
subash-suresh: for system when the user is using the remote method, then the gRPC options becomes a required parameter right?

nidaqmx/_grpc_stub_interpreter.py
subash-suresh: That makes sense, i will remove this system handle local variable from the interpreters.

nidaqmx/_grpc_stub_interpreter.py
subash-suresh: I will remove the scale property from the interpreters as well.

nidaqmx/_library_interpreter.py
```diff
@@ -0,0 +1,241 @@
...
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        lib_importer.task_handle, ctypes_byte_str,
+                        ctypes.c_int, ctypes.POINTER(ctypes.c_bool)]
+
+        error_code = cfunc(
+            self._task_handle, channel_name, attribute_id, ctypes.byref(val))
+        check_for_error(error_code)
+
+        return bool(val.value)
```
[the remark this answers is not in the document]
subash-suresh: I think we decided to move on with the implementation of using the generic attributes, so, hence closing this comment.

nidaqmx/task.py
```diff
@@ -125,33 +124,7 @@ def name(self):
...
-            elif size_or_code > 0 and temp_size == 0:
-                # Buffer size obtained, use to retrieve data.
-                temp_size = size_or_code
-            else:
-                break
-
-        check_for_error(size_or_code)
-
-        return val.value.decode('ascii')
+        return self._interpreter.get_task_attribute_string(4726)
```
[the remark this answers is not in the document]
subash-suresh: The [nidaqmx.proto](https://github.com/ni/grpc-device/blob/0f2a60c8c06449ec83db2549a9c38ed6211d9a9e/generated/nidaqmx/nidaqmx.proto#L447) does define such enums, we could plan to create similar enums or reuse the ones from the proto.

nidaqmx/system/system.py
bkeryan: That's what makes sense today because remote() only supports gRPC.

If remote() supported other transports, such as HTTP or Thrift, then the answer would be different.

nidaqmx/task.py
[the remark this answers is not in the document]
bkeryan: We should avoid using the gRPC & protobuf codegen in non-gRPC code. I would rather not require non-gRPC users to install `grpcio` and `protobuf`.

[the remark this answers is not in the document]
zhindes: Agreed. It should be an optional install.

nidaqmx/tests/conftest.py
```diff
@@ -0,0 +1,59 @@
+import pytest
+import nidaqmx
+from nidaqmx.grpc_session_options import GrpcSessionOptions
+import grpc
+
+@pytest.fixture(scope="module")
+def grpc_session_creation_kwargs():
+        channel = grpc.insecure_channel(f"localhost:31763")
```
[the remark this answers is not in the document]
bkeryan: The fixture right below it, session_creation_kwargs, uses it.

PR #211, opened 2023-02-22
DAQmx Code Generator | PR1 | Adding a basic code generator

Sakthi-SM:
This PR introduces a basic code generator for the DAQmx files which involves the following changes:
1. Re-ordered the folder structure to introduce a code generator
2. Added a `codegenerator` module
3. Added a generated folder which contains the currently `generated` files and every time a new file is generated, correspondingly it will be updated in the `generated` folder.
4. Updated the `click` as dependency in the poetry files.
5.  This PR breakdown is as per the suggestion received in [suggestion](https://github.com/ni/nidaqmx-python/pull/208#discussion_r1112108379)


generated/tests/fixtures.py
```diff
@@ -0,0 +1,151 @@
+import pytest
```
bkeryan, line 1: I don't think we should copy the `tests/` directory to `generated/tests/`. We're not going to codegen tests.

generated/__init__.py
```diff
@@ -0,0 +1,13 @@
+from nidaqmx.errors import DaqError, DaqReadError, DaqWriteError, DaqWarning, DaqResourceWarning
```
bkeryan, line 1: This copies the non-generated files to the `generated/` directory as well. I assume that they will live in `src/` and the codegen or build process will automatically copy them over that that the `generated/` directory contains a fully-formed `nidaqmx` module.

bkeryan, line 1: Please move the files to `generated/nidaqmx/` so that the `generated/` directory contains a fully-formed `nidaqmx` module. We don't want to make `generated` part of the module hierarchy.

For comparison, nimi-python modules live in `generated/${module}/${module}/` like [generated/nidcpower/nidcpower/](https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/). The extra directory level is necessary to give each nimi-python module a separate `setup.py`. There is only one `nidaqmx` module in this repo, so this repo doesn't need the extra directory level.

src/codegen/__main__.py
```diff
@@ -0,0 +1,48 @@
...
+    logging.basicConfig(level=logging_level, format=log_format)
+
+    try:
+        generator.generate(dest)
+    except Exception:
+        _logger.exception("Failed to generate")
+        return 1
+
+
+if __name__ == "__main__":
```
bkeryan, line 47: Does the codegen script work if you delete your venv, create a clean venv with `poetry install`, and run the script with `poetry run`? What is the exact `poetry run` command to use?

src/codegen/__main__.py
```diff
@@ -0,0 +1,48 @@
+import logging
```
bkeryan, line 1: FYI: We do not want to include a `generated`, `codegen`, or `src` directory inside the `nidaqmx` package that we upload to PyPI. 

I ran `poetry build` and opened the resulting wheel with 7-Zip, and I don't see anything out of place.

Then I tried deleting the top-level `nidaqmx/`, moving `generated/*` to `generated/nidaqmx/*`, adding this to `pyproject.toml`, and running `poetry build`:
```
packages = [
    { include = "nidaqmx", from = "generated" }
]
```

The resulting wheel still has everything in the right place, so I think this directory structure will work once you move `generated/*` to `generated/nidaqmx/*`.

poetry.lock
```diff
@@ -26,7 +26,7 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 dev = ["cloudpickle", "coverage[toml] (>=5.0.2)", "furo", "hypothesis", "mypy", "pre-commit", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six", "sphinx", "sphinx-notfound-page", "zope.interface"]
 docs = ["furo", "sphinx", "sphinx-notfound-page", "zope.interface"]
 tests = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six", "zope.interface"]
-tests-no-zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six"]
+tests_no_zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six"]
```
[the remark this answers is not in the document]
zhindes: I've never thought about it before. I see we're upgrading poetry within NI, but that's not ready, yet. So I'd say 1.1.x for now. We can update this to 1.2 later.

generated/tests/fixtures.py
zhindes: Agree, let's keep those separate. In fact, I really dislike that the `tests/` folder is under `nidaqmx/` at the moment. So if you wanted to move it to the root now, I wouldn't mind.

generated/__init__.py
zhindes: Yup, `generated/nidaqmx` should be everything. Some part of the build/release process should copy from `src/nidaqmx` to `generated/nidaqmx` in addition to regenerating all the codegen.

zhindes:
Once this is in, I will want to do a new minor release of [nidaqmx](https://pypi.org/project/nidaqmx/), version 0.6.6. This will help me validate everything is functional and update anything in CONTRIBUTING, etc. I mention that just to say I don't expect this PR to get everything perfectly correct up front. I don't plan to do that until this PR is complete, so this doesn't have to be absolutely perfect. I can make a follow-up PR resolving any issues I find as well as updating documentation, etc.

FYI @bkeryan.

src/codegen/generator.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def generate(dest):
+    _logger.info(f"Generating files into {dest}")
+
+    os.makedirs(dest, exist_ok=True)
+
+    codegen_metadata = _get_metadata()
+
+    for info in codegen_metadata["script_info"]["modules"]:
```
zhindes, line 44: Ah, good, it looks like this gracefully does nothing at the moment. So I can update directions to say "run the generator" even if it doesn't do anything useful, yet.

zhindes (commented):
I think Brad covered everything! I'll follow along with the updates and give approval once all of that is addressed.

zhindes (changes requested):
Oh, I did think of something. Please update CHANGELOG.md to add a link to this PR as a "Merged Pull Request" and mention the refactoring under "Major Changes" with a one sentence description. Something like "Refactored repository structure for future code generation improvements".

generated/tests/fixtures.py
Sakthi-SM: I have moved `tests/` to the root folder and currently the test files are still referencing the `nidaqmx` module present externally and not the generated files. Once the code generation is complete we can take up the task updating the tests to make use of the generated files.

generated/__init__.py
Sakthi-SM: Yes, some part of the files which are hand written will be present in `src` and the generator will make sure of copying these files to the `generated` folder as part of the generation

Sakthi-SM: Moved the generated files to `generated/nidaqmx`

src/codegen/__main__.py
Sakthi-SM: Yes Brad, it works fine when I delete the venv and recreate them again. I use `poetry run python -m codegen --dest out` from the `src` folder.

generated/nidaqmx/tests/test_channel_creation.py
```diff
@@ -0,0 +1,292 @@
+from nidaqmx import constants
```
bkeryan, line 1: Delete this file.

bkeryan (commented):
Approved with suggestions:
- Delete the remaining `generated/nidaqmx/tests`

bkeryan (approved):

generated/tests/fixtures.py
bkeryan: I think changing the tests to reference the generated files is unnecessary. Once we update `pyproject.toml` to use the generated files, then `poetry install` will install the generated files as the `nidaqmx` module, and it will work just like when you install it from PyPI.

tests/test_channel_creation.py
```diff
@@ -0,0 +1,292 @@
+from nidaqmx import constants
```
zhindes, line 1: any idea why this wasn't picked up as a rename? can you fix that?

zhindes (changes requested):

tests/test_channel_creation.py
zhindes: Oh, I see brad commented on this. Let me resolve this and pile on there.

generated/nidaqmx/tests/test_channel_creation.py
zhindes: It exists in the root, so I think it just didn't get picked up as a rename for some reason?

zhindes (approved):

generated/nidaqmx/tests/test_channel_creation.py
bkeryan: There should not be any tests in the `generated/` directory.

zhindes:
I thought of something else... the pyproject.toml after these changes is still pulling from `/nidaqmx`. I think what we to do something like this:
1. move `nidaqmx/` to `src/nidaqmx`
2. move `poetry.lock`, `pyproject.toml`, and `tox.ini` to `src/` (note: after this change, those files still create a well-defined python package out of `src/nidaqmx`, but after we start generating stuff and delete files from `src/nidaqmx`, the package will be ill-formed)
3. update release instructions to indicate the files from 2 are copies to `generated` (in addition to `src/nidaqmx`), build the package out of `generated`, etc.

Does that sound right @bkeryan ? If so, I think we should do 1. and 2. as part of this PR, too.

generated/nidaqmx/tests/test_channel_creation.py
zhindes: D'oh, I totally spaced on the path!

bkeryan:
> I thought of something else... the pyproject.toml after these changes is still pulling from `/nidaqmx`. I think we do this:
> 
> 1. move `nidaqmx/` to `src/nidaqmx`
> 2. move `poetry.lock`, `pyproject.toml`, and `tox.ini` to `src/` (note: after this change, those files still create a well-defined python package out of `src/nidaqmx`, but after we start generating stuff and delete files from `src/nidaqmx`, the package will be ill-formed)
> 3. update release instructions to indicate the files from 2 are copies to `generated` (in addition to `src/nidaqmx`), build the package out of `generated`, etc.
> 
> Does that sound right @bkeryan ? If so, I think we should do 1. and 2. as part of this PR, too.

No, see https://github.com/ni/nidaqmx-python/pull/211#discussion_r1114682782

zhindes:
> We don't have to move `pyproject.toml`. We can move `nidaqmx/` to `src/nidaqmx/` and specify` packages = [{ include = "nidaqmx", from = "generated" }]` though, if you want.

Ah, right. Let's disregard the moving of `nidaqmx/` to `src/nidaqmx`. But I would like to have this PR include that change to `pyproject.toml`, though. Let's do that @Sakthi-SM !

PR #212, opened 2023-02-22
DAQmx Code Generator | PR2 | Generate AI_Channel.py

Sakthi-SM:
This PR adds the necessary `mako` template for generating the `ai_channel.py` and updates the generated file.
* Added the `mako` for generating attributes as `ai_channel.py` contains only attributes
* Added the latest `attribute_metadata` from scrapigen
* Added a structure called `attributes` to store the data in codegen and use them in the `mako` files.
* Updated the generated folder with the latest `ai_channel.py`
* Comparing the generated files with the existing files, the difference is mostly observed on the name of the properties. I have listed down such changes observed in the currently generated file.

Current Name | Changed To
-- | --
ai_custom_scale | ai_custom_scale_name
ai_eddy_current_prox_sensitivity | ai_eddy_current_prox_probe_sensitivity
ai_eddy_current_prox_sensitivity_units | ai_eddy_current_prox_probe_sensitivity_units
ai_eddy_current_prox_units | ai_eddy_current_prox_probe_units
ai_rosette_strain_gage_gage_orientation | ai_rosette_strain_gage_orientation
ai_rtd_r_0 | ai_rtd_r0
ai_sound_pressure_b_ref | ai_sound_pressure_db_ref
ai_strain_force_read_from_chan | ai_strain_gage_force_read_from_chan
ai_thrmstr_r1 | ai_thrmstr_r_1
ai_acceld_b_ref | ai_acceld_db_ref
ai_voltaged_b_ref | ai_voltaged_db_ref
ai_teds_is_teds | ai_is_teds
ai_velocity_iepe_sensord_b_ref | ai_velocity_iepe_sensord_db_ref

</body></html></div>
</body>
</html>


generated/_task_modules/channels/ai_channel.py
```diff
@@ -20,7 +20,7 @@
     EddyCurrentProxProbeSensitivityUnits, ExcitationDCorAC,
     ExcitationIdleOutputBehavior, ExcitationSource,
     ExcitationVoltageOrCurrent, FilterResponse, FilterType,
-    ForceIEPESensorSensitivityUnits, ForceUnits, FrequencyUnits, Impedance1,
```
bkeryan, line 23: `Impedance1` is no longer imported, which suggests that the properties that use it were deleted or changed.

generated/_task_modules/channels/ai_channel.py
```diff
@@ -20,7 +20,7 @@
     EddyCurrentProxProbeSensitivityUnits, ExcitationDCorAC,
```
bkeryan, line 20: This is pretty close to the original codegen, but I don't think we should consider it done until there are no unexpected changes.

I flagged several sources of unexpected changes, but I didn't review the diff thoroughly because there were so many unexpected changes. Once you fix the ones I flagged, we may find others.

generated/_task_modules/channels/ai_channel.py
```diff
@@ -4997,11 +5129,10 @@ def ai_impedance(self):
             self._handle, self._name, ctypes.byref(val))
         check_for_error(error_code)
 
-        return Impedance1(val.value)
```
bkeryan, line 5000: This property used to return an `Impedance1` enum, but now it returns an `int`.

bkeryan (changes requested):

generated/_task_modules/channels/ai_channel.py
```diff
@@ -297,6 +302,8 @@ def ai_accel_4_wire_dc_voltage_sensitivity_units(self):
             self._handle, self._name)
         check_for_error(error_code)
 
+
+
     @property
     def ai_accel_charge_sensitivity(self):
```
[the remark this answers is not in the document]
zhindes: Yep, I was gonna make the same comment. Fixing that up will make this diff quite a bit smaller.

generated/_task_modules/channels/ai_channel.py
```diff
@@ -549,61 +564,13 @@ def ai_accel_units(self):
             self._handle, self._name)
         check_for_error(error_code)
 
-    @property
-    def ai_acceld_b_ref(self):
```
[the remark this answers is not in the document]
zhindes: Ugh, the original name was horribly mangled since dB makes camelCasing confused. I'm OK if this changes as long as the new one is correct. I think it should be ai_accel_db_ref. I see it in `ATTRIBUTES_BLACKLIST`, but I forget what that does.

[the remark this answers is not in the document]
bkeryan: BTW, "dB" tends to confuse the camel-to-snake-case conversions, so `ai_acceld_b_ref` may not be the name we really want in the end.

Cc: @zhindes @maxxboehme

generated/_task_modules/channels/ai_channel.py
```diff
@@ -2618,11 +2657,13 @@ def ai_current_units(self):
             self._handle, self._name)
         check_for_error(error_code)
 
+
+
     @property
-    def ai_custom_scale(self):
```
zhindes, line 2622: It looks like there was a special case for this property since we aren't setting the name in python, we are setting an instance of a class (but we only need the name). Definitely need to add a special case for this in the generator somehow.

src/codegen/utilities/attribute_helpers.py
```diff
@@ -0,0 +1,88 @@
+import logging
+from codegen.properties.attribute import Attribute
+
+_logger = logging.getLogger(__name__)
+_logger.addHandler(logging.NullHandler())
+
+ATTRIBUTES_BLACKLIST = [
+    "AI_STRAIN_GAGE_FORCE_READ_FROM_CHAN" "AI_SOUND_PRESSURE_DB_REF",
```
zhindes, line 8: I think you meant to make this two items in the list?

generated/_task_modules/channels/ai_channel.py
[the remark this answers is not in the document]
bkeryan: I think it changed to `ai_sound_pressure_db_ref`. The `_db_ref` part is fine but I'm not sure about the change from "accel" to "sound_pressure".

```
PS C:\dev\nidaqmx-python> rg db_ref
generated\_task_modules\channels\ai_channel.py
7733:    def ai_sound_pressure_db_ref(self):
7757:    @ai_sound_pressure_db_ref.setter
7758:    def ai_sound_pressure_db_ref(self, val):
7771:    @ai_sound_pressure_db_ref.deleter
7772:    def ai_sound_pressure_db_ref(self):
9143:    def ai_voltage_db_ref(self):
9165:    @ai_voltage_db_ref.setter
9166:    def ai_voltage_db_ref(self, val):
9179:    @ai_voltage_db_ref.deleter
9180:    def ai_voltage_db_ref(self):
```

zhindes (changes requested):

generated/_task_modules/channels/ai_channel.py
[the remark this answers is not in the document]
zhindes: It looks like both `accel` and `sound_pressure` should exist: https://www.ni.com/docs/en-US/bundle/ni-daqmx-properties/page/daqmxprop/attr29b2.html and https://www.ni.com/docs/en-US/bundle/ni-daqmx-properties/page/daqmxprop/attr29b1.html. I think this one is getting blacklisted, and sound pressure is getting handled correctly.

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -398,6 +397,56 @@ def ai_accel_charge_sensitivity_units(self):
             self._handle, self._name)
         check_for_error(error_code)
 
+    @property
+    def ai_accel_db_ref(self):
```
bkeryan, line 401: I think the `_db_ref` rename is good. 

However, I think we should consider deprecating the old names rather than just removing them. One way to do this is by adding hardcoded property definitions to the bottom of `ai_channel.py.mako`. 

```
%for old_property, new_property in DEPRECATED_PROPERTIES:
    @deprecation.deprecated
    @property
    def ${old_property}(self):
        return self.${new_property}

    @deprecation.deprecated
    @${old_property}.setter
    def ${old_property}(self, val):
        self.${new_property} = val

    @deprecation.deprecated
    @${old_property}.deleter
    def ${old_property}(self): 
        del self.${new_property}

%endfor
```

Another way is to directly instantiate properties (assuming this works). I'm not sure how to make this generate a `DeprecationWarning`.
```
%for old_property, new_property in DEPRECATED_PROPERTIES:
    ${old_property} = property(${new_property}.fset, ${new_property}.fget, ${new_property}.fdel)
%endfor    
```

Yet another way is to override `__getattr__` etc. and do it at run time (generate `DeprecationWarning` and redirect to the new proeprty). This is more complicated, but it hides the old properties from documentation, PyCharm, VS Code, etc.

@zhindes @maxxboehme @mshafer-NI Any thoughts?

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -2619,7 +2618,7 @@ def ai_current_units(self):
         check_for_error(error_code)
 
     @property
-    def ai_custom_scale(self):
+    def ai_custom_scale_name(self):
```
bkeryan, line 2621: I think we need to figure out how/where the old generator renamed these properties and add it into grpc-device-scrapigen, so that we fix up both the property names and any references to them in docstrings.

generated/nidaqmx/_task_modules/channels/ai_channel.py
zhindes: I like your recommendation of adding a `deprecation` dependency and keeping those old mangled names.

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -20,7 +20,7 @@
     EddyCurrentProxProbeSensitivityUnits, ExcitationDCorAC,
```
zhindes, line 20: For each of these PRs, I would expect the existing source file to be deleted. So in this case `nidaqmx/_task_modules/channels/ai_channel.py`.

generated/nidaqmx/_task_modules/channels/ai_channel.py
zhindes: I agree. This needs to be fixed to `ai_custom_scale`. I think custom logic in the generator, similar to `EXCLUDED_ATTRIBUTES`, that has a list of things we manually rename for Python is fine.

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -3579,7 +3578,7 @@ def ai_dither_enable(self):
         check_for_error(error_code)
 
     @property
-    def ai_eddy_current_prox_sensitivity(self):
+    def ai_eddy_current_prox_probe_sensitivity(self):
```
zhindes, line 3581: The old short names matches the LV short name. The new short name matches the C name. So there isn't a right answer here. I prefer this new name, but I really like the deprecation idea @bkeryan suggested, so let's add the old one as deprecated.

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -5253,6 +5250,28 @@ def ai_input_src(self):
             self._handle, self._name)
         check_for_error(error_code)
 
+    @property
+    def ai_is_teds(self):
```
zhindes, line 5254: this rename is good, let's add a deprecation

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -6733,7 +6752,7 @@ def ai_rng_low(self):
         check_for_error(error_code)
 
     @property
-    def ai_rosette_strain_gage_gage_orientation(self):
+    def ai_rosette_strain_gage_orientation(self):
```
zhindes, line 6755: Same story here as eddy current. The old name was what LV uses (it has periods to break up the double gage). I like the new name, but lets add a deprecation.

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -7041,7 +7060,7 @@ def ai_rtd_c(self):
         check_for_error(error_code)
 
     @property
-    def ai_rtd_r_0(self):
+    def ai_rtd_r0(self):
```
zhindes, line 7063: good change, lets deprecate the old one

zhindes (changes requested):
I stopped commenting on the few name changes that happened. I think all of them are good, but I love the idea of the deprecation. Let's do that :)

Also, your PR description has some typos...
* `ai_thrmstr_r_1` changed to `ai_thrmstr_r1`
* `ai_acceld_b_ref` changed to `ai_accel_db_ref`
* `ai_voltaged_b_ref` changed to `ai_voltage_db_ref`
* `ai_velocity_iepe_sensord_b_ref` changed to `ai_velocity_iepe_sensor_db_ref`

zhindes (changes requested):
In #211, we are going to update the `pyproject.toml` to specify `packages = [{ include = "nidaqmx", from = "src" }]`. Since this PR is the first change generating something, let's update it to `from = "generated"` as part of this change!

generated/nidaqmx/_task_modules/channels/ai_channel.py
zhindes: OK, disregard this. See below.

zhindes:
> In #211, we are going to update the `pyproject.toml` to specify `packages = [{ include = "nidaqmx", from = "src" }]`. Since this PR is the first change generating something, let's update it to `from = "generated"` as part of this change!

@bkeryan and I talked offline, and we'll wait on pulling the `from = "generated"` change until there is more substantial set of codegen available.

bkeryan:
> > In #211, we are going to update the `pyproject.toml` to specify `packages = [{ include = "nidaqmx", from = "src" }]`. Since this PR is the first change generating something, let's update it to `from = "generated"` as part of this change!
> 
> @bkeryan and I talked offline, and we'll wait on pulling the `from = "generated"` change until there is more substantial set of codegen available.

Also wait on deleting the old `ai_channel.py`

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -4980,8 +4981,7 @@ def ai_gain(self):
     @property
     def ai_impedance(self):
         """
-        :class:`nidaqmx.constants.Impedance1`: Specifies the input
-            impedance of the channel.
+        Impedance1: Specifies the input impedance of the channel.
```
zhindes, line 4984: Something strange is going on here. Impedance1 isn't a type, but [nidaqmx.constants.Impedance1]https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/constants.py#L383) is. Also the import disappeared at the top of this file, too.

zhindes (changes requested):
One minor issue I think is all that's left for me!

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -9373,3 +9372,129 @@ def pwr_voltage_setpoint(self):
...
+
+        @deprecation.deprecated
+        @property
+        def ai_eddy_current_prox_sensitivity(self):
+            return self.ai_eddy_current_prox_probe_sensitivity
+
+        @deprecation.deprecated
+        @ai_eddy_current_prox_sensitivity.setter
+        def ai_eddy_current_prox_sensitivity(self, val):
+            self.ai_eddy_current_prox_probe_sensitivity = val
```
bkeryan, line 9384: Did you test these deprecated properties? None of them work. They have an extra level of indentation, which causes them to be defined inside the body of `pwr_voltage_setpoint`.

```
PS C:\dev\nidaqmx-python> poetry1.2 run python
Python 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:58:18) [MSC v.1900 64 bit (AMD64)] on win32
>>> import nidaqmx
>>> help(nidaqmx._task_modules.channels.ai_channel.AIChannel.ai_strain_gage_force_read_from_chan)
Help on property:

    bool: Specifies whether the data is returned by DAQmx Read when
        set on a raw strain channel that is part of a rosette
        configuration.

>>> help(nidaqmx._task_modules.channels.ai_channel.AIChannel.ai_strain_force_read_from_chan)
AttributeError: type object 'AIChannel' has no attribute 'ai_strain_force_read_from_chan'
```

If I fix that, there are more problems.
```
>>> import nidaqmx
  File "C:\Dev\nidaqmx-python\src\nidaqmx\__init__.py", line 3, in <module>
    from nidaqmx.task import Task
  File "C:\Dev\nidaqmx-python\src\nidaqmx\task.py", line 7, in <module>
    from nidaqmx._task_modules.channels.channel import Channel
  File "C:\Dev\nidaqmx-python\src\nidaqmx\_task_modules\channels\__init__.py", line 6, in <module>
    from nidaqmx._task_modules.channels.ai_channel import AIChannel
  File "C:\Dev\nidaqmx-python\src\nidaqmx\_task_modules\channels\ai_channel.py", line 35, in <module>
    class AIChannel(Channel):
  File "C:\Dev\nidaqmx-python\src\nidaqmx\_task_modules\channels\ai_channel.py", line 9382, in AIChannel
    @ai_eddy_current_prox_sensitivity.setter
AttributeError: 'function' object has no attribute 'setter'
```

I think `@deprecated` needs to be after `@property` etc.
```
    @property
    @deprecation.deprecated(deprecated_in="0.6", current_version=__version__, details="Use ai_strain_gage_force_read_from_chan instead.")
    def ai_strain_force_read_from_chan(self):
        return self.ai_strain_gage_force_read_from_chan

    @ai_strain_force_read_from_chan.setter
    @deprecation.deprecated(deprecated_in="0.6", current_version=__version__, details="Use ai_strain_gage_force_read_from_chan instead.")
    def ai_strain_force_read_from_chan(self, val):
        self.ai_strain_gage_force_read_from_chan = val

    @ai_strain_force_read_from_chan.deleter
    @deprecation.deprecated(deprecated_in="0.6", current_version=__version__, details="Use ai_strain_gage_force_read_from_chan instead.")
    def ai_strain_force_read_from_chan(self):
        del self.ai_strain_gage_force_read_from_chan
```

src/codegen/templates/property_deprecated_template.py.mako
```diff
@@ -0,0 +1,27 @@
+<%def name="script_deprecated_property(attributes)">
+<%
+        from codegen.utilities.attribute_helpers import get_deprecated_attributes
+        deprecated_attributes = get_deprecated_attributes(attributes)
+    %>\
+%for old_property, attribute in deprecated_attributes.items():
+        @deprecation.deprecated
+        @property
+        def ${old_property}(self):
```
bkeryan, line 9: `deprecated` adds a deprecation message to the end of the function's docstring. If we were deprecating hand-written functions, we would leave the old docstrings alone, but since we're doing codegen, I don't think it's worth the effort to generate the old docstrings.

I tried it out:
```
    @property
    @deprecation.deprecated(deprecated_in="0.6", current_version=__version__, details="Use ai_strain_gage_force_read_from_chan instead.")
    def ai_strain_force_read_from_chan(self):
        return self.ai_strain_gage_force_read_from_chan

    @ai_strain_force_read_from_chan.setter
    @deprecation.deprecated(deprecated_in="0.6", current_version=__version__, details="Use ai_strain_gage_force_read_from_chan instead.")
    def ai_strain_force_read_from_chan(self, val):
        self.ai_strain_gage_force_read_from_chan = val

    @ai_strain_force_read_from_chan.deleter
    @deprecation.deprecated(deprecated_in="0.6", current_version=__version__, details="Use ai_strain_gage_force_read_from_chan instead.")
    def ai_strain_force_read_from_chan(self):
        del self.ai_strain_gage_force_read_from_chan
```

Here's the documentation output:
```
...
|  ai_sound_pressure_units
 |      :class:`nidaqmx.constants.SoundPressureUnits`: Specifies the
 |          units to use to return sound pressure measurements from the
 |          channel.
 |
 |  ai_strain_force_read_from_chan
 |      .. deprecated:: 0.6 Use ai_strain_gage_force_read_from_chan instead.
 |
 |  ai_strain_gage_cfg
 |      :class:`nidaqmx.constants.StrainGageBridgeType`: Specifies the
 |          bridge configuration of the strain gages.
 |
 |  ai_strain_gage_force_read_from_chan
 |      bool: Specifies whether the data is returned by DAQmx Read when
 |          set on a raw strain channel that is part of a rosette
 |          configuration.
 |
 |  ai_strain_gage_gage_factor
 |      float: Specifies the sensitivity of the strain gage.  Gage
 |          factor relates the change in electrical resistance to the
 |          change in strain. Refer to the sensor documentation for this
 |          value.
...
```

It would also be possible for the template to directly generate this docstring, but having `deprecated` do it causes this information to be included in any deprecation warnings (if they are enabled).

bkeryan (changes requested):

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -4997,11 +4997,10 @@ def ai_impedance(self):
             self._handle, self._name, ctypes.byref(val))
         check_for_error(error_code)
 
-        return Impedance1(val.value)
+        return val.value
 
     @ai_impedance.setter
     def ai_impedance(self, val):
-        val = val.value
         cfunc = lib_importer.windll.DAQmxSetAIImpedance
```
bkeryan, line 5004: In addition to @zhindes 's last comment, the line that dereferences val.value is gone.

bkeryan (changes requested):

src/codegen/templates/_task_modules/channels/ai_channel.py.mako
```diff
@@ -0,0 +1,40 @@
...
+    lib_importer, wrapped_ndpointer, ctypes_byte_str, c_bool32)
+from nidaqmx.scale import Scale
+from nidaqmx.errors import (
+    check_for_error, is_string_buffer_too_small, is_array_buffer_too_small)
+from nidaqmx._task_modules.channels.channel import Channel
+from nidaqmx.utils import unflatten_channel_string
+from nidaqmx.constants import (
+    ${', '.join([c for c in enums_used]) | wrap(4, 4)})
+
+__version__ = __init__.__version__
```
mshafer-NI, line 24: `__version__` is not required. and copying the value from `__init__` seems like overkill.

Can you help me understand why we're doing this?

zhindes (approved):

src/codegen/templates/_task_modules/channels/ai_channel.py.mako
bkeryan: We're doing this to pass the current version to the `deprecation` module. Perhaps we shouldn't bother? I think it will still print the `deprecated_in` version without `current_version`.

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -2,6 +2,8 @@
 
 import ctypes
 import numpy
+import deprecation
+import __init__
```
bkeryan, line 6: Did you test this? It doesn't work on my machine. 

`import __init__` imports a global module named `__init__`, not a submodule. I think it's loading `nidaqmx_examples/__init__.py`.
```
PS Z:\nidaqmx-python> poetry run python .\nidaqmx_examples\ai_voltage_sw_timed.py
  File ".\nidaqmx_examples\ai_voltage_sw_timed.py", line 2, in <module>
    import nidaqmx
  File "\\xidax-xyzzy\dev\nidaqmx-python\src\nidaqmx\__init__.py", line 3, in <module>
    from nidaqmx.task import Task
  File "\\xidax-xyzzy\dev\nidaqmx-python\src\nidaqmx\task.py", line 7, in <module>
    from nidaqmx._task_modules.channels.channel import Channel
  File "\\xidax-xyzzy\dev\nidaqmx-python\src\nidaqmx\_task_modules\channels\__init__.py", line 12, in <module>
    from nidaqmx._task_modules.channels.ai_channel import AIChannel
  File "\\xidax-xyzzy\dev\nidaqmx-python\src\nidaqmx\_task_modules\channels\ai_channel.py", line 36, in <module>
    __version__ = __init__.__version__
AttributeError: module '__init__' has no attribute '__version__'
```

If I use `from . import __init__`, it fails in a different way:

```
PS Z:\nidaqmx-python> poetry run python .\nidaqmx_examples\ai_voltage_sw_timed.py
  File ".\nidaqmx_examples\ai_voltage_sw_timed.py", line 2, in <module>
    import nidaqmx
  File "\\xidax-xyzzy\dev\nidaqmx-python\src\nidaqmx\__init__.py", line 3, in <module>
    from nidaqmx.task import Task
  File "\\xidax-xyzzy\dev\nidaqmx-python\src\nidaqmx\task.py", line 7, in <module>
    from nidaqmx._task_modules.channels.channel import Channel
  File "\\xidax-xyzzy\dev\nidaqmx-python\src\nidaqmx\_task_modules\channels\__init__.py", line 12, in <module>
    from nidaqmx._task_modules.channels.ai_channel import AIChannel
  File "\\xidax-xyzzy\dev\nidaqmx-python\src\nidaqmx\_task_modules\channels\ai_channel.py", line 36, in <module>
    __version__ = __init__.__version__
AttributeError: 'method-wrapper' object has no attribute '__version__'
```

https://stackoverflow.com/questions/50381749/python-3-how-do-i-import-init-py-from-a-module-in-its-package sort of explains what's going on here. It suggests a workaround, which seems to work:
```
from .. import channels

__version__ = channels.__version__
```

However, I think this comes very close to introducing a circular dependency, which seems brittle. (Consider why `import nidaqmx` in a submodule doesn't allow the submodule to use `nidaqmx.__version__`.)

Since we're only using `deprecated_in` right now, not `removed_in`, I think it would be simpler to remove `current_version=`.

src/codegen/templates/_task_modules/channels/ai_channel.py.mako
bkeryan: Getting the version from `__init__` isn't working anyway. I think we would need to introduce a new submodule that is not named `__init__.py` to make this work.

Since we're only using `deprecated_in` right now, not `removed_in`, I think it would be simpler to remove `current_version=`.

bkeryan (changes requested):

generated/nidaqmx/_task_modules/channels/__init__.py
```diff
@@ -1,3 +1,9 @@
+try:
+    from importlib.metadata import version
+except ImportError:
+    from importlib_metadata import version
+
+__version__ = version("nidaqmx")
```
bkeryan, line 6: revert

generated/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -9373,3 +9375,173 @@ def pwr_voltage_setpoint(self):
             self._handle, self._name)
         check_for_error(error_code)
 
+
+    @property
+    @deprecation.deprecated(deprecated_in="0.6.6", details="Use ai_eddy_current_prox_probe_sensitivity instead.")
```
bkeryan, line 9380: The deprecation behavior looks good.

```
with nidaqmx.Task() as task:
    task.ai_channels.add_ai_voltage_chan("Dev1/ai0")
    print(task.ai_channels[0].ai_acceld_b_ref)
```

```
PS Z:\nidaqmx-python> poetry run python .\nidaqmx_examples\ai_voltage_sw_timed.py
.\nidaqmx_examples\ai_voltage_sw_timed.py:9: DeprecatedWarning: ai_acceld_b_ref is deprecated as of 0.6.6. Use ai_accel_db_ref instead.
```

bkeryan (approved):
Approved with suggestions:
- Revert `__init__.py`

PR #214, opened 2023-03-01
Add actions to build and run tests in DAQmx module

Sakthi-SM:
This PR adds actions to the DAQmx module that builds the module and verifies tests.
* Created `workflows` folder in `.github` and added `main.yaml` file
* Fixed the `lint` errors in the tests, examples and codegen


zhindes:
@Sakthi-SM  - Like I mentioned elsewhere all existing tests require DAQmx to be installed. Once gRPC support is integrated there will be new tests that don't require that. We've already tackled a similar problem. A subset of the tests require physical hardware. We made a [fixture](https://github.com/ni/nidaqmx-python/blob/master/tests/fixtures.py#L21) that detects if that hardware is available and skips otherwise. We should do a similar thing here. Create a fixture that detects if DAQmx is installed and skips otherwise. So for a while running the tests in a workflow will be sort of useless, but over time we can improve that.

zhindes:
Overall, I really like this. Once we skip all the tests that require DAQmx if it is isn't installed (which is nearly all of them), then I think we're good. I also really like the cleanup of the source by introducing the NI Style Guide. Woo!!!

pyproject.toml
```diff
@@ -56,6 +56,8 @@ tox = "^3.24"
 click = "^8.1"
 Mako = "^1.2"
 deprecation = "^2.1"
+ni-python-styleguide = "^0.3"
```
bkeryan, line 59: The current version of ni-python-styleguide is 0.4: https://github.com/ni/python-styleguide/releases

pyproject.toml
```diff
@@ -38,11 +38,11 @@ numpy = [
...
 six = "^1.16"
 # This functionality was merged into python stdlib beginning in 3.8
-importlib_metadata = {version="^4.10", python="~3.7"}
+importlib_metadata = {version="4.2", python="~3.7"}
 # Documentation, must be in main dependencies (but optional) list for
 # readthedocs integration.
 # See https://github.com/readthedocs/readthedocs.org/issues/4912.
-Sphinx = {version="^4.4", optional=true}
+Sphinx = {version="^4.3", optional=true}
 sphinx_rtd_theme = {version="^1.0", optional=true}
```
bkeryan, line 46: Why did you downgrade these?

pyproject.toml
```diff
@@ -56,6 +56,8 @@ tox = "^3.24"
 click = "^8.1"
 Mako = "^1.2"
 deprecation = "^2.1"
+ni-python-styleguide = "^0.3"
+black = "^23.1.0"
```
bkeryan, line 60: I don't think we need to explicitly specify a dependency on `black`. `ni-python-styleguide` depends on `black`, so it will be installed.

@mshafer-NI, please correct me if I'm wrong.

pyproject.toml
```diff
@@ -56,6 +56,8 @@ tox = "^3.24"
...
```
bkeryan, line 61: Please add 
```
[tool.black]
line-length = 100
```

src/nidaqmx/_task_modules/channels/ai_channel.py
```diff
@@ -3,43 +3,90 @@
 import ctypes
```
bkeryan, line 3: Reformatting *everything* is overdoing it. Please specify exclusions in the `[tool.black]` section of `pyproject.toml`:
- All code-generated files (both in src/ and generated/). You might need to list out individidual files.
- All metadata files.

Once the new codegen is done, we can _consider_ using `black` to reformat the output, but until then we shouldn't bother because it will just make it harder to compare codegen.

.github/workflows/main.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install poetry
```
bkeryan, line 22: There is a setup-poetry action: https://github.com/Gr1N/setup-poetry

We have been successfully using it for measurementlink-python: https://github.com/ni/measurementlink-python/blob/main/.github/workflows/Publish_NIMS.yml#L23

.github/workflows/main.yml
```diff
@@ -0,0 +1,34 @@
...
+          poetry install
+          pip install flake8 pytest
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
bkeryan, line 25: Don't do this. It doesn't install into the right venv and it doesn't use the versions listed in poetry.lock.

ni-python-styleguide already installs flake8.

pyproject.toml already specifies pytest.

This repo doesn't have a requirements.txt.

.github/workflows/main.yml
```diff
@@ -0,0 +1,34 @@
+name: NIDAQmx package
```
bkeryan, line 1: Rename this file to PR.yml. We will probably want a separate workflow for publishing, like in https://github.com/ni/measurementlink-python/tree/main/.github/workflows

bkeryan (changes requested):

bkeryan:
> @Sakthi-SM - Like I mentioned elsewhere all existing tests require DAQmx to be installed. Once gRPC support is integrated there will be new tests that don't require that. 

Once gRPC support is integrated, there will be new tests that require both DAQmx and NI gRPC Device Server :)

We aren't planning to write any tests that use a mock NI gRPC Device Server or mock gRPC channel.

.github/workflows/PR.yml
```diff
@@ -0,0 +1,36 @@
+name: NIDAQmx package
+
+on: [push]
```
bkeryan, line 3: In addition to naming this action "PR", it should execute when someone creates or updates a PR:
```
on:
  pull_request:
    branches:
      - main
```

If the tests end up taking too long to run for every PR, we could also create a "CI" action that executes when someone pushes to main. However, I don't think we need this yet.

BTW, this repo still uses "master" not "main".
Cc: @zhindes

bkeryan (changes requested):

.github/workflows/PR.yml
zhindes: Yeah, sorry about that. I intend to change it at some point, but no concrete plans at the moment.

.github/workflows/PR.yml
```diff
@@ -26,10 +26,11 @@ jobs:
...
         run: |
-          poetry run ni-python-styleguide lint ./generated/
-          poetry run ni-python-styleguide lint ./nidaqmx_examples/
-          poetry run ni-python-styleguide lint ./src/
-          poetry run ni-python-styleguide lint ./tests/
+          poetry run ni-python-styleguide lint ./src/codegen/properties
+          poetry run ni-python-styleguide lint ./src/codegen/utilities
+          poetry run ni-python-styleguide lint ./src/codegen/__init__.py
+          poetry run ni-python-styleguide lint ./src/codegen/__main__.py
+          poetry run ni-python-styleguide lint ./src/codegen/generator.py
```
bkeryan, line 33: Why not lint all of ./src/codegen? Is there something in ./src/codegen that you are trying to avoid linting?

zhindes: This question is still relevant

pyproject.toml
zhindes: This question is still relevant

zhindes (changes requested):
Brad's two open conversations are still relevant. I don't see anything else new.

.github/workflows/PR.yml
bkeryan: You can exclude the metadata:
`poetry run ni-python-styleguide lint ./src/codegen --exclude ./src/codegen/metadata`

mshafer-NI: better yet, put this in the pyproj file:
```toml
[tool.ni-python-styleguide]
extend_exclude = "./src/codegen/metadata"
```
that way, dev's don't have to use the param and can just "poetry run ni-python-styleguide lint"

pyproject.toml
bkeryan: To answer my own question, if we don't downgrade these, `poetry lock` fails with:

```
  SolverProblemError

  Because no versions of flake8 match >5.0,<5.0.1 || >5.0.1,<5.0.2 || >5.0.2,<5.0.3 || >5.0.3,<5.0.4 || >5.0.4,<6.0
   and flake8 (5.0.0) depends on importlib-metadata (<4.3), flake8 (>=5.0,<5.0.1 || >5.0.1,<5.0.2 || >5.0.2,<5.0.3 || >5.0.3,<5.0.4 || >5.0.4,<6.0) requires importlib-metadata (<4.3).
  And because flake8 (5.0.1) depends on importlib-metadata (<4.3)
   and flake8 (5.0.2) depends on importlib-metadata (<4.3), flake8 (>=5.0,<5.0.3 || >5.0.3,<5.0.4 || >5.0.4,<6.0) requires importlib-metadata (<4.3).
  And because flake8 (5.0.3) depends on importlib-metadata (<4.3)
   and flake8 (5.0.4) depends on importlib-metadata (>=1.1.0,<4.3), flake8 (>=5.0,<6.0) requires importlib-metadata (<4.3).
  Because no versions of ni-python-styleguide match >0.4,<0.5
   and ni-python-styleguide (0.4.0) depends on flake8 (>=5.0,<6.0), ni-python-styleguide (>=0.4,<0.5) requires flake8 (>=5.0,<6.0).
  Thus, ni-python-styleguide (>=0.4,<0.5) requires importlib-metadata (<4.3).
  So, because nidaqmx depends on both importlib_metadata (^4.10) and ni-python-styleguide (^0.4), version solving failed.
```

If we do downgrade these, `poetry lock` hangs forever on
```
   2: selecting sphinxcontrib.jquery (2.0.0)
   2: selecting sphinxcontrib.jquery (2.0.0)
   2: selecting sphinxcontrib.jquery (2.0.0)
   2: selecting sphinxcontrib.jquery (2.0.0)
   2: selecting sphinxcontrib.jquery (2.0.0)
   2: selecting sphinxcontrib.jquery (2.0.0)
   2: selecting sphinxcontrib.jquery (2.0.0)
```

Cc: @mshafer-NI

mshafer-NI: Options:
1. keep the downgrade till June (when Python 3.7 is EOL and can officially be dropped)
1. Move sphinx (and related) to a separate env that doesn't participate in the lock (this is what we're doing with a lot of modules internally)
1. Drop 3.7 support early (the conflict on importlib-metada is a <3.8 backport dep)

I would suggest number 1 as these constraints don't affect customers, but dropping 3.7 woulld.

bkeryan: Okay, option 1 sounds fine to me.

Regarding `poetry lock` hanging:
- The `sphinxcontrib.jquery` hang seems to be related to upgrading `sphinx_rtd_theme` from 1.0 to 1.1 or 1.2.
- Poetry spends a few minutes downloading multiple versions of `numpy` to figure out that Python 3.7 needs version 1.21.6 but Python 3.8+ can use version 1.24.2. I guess this is acceptable because it finishes in less than 5 minutes.
- 
These are dependencies of nidaqmx-python, not just dev-dependencies, so specifying a version constraint would affect customers. @mshafer-NI How should we handle this?

@Sakthi-SM Did `poetry lock` hang when you ran it? How long did it take?

mshafer-NI: For numpy, you can speed that up by telling poetry (and pip via the published whl) to use an older version for 3.7 

(something like):
```toml
"numpy" = [
  {version=">=1.20,<1.24.2", python="<3.8"},
  {version=">=1.24.2", python="^3.8"},
]
```

.github/workflows/PR.yml
bkeryan: Try this exclude list:
```
[tool.ni-python-styleguide]
extend_exclude = "docs,generated/nidaqmx,src/nidaqmx,src/codegen/metadata"
```

You will need to use `black` to reformat a few files in `tests/` before `ni-python-styleguide lint` will pass with this exclude list.

pyproject.toml
bkeryan: `numpy` dropped Python 3.7 support in version 1.22, so I think the 1st line should have `version=">=1.20,<1.22", python="<3.8"`.

bkeryan: We can add `sphinx_rtd_theme` to dev-dependencies with a stricter version constraint:
```
sphinx_rtd_theme = "1.0"
```

With this and the updated `numpy` constraints, `poetry lock` completes in a reasonable amount of time on my machine.

pyproject.toml
```diff
@@ -56,6 +56,15 @@ tox = "^3.24"
 click = "^8.1"
 Mako = "^1.2"
 deprecation = "^2.1"
+ni-python-styleguide = "^0.4"
+
+[tool.black]
+line-length = 100
+exclude = ["generated/",
```
mshafer-NI, line 63: specifying `exclude` can make black take a long time due to it now scanning a local .venv.
```suggestion
extend_exclude = ["generated/",
```

.github/workflows/PR.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          poetry run ni-python-styleguide lint ./src/codegen/properties
+          poetry run ni-python-styleguide lint ./src/codegen/utilities
+          poetry run ni-python-styleguide lint ./src/codegen/__init__.py
+          poetry run ni-python-styleguide lint ./src/codegen/__main__.py
+          poetry run ni-python-styleguide lint ./src/codegen/generator.py
+          poetry run ni-python-styleguide lint ./tests/
+          poetry run ni-python-styleguide lint ./nidaqmx_examples/
+      - name: Generate ni-daqmx files
+        run: poetry run python -m codegen --dest generated/nidaqmx
```
bkeryan, line 39: Let's add this to CONTRIBUTING.md

Sakthi-SM:
@zhindes , @bkeryan I am noticing some tests are failing in the pipeline but it is passing in my local machine. Looks like there is some issue in loading `ctypes`. Is there a specific OS version on which these tests are supposed to run? Please share your thoughts on resolving this issue.

CONTRIBUTING.md
```diff
@@ -22,11 +22,12 @@ To contribute to this project, it is recommended that you follow these steps:
 4. Run the unit tests on your system (see Testing section). At this point, if any tests fail, do not
 begin development. Try to investigate these failures. If you're unable to do so, report an issue
 through our [GitHub issues page](http://github.com/ni/nidaqmx-python/issues).
-5. Write new tests that demonstrate your bug or feature. Ensure that these new tests fail.
-6. Make your change.
-7. Run all the unit tests again (which include the tests you just added), and confirm that they all
+5. Once the necessary changes are done, update the auto-generated code using ``poetry run python -m codegen --dest generated/nidaqmx``. This will ensure that the latest files are present in ``generated`` folder.
```
zhindes, line 25: I think this should be after the "Make your change." step?

tests/test_channel_creation.py
```diff
@@ -33,7 +34,7 @@ class TestAnalogCreateChannels(object):
     """
 
     @pytest.mark.parametrize("seed", [generate_random_seed()])
-    def test_create_ai_voltage_chan(self, any_x_series_device, seed):
+    def test_create_ai_voltage_chan(self, any_x_series_device, seed):  # noqa: F811
```
zhindes, line 37: [F811: Redefinition of unused name from line n ](https://www.flake8rules.com/rules/F811.html) seems wrong. What is the error?

tests/test_channel_creation.py
```diff
@@ -20,6 +20,7 @@
     StrainGageBridgeType,
     BridgeConfiguration,
 )
+from nidaqmx.tests.fixtures import sim_ts_power_device, any_x_series_device  # noqa: F401
```
zhindes, line 23: Just remove whatever isn't used

tests/test_channels.py
```diff
@@ -4,6 +4,7 @@
 import nidaqmx
 import nidaqmx.system
 from nidaqmx.constants import TaskMode
+from nidaqmx.tests.fixtures import any_x_series_device, real_x_series_device  # noqa: F401
```
zhindes, line 7: Remove whatever isn't used.

tests/test_channel_creation.py
zhindes: This is everywhere, I'll only mark it once :D

tests/test_container_operations.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 
 import nidaqmx
+from nidaqmx.tests.fixtures import any_x_series_device  # noqa: F401
```
zhindes, line 7: Remove whatever isn't used

tests/test_events.py
```diff
@@ -6,6 +6,7 @@
 
 import nidaqmx
 from nidaqmx.constants import AcquisitionType
+from nidaqmx.tests.fixtures import any_x_series_device  # noqa: F401
```
zhindes, line 9: Remove whatever isn't used

tests/test_export_signals.py
```diff
@@ -5,6 +5,7 @@
 
 import nidaqmx
 from nidaqmx.constants import Signal
+from nidaqmx.tests.fixtures import any_x_series_device  # noqa: F401
```
zhindes, line 8: Remove whatever isn't used

zhindes: OK, I'll stop mentioning these

zhindes (changes requested):
overall changes look great. some confusion on the noqa's you've added:

* fix whatever F401 is complaining about, just remove the unused imports
* F811 doesn't really make sense... not sure what is going on there.

CONTRIBUTING.md
bkeryan: Or put it in a separate section about updating the codegen.

tests/test_channel_creation.py
mshafer-NI: It's complaining since `any_x_series_device` is imported, but then redefined as a parameter here.

Note on how Pytest works: the fixtures don't need to be imported into the test files.

tests/test_channel_creation.py
bkeryan: pytest looks these up at run time, but flake8 doesn't know that.

I think the solution is to rename `fixtures.py` to `conftest.py` and remove the imports:
https://docs.pytest.org/en/4.6.x/fixture.html#conftest-py-sharing-fixture-functions

pyproject.toml
```diff
@@ -57,14 +57,14 @@ click = "^8.1"
...
 line-length = 100
-exclude = ["generated/",
-    "docs/",
-    "src/nidaqmx/",
-    "src/codegen/metadata/",
-    "src/codegen/templates/"]
+extend_exclude = "generated,docs,src/nidaqmx,src/codegen/metadata,src/codegen/templates"
+
+[tool.ni-python-styleguide]
+extend_exclude = "docs,generated,src/nidaqmx,src/codegen/metadata,src/codegen/templates"
```
bkeryan, line 67: These are identical except for order. Please make them identical.

pyproject.toml
```diff
@@ -32,9 +32,9 @@ packages = [{ include = "nidaqmx", from = "src" }]
 [tool.poetry.dependencies]
 python = "^3.7"
 numpy = [
-    {version="^1.20", python = "<3.10"},
-    # precompiled wheels only available in the 1.22+
-    {version="^1.22", python = "^3.10"}
+  {version=">=1.20,<1.22", python="<3.8"},
+  # precompiled wheels only available in the 1.22+
+  {version=">=1.22", python="^3.8"},
```
bkeryan, line 37: This comment is about Python 3.10, not 3.8.

Also, it's not really correct because 1.21.3 has precompiled wheels for Python 3.10.

```suggestion
  # numpy 1.22 dropped support for Python 3.7.
  {version=">=1.20,<1.22", python="<3.8"},
  {version=">=1.22", python="^3.8"},
```

bkeryan (changes requested):

tests/test_channel_creation.py
mshafer-NI: (or import them into conftest.py)

bkeryan:
> @zhindes , @bkeryan I am noticing some tests are failing in the pipeline but it is passing in my local machine. Looks like there is some issue in loading `ctypes`. Is there a specific OS version on which these tests are supposed to run? Please share your thoughts on resolving this issue.

The errors explain the problem:
`E               nidaqmx._lib.DaqNotFoundError: Could not find an installation of NI-DAQmx. Please ensure that NI-DAQmx is installed on this machine or contact National Instruments for support.`

@zhindes pointed this out previously:

> [zhindes](https://github.com/zhindes) commented [2 weeks ago](https://github.com/ni/nidaqmx-python/pull/214#issuecomment-1450517803)
> @Sakthi-SM - Like I mentioned elsewhere all existing tests require DAQmx to be installed. Once gRPC support is integrated there will be new tests that don't require that. We've already tackled a similar problem. A subset of the tests require physical hardware. We made a [fixture](https://github.com/ni/nidaqmx-python/blob/master/tests/fixtures.py#L21) that detects if that hardware is available and skips otherwise. We should do a similar thing here. Create a fixture that detects if DAQmx is installed and skips otherwise. So for a while running the tests in a workflow will be sort of useless, but over time we can improve that.

Sakthi-SM:
> > @zhindes , @bkeryan I am noticing some tests are failing in the pipeline but it is passing in my local machine. Looks like there is some issue in loading `ctypes`. Is there a specific OS version on which these tests are supposed to run? Please share your thoughts on resolving this issue.
> 
> The errors explain the problem: `E nidaqmx._lib.DaqNotFoundError: Could not find an installation of NI-DAQmx. Please ensure that NI-DAQmx is installed on this machine or contact National Instruments for support.`
> 
> @zhindes pointed this out previously:
> 
> > [zhindes](https://github.com/zhindes) commented [2 weeks ago](https://github.com/ni/nidaqmx-python/pull/214#issuecomment-1450517803)
> > @Sakthi-SM - Like I mentioned elsewhere all existing tests require DAQmx to be installed. Once gRPC support is integrated there will be new tests that don't require that. We've already tackled a similar problem. A subset of the tests require physical hardware. We made a [fixture](https://github.com/ni/nidaqmx-python/blob/master/tests/fixtures.py#L21) that detects if that hardware is available and skips otherwise. We should do a similar thing here. Create a fixture that detects if DAQmx is installed and skips otherwise. So for a while running the tests in a workflow will be sort of useless, but over time we can improve that.
> 
> Here's an idea for how to run the subset of tests that don't require installing DAQmx:
> 
> * Create a fixture that calls `nidaqmx.system.System.local()`
> * In this fixture, catch `DaqNotFoundError` and call `pytest.skip(e.message)`
> * Update fixtures and tests that call `nidaqmx.system.System.local()` to use the new fixture
> 
> I think it's ok to disable pytest in the workflow for now and add this fixture in a separate PR.

I have disabled the pytest as of now. The above suggestion will be taken in a separate PR and then pytest will be enabled.

bkeryan (approved):
Approved with suggestions:
- Rerun `poetry lock`
- Rerun `black`

zhindes (approved):
Woo, thanks for the contribution! Once Brad's changes are addressed, I'll merge :)

poetry.lock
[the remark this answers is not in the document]
samchris007: Updated.

tests/test_read_exceptions.py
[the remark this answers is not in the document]
samchris007: Updated.

PR #215, opened 2023-03-01
DAQmx code generator for error_codes.py and constants.py using scrapigen metadata

vigkre:
- This PR creates DAQmx code generator to generate `error_codes.py` and `constants.py`.
- Added attributes.py and enums.py to the metadata. Currently the attribute.py and enums.py are created under PR's [119](https://github.com/ni/grpc-device-scrapigen/pull/119) and [127](https://github.com/ni/grpc-device-scrapigen/pull/127) in [scrapigen](https://github.com/ni/grpc-device-scrapigen) 
- This PR can be merged once the scrapigen metadata is merged such that both repos are synced.
- Added the `mako` templates necessary for generating `error_codes.py` and `constants.py`.
- The newly generated `error_codes.py` and `constants.py` replaces the files generated by native daqmx code generator (https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nidaqmx_py/source/nidaqmx_codegen) which is currently residing in (https://github.com/ni/nidaqmx-python/tree/master/src/nidaqmx)


generated/nidaqmx/constants.py
```diff
@@ -57,8 +57,8 @@ class AccelUnits(Enum):
 
 
 class AcquisitionType(Enum):
-    FINITE = 10178  #: Acquire or generate a finite number of samples.
-    CONTINUOUS = 10123  #: Acquire or generate samples until you stop the task.
+    FINITE_SAMPS = 10178  #: Acquire or generate a finite number of samples.
+    CONT_SAMPS = 10123  #: Acquire or generate samples until you stop the task.
```
zhindes, line 61: This is going to nearly everything. I would prefer if these names didn't change.

generated/nidaqmx/constants.py
```diff
@@ -68,8 +68,8 @@ class Action(Enum):
 
 
 class ActiveLevel(Enum):
-    ABOVE = 10093  #: Pause the measurement or generation while the signal is above the threshold.
-    BELOW = 10107  #: Pause the measurement or generation while the signal is below the threshold.
+    ABOVE_LVL = 10093  #: Pause the measurement or generation while the signal is above the threshold.
+    BELOW_LVL = 10107  #: Pause the measurement or generation while the signal is below the threshold.
```
zhindes, line 72: This change I have less of a problem with. However, I think we should avoid shorthand versions of words in Python. So LVL should be LEVEL, I think.

generated/nidaqmx/constants.py
```diff
@@ -121,8 +121,8 @@ class BridgePhysicalUnits(Enum):
     POUNDS = 15876  #: Pounds.
     KILOGRAM_FORCE = 15877  #: kilograms-force.
     PASCALS = 10081  #: Pascals.
+    POUNDS_PER_SQUARE_INCH = 15879  #: Pounds per square inch.
```
zhindes, line 124: I like this change... shorthand SQ became SQUARE. I think that is a good and more consistent change.

generated/nidaqmx/constants.py
```diff
@@ -169,12 +169,12 @@ class CalibrationTerminalConfig(Enum):
...
-    DIGITAL_INPUT = 10151  #: Digital input channel.
-    DIGITAL_OUTPUT = 10153  #: Digital output channel.
-    COUNTER_INPUT = 10131  #: Counter input channel.
-    COUNTER_OUTPUT = 10132  #: Counter output channel.
+    AI = 10100  #: Analog input channel.
+    AO = 10102  #: Analog output channel.
+    DI = 10151  #: Digital input channel.
+    DO = 10153  #: Digital output channel.
+    CI = 10131  #: Counter input channel.
+    CO = 10132  #: Counter output channel.
```
zhindes, line 177: I dislike these changes. I prefer fully spelled out.

generated/nidaqmx/constants.py
```diff
@@ -197,16 +197,18 @@ class CountDirection(Enum):
...
-    LARGE_RANGE_2_COUNTERS = 10205  #: Use one counter to divide the frequency of the input signal to create a lower-frequency signal that the second counter can more easily measure.
-    DYNAMIC_AVERAGING = 16065  #: Uses one counter with configuration options to control the amount of averaging or filtering applied to the counter measurements. Set filtering options to balance measurement accuracy and noise versus latency.
+    LOW_FREQ_1_CTR = 10105  #: Use one counter that uses a constant timebase to measure the input signal.
+    HIGH_FREQ_2_CTR = 10157  #: Use two counters, one of which counts pulses of the signal to measure during the specified measurement time.
+    LARGE_RNG_2_CTR = 10205  #: Use one counter to divide the frequency of the input signal to create a lower-frequency signal that the second counter can more easily measure.
+    DYN_AVG = 16065  #: Uses one counter with configuration options to control the amount of averaging or filtering applied to the counter measurements. Set filtering options to balance measurement accuracy and noise versus latency.
```
zhindes, line 203: Dislike the shorthand change

generated/nidaqmx/constants.py
```diff
@@ -234,9 +238,9 @@ class DataTransferActiveTransferMode(Enum):
 
 
 class DeassertCondition(Enum):
-    ON_BOARD_MEMORY_MORE_THAN_HALF_FULL = 10237  #: Deassert the signal when more than half of the onboard memory of the device fills.
-    ON_BOARD_MEMORY_FULL = 10236  #: Deassert the signal when the onboard memory fills.
-    ONBOARD_MEMORY_CUSTOM_THRESHOLD = 12577  #: Deassert the signal when the amount of space available in the onboard memory is below the value specified with **rdy_for_xfer_event_deassert_cond_custom_threshold**.
+    ONBRD_MEM_MORE_THAN_HALF_FULL = 10237  #: Deassert the signal when more than half of the onboard memory of the device fills.
+    ONBRD_MEM_FULL = 10236  #: Deassert the signal when the onboard memory fills.
+    ONBRD_MEM_CUSTOM_THRESHOLD = 12577  #: Deassert the signal when the amount of space available in the onboard memory is below the value specified with **rdy_for_xfer_event_deassert_cond_custom_threshold**.
```
zhindes, line 243: This one is a mixed bag. The old names were inconsistent (ON_BOARD vs ONBOARD), so fixing the inconsistency is great. However, I don't like ONBRD, and would prefer ONBOARD

generated/nidaqmx/constants.py
```diff
@@ -269,9 +273,9 @@ class Edge(Enum):
 
 
 class EncoderType(Enum):
-    X_1 = 10090  #: If signal A leads signal B, count the rising edges of signal A. If signal B leads signal A, count the falling edges of signal A.
-    X_2 = 10091  #: Count the rising and falling edges of signal A.
-    X_4 = 10092  #: Count the rising and falling edges of signal A and signal B.
+    X1 = 10090  #: If signal A leads signal B, count the rising edges of signal A. If signal B leads signal A, count the falling edges of signal A.
+    X2 = 10091  #: Count the rising and falling edges of signal A.
+    X4 = 10092  #: Count the rising and falling edges of signal A and signal B.
```
zhindes, line 278: This is fine

generated/nidaqmx/constants.py
```diff
@@ -293,12 +297,12 @@ class EveryNSamplesEventType(Enum):
 
 
 class ExcitationDCorAC(Enum):
-    USE_DC = 10050  #: DC excitation.
-    USE_AC = 10045  #: AC excitation.
+    DC = 10050  #: DC excitation.
+    AC = 10045  #: AC excitation.
```
zhindes, line 301: This is fine

generated/nidaqmx/constants.py
```diff
@@ -309,8 +313,8 @@ class ExcitationSource(Enum):
 
 
 class ExcitationVoltageOrCurrent(Enum):
-    USE_VOLTAGE = 10322  #: Voltage excitation.
-    USE_CURRENT = 10134  #: Current excitation.
+    VOLTAGE = 10322  #: Voltage excitation.
+    CURRENT = 10134  #: Current excitation.
```
zhindes, line 317: This is fine. USE_X was always weird

generated/nidaqmx/constants.py
```diff
@@ -394,15 +401,15 @@ class InputCalSource(Enum):
...
 
 class InputDataTransferCondition(Enum):
-    ON_BOARD_MEMORY_MORE_THAN_HALF_FULL = 10237  #: Transfer data from the device when more than half of the onboard memory of the device fills.
-    ON_BOARD_MEMORY_NOT_EMPTY = 10241  #: Transfer data from the device when there is data in the onboard memory.
-    ONBOARD_MEMORY_CUSTOM_THRESHOLD = 12577  #: Transfer data from the device when the number of samples specified with **ai_data_xfer_custom_threshold** are in the device FIFO.
-    WHEN_ACQUISITION_COMPLETE = 12546  #: Transfer data when the acquisition is complete.
+    ONBRD_MEM_MORE_THAN_HALF_FULL = 10237  #: Transfer data from the device when more than half of the onboard memory of the device fills.
+    ONBRD_MEM_NOT_EMPTY = 10241  #: Transfer data from the device when there is data in the onboard memory.
+    ONBRD_MEM_CUSTOM_THRESHOLD = 12577  #: Transfer data from the device when the number of samples specified with **ai_data_xfer_custom_threshold** are in the device FIFO.
+    WHEN_ACQ_COMPLETE = 12546  #: Transfer data when the acquisition is complete.
```
zhindes, line 407: ONBOARD, MEMORY, and ACQUISITION please

generated/nidaqmx/constants.py
```diff
@@ -394,15 +401,15 @@ class InputCalSource(Enum):
...
 
 
 class InvertPolarity(Enum):
-    NO = 0
-    YES = 1
+    DO_NOT_INVERT_POLARITY = 0  #: Do not invert polarity
+    INVERT_POLARITY = 1  #: Invert polarity
```
zhindes, line 412: This is fine, but YES and NO are a bit nicer, if we can easily fix it.

generated/nidaqmx/constants.py
```diff
@@ -469,9 +478,9 @@ class ModulationType(Enum):
 
 
 class OutputDataTransferCondition(Enum):
-    ON_BOARD_MEMORY_EMPTY = 10235  #: Transfer data to the device only when there is no data in the onboard memory of the device.
-    ON_BOARD_MEMORY_HALF_FULL_OR_LESS = 10239  #: Transfer data to the device any time the onboard memory is less than half full.
-    ON_BOARD_MEMORY_LESS_THAN_FULL = 10242  #: Transfer data to the device any time the onboard memory of the device is not full.
+    ONBRD_MEM_EMPTY = 10235  #: Transfer data to the device only when there is no data in the onboard memory of the device.
+    ONBRD_MEM_HALF_FULL_OR_LESS = 10239  #: Transfer data to the device any time the onboard memory is less than half full.
+    ONBRD_MEM_NOT_FULL = 10242  #: Transfer data to the device any time the onboard memory of the device is not full.
```
zhindes, line 483: ONBOARD and MEMORY please

generated/nidaqmx/constants.py
```diff
@@ -525,12 +534,12 @@ class PowerUpChannelType(Enum):
...
     HIGH = 10192  #: Logic high.
     LOW = 10214  #: Logic low.
-    TRISTATE = 10310  #: High-impedance state. You can select this state only on devices with bidirectional lines.  You cannot select this state for dedicated digital output lines. On some devices, you can select this value only for entire ports.
+    TRISTATE = 10310  #: High-impedance state. You can select this state only on devices with bidirectional lines. You cannot select this state for dedicated digital output lines. On some devices, you can select this value only for entire ports.
 
 
 class PressureUnits(Enum):
     PASCALS = 10081  #: Pascals.
-    POUNDS_PER_SQ_INCH = 15879  #: Pounds per square inch.
+    POUNDS_PER_SQUARE_INCH = 15879  #: Pounds per square inch.
```
zhindes, line 542: Good!

generated/nidaqmx/constants.py
```diff
@@ -546,7 +555,7 @@ class ProductCategory(Enum):
     AO_SERIES = 14647  #: AO Series.
     DIGITAL_IO = 14648  #: Digital I/O.
     TIO_SERIES = 14661  #: TIO Series.
-    DSA = 14649  #: Dynamic Signal Acquisition.
+    DYNAMIC_SIGNAL_ACQUISITION = 14649  #: Dynamic Signal Acquisition.
```
zhindes, line 558: Good!

generated/nidaqmx/constants.py
```diff
@@ -586,10 +595,10 @@ class RawDataCompressionType(Enum):
...
-    FIRST_PRETRIGGER_SAMPLE = 10427  #: Start reading samples relative to the first pretrigger sample. You specify the number of pretrigger samples to acquire when you configure a reference trigger.
-    MOST_RECENT_SAMPLE = 10428  #: Start reading samples relative to the next sample acquired. For example, use this value and set **offset** to -1 to read the last sample acquired.
+    CURR_READ_POS = 10425  #: Start reading samples relative to the last sample returned by the previous read. For the first read operation, this position is the first sample acquired or the first pretrigger sample if you configured a reference trigger for the task.
+    REF_TRIG = 10426  #: Start reading samples relative to the first sample after the reference trigger occurred.
+    FIRST_PRETRIG_SAMP = 10427  #: Start reading samples relative to the first pretrigger sample. You specify the number of pretrigger samples to acquire when you configure a reference trigger.
+    MOST_RECENT_SAMP = 10428  #: Start reading samples relative to the next sample acquired. For example, use this value and set **offset** to -1 to read the last sample acquired.
```
zhindes, line 601: Old names were preferable

generated/nidaqmx/constants.py
```diff
@@ -609,33 +618,36 @@ class ResistanceConfiguration(Enum):
...
 class SCXI1124Range(Enum):
     ZERO_TO_ONE_V = 14629
     ZERO_TO_FIVE_V = 14630
     ZERO_TO_TEN_V = 14631
-    NEG_1_TO_1_V = 14632
-    NEG_5_TO_5_V = 14633
+    NEG_1TO_1_V = 14632
+    NEG_5TO_5_V = 14633
     NEG_10_TO_10_V = 14634
     ZERO_TO_TWENTY_M_A = 14635
```
zhindes, line 645: Ugh both old and new are bad. If an attribute doesn't actually use this, lets delete it, But I suspect it does despite SCXI being EOL since `nidaqmx` works with old versions. Can you make these all consistent? I hate spelling out numbers, but we may not be able to lead with a number, so if that is the case them make them all spelled out. Also M_A is bad, make it MA.

generated/nidaqmx/constants.py
```diff
@@ -609,33 +618,36 @@ class ResistanceConfiguration(Enum):
...
 
 
 class SampClkOverrunBehavior(Enum):
-    REPEAT_LAST_SAMPLE = 16062  #: Repeat the last sample.
-    RETURN_SENTINEL_VALUE = 16063  #: Return the sentinel value.
+    REPEATED_DATA = 16062  #: Repeat the last sample.
+    SENTINEL_VALUE = 16063  #: Return the sentinel value.
```
zhindes, line 650: I'm fine with this.

generated/nidaqmx/constants.py
```diff
@@ -644,13 +656,13 @@ class SampleInputDataWhen(Enum):
 
 
 class SampleTimingType(Enum):
-    SAMPLE_CLOCK = 10388  #: Acquire or generate samples on the specified edge of the sample clock.
+    SAMP_CLK = 10388  #: Acquire or generate samples on the specified edge of the sample clock.
```
zhindes, line 659: old is preferable

generated/nidaqmx/constants.py
```diff
@@ -644,13 +656,13 @@ class SampleInputDataWhen(Enum):
...
-    PIPELINED_SAMPLE_CLOCK = 14668  #: Device acquires or generates samples on each sample clock edge, but does not respond to certain triggers until a few sample clock edges later. Pipelining allows higher data transfer rates at the cost of increased trigger response latency.  Refer to the device documentation for information about which triggers pipelining affects. This timing type allows handshaking with some devices using the Pause trigger, the Ready for Transfer event, or the Data Active event. Refer to the device documentation for more information.
+    PIPELINED_SAMP_CLK = 14668  #: Device acquires or generates samples on each sample clock edge, but does not respond to certain triggers until a few sample clock edges later. Pipelining allows higher data transfer rates at the cost of increased trigger response latency. Refer to the device documentation for information about which triggers pipelining affects. This timing type allows handshaking with some devices using the Pause trigger, the Ready for Transfer event, or the Data Active event. Refer to the device documentation for more information.
```
zhindes, line 665: old is preferable

generated/nidaqmx/constants.py
```diff
@@ -691,11 +703,11 @@ class ShuntCalSelect(Enum):
...
-    R_1 = 12465
-    R_2 = 12466
-    R_3 = 12467
-    R_4 = 14813
-    NONE = 10230
+    R1 = 12465  #: xx
+    R2 = 12466  #: xx
+    R3 = 12467  #: xx
+    R4 = 14813  #: xx
+    NONE = 10230  #: xx
```
zhindes, line 710: I'm fine with either old or new, but lets delete placeholder comments like this.

generated/nidaqmx/constants.py
```diff
@@ -715,8 +727,8 @@ class Signal(Enum):
     AI_HOLD_CMPLT_EVENT = 12493
     COUNTER_OUTPUT_EVENT = 12494  #: Timed Loop executes each time the Counter Output Event occurs.
     CHANGE_DETECTION_EVENT = 12511  #: Timed Loop executes each time the Change Detection Event occurs.
-    WATCHDOG_TIMER_EXPIRED_EVENT = 12512
-    SAMPLE_COMPLETE = 12530  #: Timed Loop executes each time the Sample Complete Event occurs.
+    WDT_EXPIRED_EVENT = 12512
```
zhindes, line 730: Old please

generated/nidaqmx/constants.py
```diff
@@ -715,8 +727,8 @@ class Signal(Enum):
...
+    SAMPLE_COMPLETE_EVENT = 12530  #: Timed Loop executes each time the Sample Complete Event occurs.
```
zhindes, line 731: This is fine

generated/nidaqmx/constants.py
```diff
@@ -731,6 +743,12 @@ class Slope(Enum):
 
 class SoftwareTrigger(Enum):
     ADVANCE_TRIGGER = 12488  #: Place holder enum to make editting internal enum easier.
+    REFERENCE_TRIGGER = 12490
+    ARM_REFERENCE_TRIGGER = 12496
+    STOP_TRIGGER = 14834
+    IGGER_PAUSE_TRIGGER = 14654
+    IGGER_START_TRIGGER = 14656
+    IGGER_SCRIPT_TRIGGER = 14651
```
zhindes, line 751: No idea what's going on here. Delete?

generated/nidaqmx/constants.py
```diff
@@ -765,9 +783,9 @@ class StrainGageRosetteMeasurementType(Enum):
 
 
 class StrainGageRosetteType(Enum):
-    RECTANGULAR = 15968  #: A rectangular rosette consists of three strain gages, each separated by a 45 degree angle.
-    DELTA = 15969  #: A delta rosette consists of three strain gages, each separated by a 60 degree angle.
-    TEE = 15970  #: A tee rosette consists of two gages oriented at 90 degrees with respect to each other.
+    RECTANGULAR_ROSETTE = 15968  #: A rectangular rosette consists of three strain gages, each separated by a 45 degree angle.
+    DELTA_ROSETTE = 15969  #: A delta rosette consists of three strain gages, each separated by a 60 degree angle.
+    TEE_ROSETTE = 15970  #: A tee rosette consists of two gages oriented at 90 degrees with respect to each other.
```
zhindes, line 788: Old seems preferable since its less repetition. Rosette is in the enum name.

generated/nidaqmx/constants.py
```diff
@@ -783,7 +801,7 @@ class SwitchChannelUsage(Enum):
 
 class SyncPulseType(Enum):
     ONBOARD = 16128  #: Use the synchronization pulse type specified by the device.
-    DIGITAL_EDGE = 10150  #: Digital Edge synchronization.
+    DIG_EDGE = 10150  #: Digital Edge synchronization.
```
zhindes, line 804: Prefer old

generated/nidaqmx/constants.py
```diff
@@ -822,10 +841,10 @@ class TaskState(Enum):
 
 
 class TemperatureUnits(Enum):
-    DEG_C = 10143  #: Degrees Celsius.
-    DEG_F = 10144  #: Degrees Fahrenheit.
-    K = 10325  #: Kelvins.
-    DEG_R = 10145  #: Degrees Rankine.
+    DEG_C = 10143  #: Deg C
+    DEG_F = 10144  #: Deg F
+    KELVINS = 10325  #: Kelvins
```
zhindes, line 846: Both and and new are wildly inconsistent. Maybe this should be DEG_K?

generated/nidaqmx/constants.py
```diff
@@ -850,27 +869,28 @@ class ThermocoupleType(Enum):
...
 
 class TimestampEvent(Enum):
-    START_TRIGGER = 12491  #: Start Trigger timestamp.
-    REFERENCE_TRIGGER = 12490  #: Reference Trigger timestamp.
-    ARM_START_TRIGGER = 14641  #: Arm Start Trigger timestamp.
-    FIRST_SAMPLE = 16130  #: First Sample timestamp.
+    START_TRIGGER = 12491  #: Start Trigger
+    REFERENCE_TRIGGER = 12490  #: Reference Trigger
+    ARM_START_TRIGGER = 14641  #: Arm Start Trigger
+    FIRST_SAMPLE_TIMESTAMP = 16130  #: First Sample Timestamp
```
zhindes, line 886: Prefer old since its less repetitive

generated/nidaqmx/constants.py
```diff
@@ -881,10 +901,14 @@ class TriggerType(Enum):
     DIGITAL_EDGE = 10150  #: Trigger on a rising or falling edge of a digital pulse.
     DIGITAL_LEVEL = 10152  #: Pause the measurement or generation while a digital signal is at either a high or low state.
     DIGITAL_PATTERN = 10398  #: Pause the measurement or generation while digital physical channels either match or do not match a digital pattern.
+    EXTERNAL_DIGITAL_CLOCK = 10400
+    RATE = 10402
```
zhindes, line 905: This looks like old internal constant... can we remove?

generated/nidaqmx/constants.py
```diff
@@ -881,10 +901,14 @@ class TriggerType(Enum):
...
     ANALOG_EDGE = 10099  #: Trigger when an analog signal crosses a threshold.
     ANALOG_MULTI_EDGE = 16108  #: Trigger when any of the configured analog signals cross their respective thresholds.
+    VIDEO = 10476
```
zhindes, line 908: This looks like old internal constant... can we remove?

generated/nidaqmx/constants.py
```diff
@@ -907,10 +931,10 @@ class UnitsPreScaled(Enum):
     DEG_F = 10144  #: Degrees Fahrenheit.
     DEG_C = 10143  #: Degrees Celsius.
     DEG_R = 10145  #: Degrees Rankine.
-    K = 10325  #: Kelvins.
+    KELVINS = 10325  #: Kelvins.
     STRAIN = 10299  #: Strain.
     OHMS = 10384  #: Ohms.
-    HERTZ = 10373  #: Hertz.
+    HZ = 10373  #: Hertz.
```
zhindes, line 937: prefer old

generated/nidaqmx/constants.py
```diff
@@ -907,10 +931,10 @@ class UnitsPreScaled(Enum):
     DEG_F = 10144  #: Degrees Fahrenheit.
     DEG_C = 10143  #: Degrees Celsius.
     DEG_R = 10145  #: Degrees Rankine.
-    K = 10325  #: Kelvins.
+    KELVINS = 10325  #: Kelvins.
```
zhindes, line 934: whatever you do on the previous instance, match it here

generated/nidaqmx/constants.py
```diff
@@ -925,12 +949,12 @@ class UnitsPreScaled(Enum):
     INCHES_PER_SECOND_SQUARED = 12471  #: Inches per second per second.
     METERS_PER_SECOND = 15959  #: Meters per second.
     INCHES_PER_SECOND = 15960  #: Inches per second.
-    PA = 10081  #: Pascals.
+    PASCALS = 10081  #: Pascals.
```
zhindes, line 952: I like this

generated/nidaqmx/constants.py
```diff
@@ -925,12 +949,12 @@ class UnitsPreScaled(Enum):
...
     NEWTONS = 15875  #: Newtons.
     POUNDS = 15876  #: Pounds.
     KILOGRAM_FORCE = 15877  #: Kilograms-force.
+    POUNDS_PER_SQUARE_INCH = 15879  #: Pounds per square inch.
```
zhindes, line 956: I like this

generated/nidaqmx/constants.py
```diff
@@ -962,7 +986,7 @@ class UsageTypeAI(Enum):
     ROSETTE_STRAIN_GAGE = 15980  #: Strain measurement using a rosette strain gage.
     ACCELERATION_ACCELEROMETER_CURRENT_INPUT = 10356  #: Acceleration measurement using an accelerometer.
     ACCELERATION_CHARGE = 16104  #: Acceleration measurement using a charge-based sensor.
-    ACCELERATION_4_WIRE_DC_VOLTAGE = 16106  #: Acceleration measurement using a 4 wire DC voltage based sensor.
+    ACCELERATION_FOUR_WIRE_DC_VOLTAGE = 16106  #: Acceleration measurement using a 4 wire DC voltage based sensor.
```
zhindes, line 989: I think I like this?

generated/nidaqmx/constants.py
```diff
@@ -993,12 +1017,15 @@ class UsageTypeCI(Enum):
     POSITION_ANGULAR_ENCODER = 10360  #: Angular position measurement using an angular encoder.
     POSITION_LINEAR_ENCODER = 10361  #: Linear position measurement using a linear encoder.
     TIME_GPS = 10362  #: Timestamp measurement, synchronizing the counter to a GPS receiver.
+    TIMESTAMP = 15986  #: Timestamp measurement.
+    CHANNEL_TYPE_INDICATOR = 16030
```
zhindes, line 1021: This looks like an internal value. Can you remove?

generated/nidaqmx/constants.py
```diff
@@ -1013,9 +1040,15 @@ class VelocityUnits(Enum):
...
     FROM_CUSTOM_SCALE = 10065  #: Units a custom scale specifies. If you select this value, you must specify a custom scale name.
+    KILOMETERS_PER_HOUR = 16007  #: Kilometers per hour.
+    FEET_PER_SECOND = 16008  #: Feet per second.
+    MILES_PER_HOUR = 16009  #: Milers per hour.
+    KNOTS = 16010  #: Knots.
 
 
 class VoltageUnits(Enum):
+    NANO_VOLTS = 10345
+    MILLIVOLTS = 10347
```
zhindes, line 1051: These are inconsistent with underscores. Make them match or just remove these constants since they look unused

generated/nidaqmx/constants.py
```diff
@@ -1057,24 +1090,24 @@ class WaveformAttributes(Enum):
...
-    LEAVING_WINDOW = 10208  #: Trigger when the signal leaves the window.
+    ENTERING_WIN = 10163  #: Trigger when the signal enters the window.
+    LEAVING_WIN = 10208  #: Trigger when the signal leaves the window.
 
 
 class WindowTriggerCondition2(Enum):
-    INSIDE_WINDOW = 10199  #: Pause the measurement or generation while the trigger is inside the window.
-    OUTSIDE_WINDOW = 10251  #: Pause the measurement or generation while the signal is outside the window.
+    INSIDE_WIN = 10199  #: Pause the measurement or generation while the trigger is inside the window.
+    OUTSIDE_WIN = 10251  #: Pause the measurement or generation while the signal is outside the window.
```
zhindes, line 1099: WINDOW

generated/nidaqmx/constants.py
```diff
@@ -1057,24 +1090,24 @@ class WaveformAttributes(Enum):
...
-    DO_NOT_WRITE = 12540  #: blah
+    WRITE_TO_EEPROM = 12538  #: Write basic TEDS data to the EEPROM, even if the sensor includes a PROM. You cannot write basic TEDS data if the PROM contains data.
+    WRITE_TO_PROM = 12539  #: Write basic TEDS data to the PROM. Any subsequent attempts to write basic TEDS data result in an error.
+    DO_NOT_WRITE = 12540  #: Ignore basic TEDS data.
 
 
 class WriteRelativeTo(Enum):
     FIRST_SAMPLE = 10424  #: Write samples relative to the first sample.
-    CURRENT_WRITE_POSITION = 10430  #: Write samples relative to the current position in the buffer.
+    CURR_WRITE_POS = 10430  #: Write samples relative to the current position in the buffer.
```
zhindes, line 1110: prefer old

zhindes (changes requested):
I went through constants.py. My general guidelines
* I'm fine breaking compatibility here, but lets be consistent moving forward.
* I prefer spelling out names, eg ONBOARD not ONBRD
* I prefer not repeating the name of the enum in the value. So ActiveLevel.ABOVE is better than ActiveLevel.ABOVE_LEVEL
* I don't think we can lead with numbers, eg 1_VALUE, so we try to spell out numbers. I think for consistency we should do that everywhere.

Before we hack on the generator / metadata, lets make sure @bkeryan and I agree with our guidelines.

generated/nidaqmx/constants.py
zhindes: On further reflection, SAMPS doesn't repeat the enum name, so that's fine actually. But prefer SAMPLES

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,283 @@
...
+    for _, enum in enums.items():
+        for value in enum['values']:
+            value_name = value['name']
+    
+            # For Impedance1
+            if value_name in IMPEDANCE_ENUM_VALUE_NAME_SUB.keys():
+                value_name = IMPEDANCE_ENUM_VALUE_NAME_SUB[value_name]
+                
+            for old, new in NAME_SUBSTITUTIONS.items():
+                value_name = value_name.replace(old, new)
```
bkeryan, line 150: When we do enum substitutions here, we don't do the same substitution in docstrings that reference the enum. (Maybe we could, but we would have to be careful about matching substrings.)

When we do enum substitutions in scrapigen, they cover both the enum and any docstrings that reference the enum.

generated/nidaqmx/constants.py
```diff
@@ -57,8 +57,8 @@ class AccelUnits(Enum):
```
bkeryan, line 57: I think the inconsistencies in this file highlights why we want to generate nidaqmx-python and nidaqmx.proto from the same metadata in the first place: otherwise, it will be difficult to match up things that have different names in Python vs. gRPC.

I also think that a lot of these differences are caused by using different parts of the metadata: enum tFoo vs. lvEnum Foo1, etc. Some are also caused by the Python-specific fixups and substitutions (or lack thereof).

bkeryan (changes requested):

bkeryan:
> I went through constants.py. My general guidelines
> 
> * I'm fine breaking compatibility here, but lets be consistent moving forward.
> * I prefer spelling out names, eg ONBOARD not ONBRD
> * I prefer not repeating the name of the enum in the value. So ActiveLevel.ABOVE is better than ActiveLevel.ABOVE_LEVEL
> * I don't think we can lead with numbers, eg 1_VALUE, so we try to spell out numbers. I think for consistency we should do that everywhere.
> 
> Before we hack on the generator / metadata, lets make sure @bkeryan and I agree with our guidelines.
> 
> EDIT: Oh, and these breaking changes are going to break many existing tests. We can't change them in this PR since we're building `nidaqmx` out of source still, but we will need to update them at some point. Let's make sure we have that tracked in the Features/User Stories or something. I'll at least notice when we switch the `pyproject.toml` to be from generated. I'll want to ensure all tests pass then.

I would be happy if we generated the same names as before, warts and all. I don't mind improving names that are bad, but there are so many differences and it's not as clear cut as with mangled attribute ids like `ai_acceld_b_ref`.

generated/nidaqmx/constants.py
```diff
@@ -326,26 +330,29 @@ class FillMode(Enum):
...
-    BRICKWALL = 16155  #: Brickwall filter response.
-    COMB = 16152  #: Comb filter response.
-    BESSEL = 16153  #: Bessel filter response.
     CONSTANT_GROUP_DELAY = 16075  #: Constant group delay filter response.
     BUTTERWORTH = 16076  #: Butterworth filter response.
     ELLIPTICAL = 16077  #: Elliptical filter response.
     HARDWARE_DEFINED = 10191  #: Use the hardware-defined filter response.
+    COMB = 16152  #: Comb filter response.
+    BESSEL = 16153  #: Bessel filter response.
+    BRICKWALL = 16155  #: Brickwall filter response.
```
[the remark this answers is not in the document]
zhindes: Yeah, I figured this was unavoidable. Best case is we order these by enum value or enum name, but that would be a massive shuffle. So whatever is happening here felt like the best bet for least disruptive code review. We can update them to sort consistently later.

zhindes:
> I would be happy if we generated the same names as before, warts and all. I don't mind improving names that are bad, but there are so many differences and it's not as clear cut as with mangled attribute ids like ai_acceld_b_ref.

> Remember, the goal is to generate nidaqmx-python and nidaqmx.proto with the same metadata, so we don't need a ton of substitutions and special cases to map Python names to gRPC names. The goal isn't to fix all of the weird naming inconsistencies in nidaqmx-python (though that would be nice).

I hadn't considered your second point about consistency between Python and gRPC. I think that is a really important observation. So I think my guidelines are now this in order of priority.

1. Prefer not breaking compatibility, but I expect it will happen. It isn't a showstopper.
2. Prefer consistency with gRPC.
3. If something has to change or be incompatible with gRPC, then:
   * I prefer spelling out names, eg ONBOARD not ONBRD
   * I prefer not repeating the name of the enum in the value. So ActiveLevel.ABOVE is better than ActiveLevel.ABOVE_LEVEL
   * I don't think we can lead with numbers, eg 1_VALUE, so we try to spell out numbers. I think for consistency we should do that everywhere.

Not exactly black and white, but hopefully something like this can give some direction?

bkeryan:
> > I would be happy if we generated the same names as before, warts and all. I don't mind improving names that are bad, but there are so many differences and it's not as clear cut as with mangled attribute ids like ai_acceld_b_ref.
> 
> > Remember, the goal is to generate nidaqmx-python and nidaqmx.proto with the same metadata, so we don't need a ton of substitutions and special cases to map Python names to gRPC names. The goal isn't to fix all of the weird naming inconsistencies in nidaqmx-python (though that would be nice).
> 
> I hadn't considered your second point about consistency between Python and gRPC. I think that is a really important observation. So I think my guidelines are now this in order of priority.
> 
> 1. Prefer not breaking compatibility, but I expect it will happen. It isn't a showstopper.
> 2. Prefer consistency with gRPC.
> 3. If something has to change or be incompatible with gRPC, then:
>    
>    * I prefer spelling out names, eg ONBOARD not ONBRD
>    * I prefer not repeating the name of the enum in the value. So ActiveLevel.ABOVE is better than ActiveLevel.ABOVE_LEVEL
>    * I don't think we can lead with numbers, eg 1_VALUE, so we try to spell out numbers. I think for consistency we should do that everywhere.
> 
> Not exactly black and white, but hopefully something like this can give some direction?

Ok, that seems reasonable to me.

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,283 @@
+import logging
```
bkeryan, line 1: This branch doesn't work on my machine:

```
PS C:\dev\nidaqmx-python> poetry run python -m codegen --dest .\generated\nidaqmx
[    79] ERROR main: Failed to generate
  File "C:\Dev\nidaqmx-python\src\codegen\__main__.py", line 41, in main
    generator.generate(dest)
  File "C:\Dev\nidaqmx-python\src\codegen\generator.py", line 51, in generate
    _generate_file(codegen_metadata,"constants.mako", dest / "constants.py")
  File "C:\Dev\nidaqmx-python\src\codegen\generator.py", line 34, in _generate_file
    f.write(template.render(data=metadata))
  File "C:\Users\bkeryan\AppData\Local\pypoetry\Cache\virtualenvs\nidaqmx-H39Fc4kQ-py3.9\lib\site-packages\mako\template.py", line 439, in render
    return runtime._render(self, self.callable_, args, data)
  File "C:\Users\bkeryan\AppData\Local\pypoetry\Cache\virtualenvs\nidaqmx-H39Fc4kQ-py3.9\lib\site-packages\mako\runtime.py", line 874, in _render
    _render_context(
  File "C:\Users\bkeryan\AppData\Local\pypoetry\Cache\virtualenvs\nidaqmx-H39Fc4kQ-py3.9\lib\site-packages\mako\runtime.py", line 916, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File "C:\Users\bkeryan\AppData\Local\pypoetry\Cache\virtualenvs\nidaqmx-H39Fc4kQ-py3.9\lib\site-packages\mako\runtime.py", line 943, in _exec_template
    callable_(context, *args, **kwargs)
  File "C__Dev_nidaqmx_python_src_codegen_templates_constants_mako", line 22, in render_body
ModuleNotFoundError: No module named 'utilities'
```

The `utilities` submodule needs an `__init__.py`.

src/codegen/templates/constants.mako
```diff
@@ -0,0 +1,26 @@
+<%
+import utilities.enum_helpers as enum_helpers
```
bkeryan, line 2: `utilities` is a submodule of `codegen`.
```suggestion
import codegen.utilities.enum_helpers as enum_helpers
```

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,283 @@
...
+    # names.
+    # Replace occurrences of &DAQmx(FunctionName) with correct name.
+    docstring = re.sub(
+        "&DAQmx([^;]+);",
+        lambda m: "DAQmx " + m.group(1),
+        docstring,
+    )
+
+    html_parser = HTMLParser()
+    docstring = html_parser.unescape(docstring)
```
bkeryan, line 194: `HTMLParser.unescape` was removed in Python 3.9.

```suggestion
    docstring = html.unescape(docstring)
```

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,283 @@
+import logging
+import re
+
+from html.parser import HTMLParser
```
bkeryan, line 4:
```suggestion
import html
```

src/codegen/templates/constants.mako
```diff
@@ -0,0 +1,26 @@
...
+READ_ALL_AVAILABLE = -1
+WAIT_INFINITELY = -1.0
+
+
+# Enums
+% for name, enum_metadata in enums.items():
+class ${name}(Enum):
+%   for value in enum_metadata['values']:
+    ${value['name']} = ${value['value']}${enum_helpers.get_enum_value_docstring(value, data)}
+%   endfor
```
bkeryan, line 23: This should use python_name when available.

```suggestion
% for name, enum_metadata in enums.items():
class ${enum_metadata.get('python_name', name)}(Enum):
%   for value in enum_metadata['values']:
    ${value.get('python_name', value['name'])} = ${value['value']}${enum_helpers.get_enum_value_docstring(value, data)}
%   endfor
```

src/codegen/templates/constants.mako
```diff
@@ -0,0 +1,26 @@
...
+# Constants
+AUTO = -1
+CFG_DEFAULT = -1
+DEFAULT = -1
+READ_ALL_AVAILABLE = -1
+WAIT_INFINITELY = -1.0
+
+
+# Enums
+% for name, enum_metadata in enums.items():
```
bkeryan, line 19: @vigkre @MounikaBattu17

`enums.py` now (as of https://github.com/ni/grpc-device-scrapigen/pull/127) contains multiple enums with the same python_name. This template needs to generate the right ones.

```
class ActiveLevel(Enum):
    ABOVE = 10093  #: Pause the measurement or generation while the signal is above the threshold.
    BELOW = 10107  #: Pause the measurement or generation while the signal is below the threshold.


class AngleUnits(Enum):
    DEGREES = 10146  #: Degrees.
    RADIANS = 10273  #: Radians.
    FROM_CUSTOM_SCALE = 10065  #: Units a custom scale specifies. If you select this value, you must specify a custom scale name.


class AngleUnits(Enum):
    DEGREES = 10146  #: Degrees.
    RADIANS = 10273  #: Radians.
    TICKS = 10304  #: Ticks.
    FROM_CUSTOM_SCALE = 10065  #: Units a custom scale specifies. If you select this value, you must specify a custom scale name.


class AngleUnits(Enum):
    DEGREES = 10146  #: Degrees.
    FROM_CUSTOM_SCALE = 10065  #: Units a custom scale specifies. If you select this value, you must specify a custom scale name.


class AngularVelocityUnits(Enum):
    RPM = 16080  #: Revolutions per minute.
    RADIANS_PER_SECOND = 16081  #: Radians per second.
    DEGREES_PER_SECOND = 16082  #: Degrees per second.
    FROM_CUSTOM_SCALE = 10065  #: Units a custom scale specifies. If you select this value, you must specify a custom scale name.
```

src/codegen/generator.py
```diff
@@ -45,3 +45,7 @@ def generate(dest):
         _generate_file(
             codegen_metadata, info["templateFile"], dest / info["relativeOutputPath"]
         )
+
```
bkeryan, line 48: Is this how to run the code generator? `poetry run python -m codegen --dest .\generated\nidaqmx`

We should add that to CONTRIBUTING.md.

bkeryan (changes requested):

src/codegen/templates/constants.mako
bkeryan: Update: I think updating `ENUMS_BLACKLIST` and sorting the enums by python_name will take care of this.

src/codegen/generator.py
bkeryan: This is more relevant for https://github.com/ni/nidaqmx-python/pull/214

generated/nidaqmx/error_codes.py
```diff
@@ -6,6 +6,8 @@
 
 
 class DAQmxErrors(IntEnum):
+    REMOTE_SENSE = -209888
+    OVER_TEMPERATURE_PROTECTION_ACTIVATED = -209887
```
vigkre, line 10: These two Error codes are added because of the https://github.com/ni/grpc-device-scrapigen/pull/136

vigkre, line 10: These two Error codes are added because of the https://github.com/ni/grpc-device-scrapigen/pull/136

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py (FINITE,CONTINUOUS)

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py (ABOVE,BELOW)

generated/nidaqmx/constants.py
MounikaBattu17: `BridgePhysicalUnits.POUNDS_PER_SQ_INCH` is used in nidaqm-python repo. Thus have retained the expected enum value name (`POUNDS_PER_SQ_INCH`)

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: `EncoderType.X_4` is used in nidaqm-python repo. Thus have retained the expected enum value name (`X_4`)

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py(`ON_BOARD_MEMORY_MORE_THAN_HALF_FULL,ON_BOARD_MEMORY_NOT_EMPTY,ONBOARD_MEMORY_CUSTOM_THRESHOLD,WHEN_ACQUISITION_COMPLETE`)

generated/nidaqmx/constants.py
MounikaBattu17: Updated to `YES`, `NO`

generated/nidaqmx/constants.py
MounikaBattu17: Updated.

generated/nidaqmx/constants.py
MounikaBattu17: `PressureUnits.POUNDS_PER_SQ_INCH` is used in nidaqm-python repo. Thus have retained the expected enum value name (`POUNDS_PER_SQ_INCH`)

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Updated.

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Updated

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: `ShuntElementLocation `enum is not used in nidaqm-python. Hence, it is removed.

generated/nidaqmx/constants.py
MounikaBattu17: Resolved.

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Removed unwanted enum values.

generated/nidaqmx/constants.py
MounikaBattu17: Resolved.

generated/nidaqmx/constants.py
MounikaBattu17: Updated to expected values.

generated/nidaqmx/constants.py
MounikaBattu17: `TemperatureUnits.K` is used in nidaqm-python repo. Thus have retained the expected enum value name (`K`)

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Removed.

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: `SoundPressureUnits.PA` is used in nidaqm-python repo. If replaced this `UnitsPreScaled.PA` will affect `SoundPressureUnits`. Thus have retained the enum value name (`PA`).

generated/nidaqmx/constants.py
MounikaBattu17: `POUNDS_PER_SQ_INCH` is used in nidaqm-python repo. Thus have retained the enum value name

generated/nidaqmx/constants.py
MounikaBattu17: Updated enum values as in expected constants.py

generated/nidaqmx/constants.py
MounikaBattu17: Removed.

generated/nidaqmx/constants.py
MounikaBattu17: Removed.

generated/nidaqmx/constants.py
MounikaBattu17: Updated to `WINDOW`.

generated/nidaqmx/constants.py
```diff
@@ -411,28 +418,30 @@ class LVDTSensitivityUnits(Enum):
...
-    CHAN_PER_LINE = 0  #: One Channel For Each Line
     CHAN_FOR_ALL_LINES = 1  #: One Channel For All Lines
+    CHAN_PER_LINE = 0  #: One Channel For Each Line
 
 
 class LoggingMode(Enum):
     OFF = 10231  #: Disable logging for the task.
-    LOG = 15844  #: Enable logging for the task. You cannot read data using DAQmx Read when using this mode. If you require access to the data, read from the TDMS file.
-    LOG_AND_READ = 15842  #: Enable both logging and reading data for the task. You must use DAQmx Read to read samples for NI-DAQmx to stream them to disk.
+    LOG = 15844  #: Enable logging for the task. You cannot read data using an NI-DAQmx Read function when using this mode. If you require access to the data, read from the TDMS file.
```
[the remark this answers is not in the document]
MounikaBattu17: Updated enums.py doesn't have `&DAQmxRead;`

src/codegen/templates/constants.mako
MounikaBattu17: Resolved.

src/codegen/utilities/enum_helpers.py
MounikaBattu17: Updated.

src/codegen/utilities/enum_helpers.py
MounikaBattu17: Updated.

src/codegen/templates/constants.mako
MounikaBattu17: Updated logic in enum_helpers.py to get `'python_name'` or `'name'`.

src/codegen/templates/constants.mako
MounikaBattu17: Updated `ENUMS_BLACKLIST` and logic in `enum_helpers.py` to get desired ouput.

MounikaBattu17:
Please find the differences in generated `constants.py` Vs expected `constants.py` listed below:

1.  Missing Enums -> These enums are not used in nidaqmx-python. Hence removed
```
Action
BreakMode
CalibrationTerminalConfig
EndCalAction
InputCalSource
PathCapability
PowerCalibrationType
RelayPosition
SCXI1124Range
ScanRepeatMode
ShuntElementLocation
ShuntResistorSelect
SoftwareTrigger
SwitchChannelUsage
TaskState
WatchdogTaskAction
WaveformAttributes
_Callback
```

2. Missing enum values -> These enum values are not used in nidaqmx-python. Hence removed
`BridgeConfiguration(QUARTER_BRIDGE_120_OHM_COMPLETION_RESISTOR,QUARTER_BRIDGE_350_OHM_COMPLETION_RESISTOR)`

3. Docstrings differences -> These enum have are docstrings differences
```
DataTransferActiveTransferMode
EncoderType
LoggingMode
OverWtiteMode
ResistorState
TorqueUnits
WriteBasicTEDSOptions
```

src/codegen/utilities/enum_helpers.py
vigkre: Already added in `master `branch

generated/nidaqmx/constants.py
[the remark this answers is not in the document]
vigkre: As discussed, we can order these enums by enum value or name but that will result in massive changes when we diff. So, planning to not sort these now for better code review.

generated/nidaqmx/constants.py
vigkre: Sure Brad. Most of inconsistencies are resolved and with the updated enums metadata, we have the changes in `constants.py` categorized as mentioned in this https://github.com/ni/nidaqmx-python/pull/215#issuecomment-1468081159

src/codegen/utilities/enum_helpers.py
vigkre: We have handled the docstrings using this function and look like we don't have any case where enum name and enum docstring have the same substitutions.https://github.com/ni/nidaqmx-python/blob/9e1b171a01bf9fd7bd057a55ec70eb620ce9548e/src/codegen/utilities/enum_helpers.py#L165

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,192 @@
...
+    "WatchdogControlAction",
+    "VelocityUnits2",
+]
+
+
+# Metadata issues or invalid Python names (leading number)
+NAME_SUBSTITUTIONS = {
+    '100_MHZ_TIMEBASE': 'ONE_HUNDRED_MHZ_TIMEBASE',
+    '20_MHZ_TIMEBASE': 'TWENTY_MHZ_TIMEBASE',
+    '2_POINT_5_V': 'TWO_POINT_FIVE_V',
```
bkeryan, line 36: Do any attribute/function docstrings reference these values? If so, it would be better to do these substitutions in scrapigen so that the docstrings have the substituted values.

bkeryan (changes requested):

generated/nidaqmx/constants.py
```diff
@@ -62,11 +62,6 @@ class AcquisitionType(Enum):
     HW_TIMED_SINGLE_POINT = 12522  #: Acquire or generate samples continuously using hardware timing without a buffer. Hardware timed single point sample mode is supported only for the sample clock and change detection timing types.
 
 
-class Action(Enum):
-    COMMIT = 0  #: Commit
-    CANCEL = 1  #: Cancel
-
-
 class ActiveLevel(Enum):
```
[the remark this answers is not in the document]
zhindes: Nope, no concerns.

generated/nidaqmx/constants.py
```diff
@@ -97,17 +92,10 @@ class AutoZeroType(Enum):
...
-    QUARTER_BRIDGE = 10270  #: Sensor is a quarter bridge. If you set **ai_excit_use_for_scaling** to True, NI-DAQmx divides the measurement by the excitation value. Many sensors scale data to native units using scaling of volts per excitation.
-    QUARTER_BRIDGE_120_OHM_COMPLETION_RESISTOR = 16163
-    QUARTER_BRIDGE_350_OHM_COMPLETION_RESISTOR = 16164
+    FULL_BRIDGE = 10182  #: Sensor is a full bridge. If you set **ai_excit_use_for_scaling** to TRUE, NI-DAQmx divides the measurement by the excitation value. Many sensors scale data to native units using scaling of volts per excitation.
+    HALF_BRIDGE = 10187  #: Sensor is a half bridge. If you set **ai_excit_use_for_scaling** to TRUE, NI-DAQmx divides the measurement by the excitation value. Many sensors scale data to native units using scaling of volts per excitation.
+    QUARTER_BRIDGE = 10270  #: Sensor is a quarter bridge. If you set **ai_excit_use_for_scaling** to TRUE, NI-DAQmx divides the measurement by the excitation value. Many sensors scale data to native units using scaling of volts per excitation.
```
bkeryan, line 98: Python documentation should use `True`, not `TRUE`

generated/nidaqmx/constants.py
```diff
@@ -480,30 +452,15 @@ class OverflowBehavior(Enum):
 
 
 class OverwriteMode(Enum):
-    OVERWRITE_UNREAD_SAMPLES = 10252  #: When an acquisition encounters unread data in the buffer, the acquisition continues and overwrites the unread samples with new ones. You can read the new samples by setting **relative_to** to **ReadRelativeTo.MOST_RECENT_SAMPLE** and setting **offset** to the appropriate number of samples.
+    OVERWRITE_UNREAD_SAMPLES = 10252  #: When an acquisition encounters unread data in the buffer, the acquisition continues and overwrites the unread samples with new ones. You can read the new samples by setting **relative_to** to **MOST_RECENT_SAMP** and setting **offset** to the appropriate number of samples.
```
bkeryan, line 455: `**ReadRelativeTo.MOST_RECENT_SAMPLE**` was correct. `**MOST_RECENT_SAMP**` is incorrect.

generated/nidaqmx/constants.py
```diff
@@ -868,23 +774,24 @@ class TimestampEvent(Enum):
...
-    FOOT_POUNDS = 15884  #: Pound-feet.
-    INCH_POUNDS = 15883  #: Pound-inches.
     INCH_OUNCES = 15882  #: Ounce-inches.
+    INCH_POUNDS = 15883  #: Pound-inches.
+    FOOT_POUNDS = 15884  #: Pound-feet.
     FROM_CUSTOM_SCALE = 10065  #: Units a custom scale specifies. If you select this value, you must specify a custom scale name.
 
 
 class TriggerType(Enum):
-    NONE = 10230  #: Disable reference triggering for the task.
```
bkeryan, line 878: The old docstring was incorrect because it referred to reference triggering. The new docstring is correct.

generated/nidaqmx/constants.py
```diff
@@ -1067,20 +963,16 @@ class WindowTriggerCondition2(Enum):
 
 
 class WriteBasicTEDSOptions(Enum):
-    WRITE_TO_EEPROM = 12538  #: blah
-    WRITE_TO_PROM = 12539  #: blah
-    DO_NOT_WRITE = 12540  #: blah
+    WRITE_TO_EEPROM = 12538  #: Write basic TEDS data to the EEPROM, even if the sensor includes a PROM. You cannot write basic TEDS data if the PROM contains data.
+    WRITE_TO_PROM = 12539  #: Write basic TEDS data to the PROM. Any subsequent attempts to write basic TEDS data result in an error.
+    DO_NOT_WRITE = 12540  #: Ignore basic TEDS data.
```
bkeryan, line 968: 👍

generated/nidaqmx/constants.py
```diff
@@ -272,7 +255,7 @@ class EncoderType(Enum):
     X_1 = 10090  #: If signal A leads signal B, count the rising edges of signal A. If signal B leads signal A, count the falling edges of signal A.
     X_2 = 10091  #: Count the rising and falling edges of signal A.
     X_4 = 10092  #: Count the rising and falling edges of signal A and signal B.
-    TWO_PULSE_COUNTING = 10313  #: Two pulse counting.
+    TWO_PULSE_COUNTING = 10313  #: Increment the count on rising edges of signal A. Decrement the count on rising edges of signal B.
```
bkeryan, line 258: 👍

zhindes (changes requested):
The generated code changes look great, thanks for all the hard work there. Once Brad is happy with the generator changes, I'll be happy :)

generated/nidaqmx/constants.py
```diff
@@ -80,8 +75,8 @@ class ActiveOrInactiveEdgeSelection(Enum):
 class AngleUnits(Enum):
     DEGREES = 10146  #: Degrees.
     RADIANS = 10273  #: Radians.
-    TICKS = 10304  #: Ticks.
     FROM_CUSTOM_SCALE = 10065  #: Units a custom scale specifies. If you select this value, you must specify a custom scale name.
+    TICKS = 10304  #: Ticks.
```
bkeryan, line 79: FYI: Python 3 enums remember the order in which values are defined and allow iterating over the values in definition order at run time.
https://docs.python.org/3/library/enum.html#enum.EnumType.__iter__

However, Python 2 enums didn't remember definition order by default. You had to set the `_order_` attribute. So these enums were not always ordered.

Changing the enum value order breaks compatibility, but very few DAQmx applications are likely to depend on this, and upgrading from Python 2 to Python 3 already changed this behavior.

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,192 @@
+import html
+import logging
+import re
+
+from html.parser import HTMLParser
+_logger = logging.getLogger(__name__)
+_logger.addHandler(logging.NullHandler())
```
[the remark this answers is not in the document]
MounikaBattu17: Moved the `_logger` logic from submodules to  `_init_.py` and removed unused` _logger` variables.

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,192 @@
...
+def _merge_enum_values(valueses):
+    result_set = {}
+
+    for values_array in valueses:
+        for value in values_array:
+
+            enum_values = {"name": "", "value": "",
+                           "documentation": {"description": ""}}
+            enum_values["name"] = value.get("python_name", value["name"])
+            enum_values["value"] = value["value"]
```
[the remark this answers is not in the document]
MounikaBattu17: Updated.

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,192 @@
...
+    raw_docstring = get_attr_enum_docstring(
+        enum_value['documentation']['description'])
+    raw_docstring = cleanup_docstring(raw_docstring)
+
+    if raw_docstring != '':
+        return f"  #: {raw_docstring}"
+    return ""
+
+
+def cleanup_docstring(docstring):
```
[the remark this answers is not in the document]
MounikaBattu17: Prefixed `_` to all the private methods.

src/codegen/utilities/enum_helpers.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def get_enums(metadata):
+    enums = metadata['enums']
+
+    # First remove enums we don't use.
+    enums = {name: val for (name, val) in enums.items() if name not in ENUMS_BLACKLIST}
+    # Then merge variants.
+    enums = _merge_enum_variants(enums)
+    return _sanitize_values(enums)
```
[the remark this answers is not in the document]
MounikaBattu17: Extracted `enums` with a `deepcopy` and updated the `copy enums`.

src/codegen/utilities/enum_helpers.py
MounikaBattu17: The `NAME_SUBSTITUTIONS` were not used in docstrings of attributes/functions.

generated/nidaqmx/constants.py
MounikaBattu17: Resolved.

generated/nidaqmx/constants.py
zhindes: Yes, don't do it here. It would make reviewing the changes a nightmare

zhindes (approved):
I'll merge once Brad gives his approval!

generated/nidaqmx/constants.py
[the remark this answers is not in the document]
MounikaBattu17: @bkeryan Can we please this conversation?

MounikaBattu17:
The newly added '`python_description`' in enums.py and '`python_enum`' in attributes.py is handled in grpc-device https://github.com/ni/grpc-device/pull/886

generated/nidaqmx/constants.py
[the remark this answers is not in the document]
bkeryan: yes

bkeryan (approved):
Approved with suggestions.

src/codegen/properties/attribute.py
```diff
@@ -55,6 +55,9 @@ def __init__(self, id, attribute_metadata, enum_merge_set):
         if "enum" in attribute_metadata:
             self._enum = self.merge_enums(attribute_metadata["enum"], enum_merge_set)
             self._is_enum = True
+        elif "python_enum" in attribute_metadata:
+            self._python_enum = attribute_metadata["python_enum"]
+            self._is_enum = True
```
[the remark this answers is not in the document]
MounikaBattu17: Implemented the suggestion.

src/codegen/properties/attribute.py
```diff
@@ -133,7 +136,7 @@ def enum(self):
         This key will only be available for an enum type attribute.
         During code generation an attribute would be considered as an enum if it contains this key.
         """
-        return self._enum
+        return self._enum if hasattr(self, "_enum") else self._python_enum
```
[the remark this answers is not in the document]
MounikaBattu17: Implemented the suggestion.

MounikaBattu17:
@zhindes Brad has given his approval, can you please merge this PR?

PR #221, opened 2023-03-16
Generate ao_channel.py file

subash-suresh:
This PR adds the necessary mako template for generating the `ao_channel.py` and updates the generated file.

- Added the mako for generating attributes as ao_channel.py contains only attributes
- Added the latest attribute_metadata from scrapigen
- Updated the generated folder with the latest `ao_channel.py`
- The ai_channel.py was regenerated using the updated metadata


src/codegen/templates/_task_modules/channels/ao_channel.py.mako
```diff
@@ -0,0 +1,37 @@
...
+    __slots__ = []
+
+    def __repr__(self):
+        return 'AOChannel(name={0})'.format(self._name)
+
+<%namespace name="property_template" file="/property_template.py.mako"/>\
+%for attribute in attributes:
+${property_template.script_property(attribute)}\
+%endfor
+<%namespace name="deprecated_template" file="/property_deprecated_template.py.mako"/>\
```
keer1111, line 36: is deprecation need for this ao channel, since there are no property methods deprecated? or intended to add for any future deprecation attributes comes via metadata?

keer1111 (approved):

generated/nidaqmx/_task_modules/channels/ao_channel.py
```diff
@@ -127,7 +128,7 @@ def ao_current_units(self):
         check_for_error(error_code)
 
     @property
-    def ao_custom_scale(self):
+    def ao_custom_scale_name(self):
```
zhindes, line 131: The old name is correct, this returns a `Scale` object, not the scale name. The same is true for AI. I know we've talked about this before. So follow whatever we did there.

generated/nidaqmx/_task_modules/channels/ao_channel.py
```diff
@@ -1820,8 +1821,9 @@ def ao_output_impedance(self):
     @property
     def ao_output_type(self):
         """
-        :class:`nidaqmx.constants.UsageTypeAO`: Indicates whether the
-            channel generates voltage,  current, or a waveform.
+        :class:`nidaqmx.constants.AOOutputChannelType`: Indicates
```
zhindes, line 1824: Its still `UsageTypeAO` in the constants.py PR #215. I don't think this is correct.

generated/nidaqmx/_task_modules/channels/ao_channel.py
```diff
@@ -1969,8 +1971,8 @@ def ao_resolution_units(self):
     @property
     def ao_term_cfg(self):
         """
-        :class:`nidaqmx.constants.TerminalConfiguration`: Specifies the
-            terminal configuration of the channel.
+        :class:`nidaqmx.constants.OutputTermCfg`: Specifies the terminal
```
zhindes, line 1974: Its still `TerminalConfiguration` in #215

generated/nidaqmx/_task_modules/channels/ao_channel.py
```diff
@@ -2218,8 +2220,8 @@ def ao_voltage_current_limit(self):
     @property
     def ao_voltage_units(self):
         """
-        :class:`nidaqmx.constants.VoltageUnits`: Specifies in what units
-            to generate voltage on the channel. Write data to the
+        :class:`nidaqmx.constants.VoltageUnits2`: Specifies in what
```
zhindes, line 2223: Shouldn't change.

zhindes (changes requested):
A few changes that shouldn't have been done. A lot of effort went into #215 to ensure we didn't break compatibility too bad. Make sure you're aligned with that!

src/codegen/templates/_task_modules/channels/ao_channel.py.mako
samchris007: Removed logic related to deprecation in **ao_channel.py.mako**

generated/nidaqmx/_task_modules/channels/ao_channel.py
samchris007: Sure. Updated.

generated/nidaqmx/_task_modules/channels/ao_channel.py
```diff
@@ -175,8 +176,8 @@ def ao_custom_scale(self, val):
             self._handle, self._name, val)
         check_for_error(error_code)
 
-    @ao_custom_scale.deleter
-    def ao_custom_scale(self):
+    @ao_custom_scale_name.deleter
+    def ao_custom_scale_name(self):
```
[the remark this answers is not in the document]
samchris007: Updated.

generated/nidaqmx/_task_modules/channels/ao_channel.py
samchris007: Ok. Updated

generated/nidaqmx/_task_modules/channels/ao_channel.py
samchris007: Updated.

generated/nidaqmx/_task_modules/channels/ao_channel.py
```diff
@@ -160,8 +161,8 @@ def ao_custom_scale(self):
 
         return Scale(val.value.decode('ascii'))
 
-    @ao_custom_scale.setter
-    def ao_custom_scale(self, val):
+    @ao_custom_scale_name.setter
+    def ao_custom_scale_name(self, val):
```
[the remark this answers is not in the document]
samchris007: Updated

generated/nidaqmx/_task_modules/channels/ao_channel.py
samchris007: Updated.

zhindes (approved):
Looks great. I'll give @bkeryan a chance to look it over before I merge.

generated/nidaqmx/_task_modules/channels/ao_channel.py
```diff
@@ -2266,3 +2266,39 @@ def ao_voltage_units(self):
...
 
+    @property
+    def physical_chan_ao_supported_power_up_output_types(self):
+        """
+        :class: list[`nidaqmx.constants.AOPowerUpOutputBehavior`]:
+            Indicates the power up output types supported by the
+            channel.
+        """
+        cfunc = (lib_importer.windll.
+                 DAQmxGetPhysicalChanAOSupportedPowerUpOutputTypes)
```
bkeryan, line 2277: This property does not belong in the `AOChannel` class because it is a physical channel property, not a channel property:
```
int32 __CFUNC DAQmxGetPhysicalChanAOSupportedPowerUpOutputTypes(const char physicalChannel[], int32 *data, uInt32 arraySizeInElements);
```

For the Python API, all of the physical channel properties belong here: https://github.com/ni/nidaqmx-python/blob/master/src/nidaqmx/system/physical_channel.py

See https://www.ni.com/docs/en-US/bundle/ni-daqmx/page/mxcncpts/chans.html for the difference between physical channels and (virtual) channels. 

`AOChannel._name` is a virtual channel name, so it isn't guaranteed to be a valid physical channel name because users may assign a different name when adding the channel to a task.

bkeryan (changes requested):

generated/nidaqmx/_task_modules/channels/ao_channel.py
bkeryan: The metadata is incorrect. Compare:

`PhysicalChanAOSupportedOutputTypes`: https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/attributes.py#L22067
`'python_class_name': 'PhysicalChannel',`

`PhysicalChanAOSupportedPowerUpOutputTypes`: https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/attributes.py#L22211
`'python_class_name': 'AOChannel',`

samchris007: Understood. 
- Created a new PR in **grpc-device-scrapigen** which updates `python_class_name` of property `physical_chan_ao_supported_power_up_output_types` to `PhysicalChannel`.  
https://github.com/ni/grpc-device-scrapigen/pull/148
- Used the same metadata (**attributes,py**) to generate **ao_channel.py**.

bkeryan (approved):

PR #225, opened 2023-03-21
[Internal] Generate InStream and OutStream Classes in DAQmx 

Sakthi-SM:
* Generates `in_stream.py` and `out_stream.py`.
* Added the necessary `mako` templates and updated the metadata
* The following changes have been observed with the newly generated files:

Current Name | Changed To
-- | --
change_detect_overflowed | change_detect_has_overflowed
di_num_booleans_per_chan | digital_lines_bytes_per_chan
over_write | overwrite
do_num_booleans_per_chan | digital_lines_bytes_per_chan

</body></html></div>
</body>
</html>

Implements [Task 2316156](https://dev.azure.com/ni/DevCentral/_workitems/edit/2316156): Generate the in_stream.py and out_stream.py files


generated/nidaqmx/_task_modules/in_stream.py
```diff
@@ -227,7 +227,7 @@ def avail_samp_per_chan(self):
         return val.value
 
     @property
-    def change_detect_overflowed(self):
+    def change_detect_has_overflowed(self):
```
zhindes, line 230: man these are hard... LV uses the shortened version, and C tends to use the phrases. I think I like the shortened version better. For example, we use `aux_power_error_chans_exist` instead of `aux_power_error_chans_do_exist` in Python and C. So I think the shortened one is more consistent, lets use `change_detect_overflowed`

generated/nidaqmx/_task_modules/in_stream.py
```diff
@@ -1625,6 +1577,54 @@ def overtemperature_chans_exist(self):
 
         return val.value
 
+    @property
+    def overwrite(self):
```
zhindes, line 1581: `over_write` was definitely wrong, add a deprecation for `over_write`

generated/nidaqmx/_task_modules/in_stream.py
Sakthi-SM: Reverted back to the older name

generated/nidaqmx/_task_modules/in_stream.py
Sakthi-SM: add `over_write` to deprecated list of items

zhindes (approved):
I will merge once brad is happy

bkeryan (approved):

PR #226, opened 2023-03-22
[Internal] Generate watch dog module related classes

Sakthi-SM:
* Generates `watchdog.py` and `expirationstate.py`.
* Added the necessary `mako` templates and updated the metadata
* The following changes have been observed with the newly generated files:

Current Name | Changed To
-- | --
expir_states_ao_state | ao_expir_state
expir_states_ao_type | ao_output_type
expir_states_co_state | co_expir_state
expir_states_do_state | do_expir_state
expir_trig_dig_edge_edge | dig_edge_watchdog_expir_trig_edge
expir_trig_dig_edge_src | dig_edge_watchdog_expir_trig_src
expir_trig_trig_type | expir_trig_type
expired | has_expired

</body></html></div>
</body>
</html>

Implements [Task 2316170](https://dev.azure.com/ni/DevCentral/_workitems/edit/2316170): Generate the watchdog.py and related files


generated/nidaqmx/system/watchdog.py
```diff
@@ -254,7 +252,7 @@ def expir_trig_trig_on_network_conn_loss(self):
         check_for_error(error_code)
 
     @property
-    def expir_trig_trig_type(self):
+    def expir_trig_type(self):
```
zhindes, line 255: I like the way we handle triggers for normal tasks, with the trigger object for each type: https://github.com/ni/nidaqmx-python/blob/master/src/nidaqmx/_task_modules/triggers.py. however, that is a bit overkill for this since there is only one option as far as I can tell: digital. Time is an option, but I don't think there is a way to configure it, so that's cool.

Anyways, the analog for this in a normal trigger is: `triggers.arm_start_trigger.trig_type`. Since we don't have an object, I don't think we need to repeat `trig`, so I like this: `expiration_trigger_type`

generated/nidaqmx/system/watchdog.py
```diff
@@ -302,7 +300,7 @@ def expir_trig_trig_type(self):
         check_for_error(error_code)
 
     @property
-    def expired(self):
+    def has_expired(self):
```
zhindes, line 303: no phrase, just `expired`

generated/nidaqmx/system/watchdog.py
```diff
@@ -94,7 +92,7 @@ def expiration_states(self):
         return self._expiration_states
 
     @property
-    def expir_trig_dig_edge_edge(self):
+    def dig_edge_watchdog_expir_trig_edge(self):
```
zhindes, line 95: normal task trigger name would be something like this `triggers.arm_start_trigger.dig_edge_edge`, so lets go with `expiration_trigger_dig_edge_edge`

generated/nidaqmx/system/watchdog.py
```diff
@@ -142,7 +140,7 @@ def expir_trig_dig_edge_edge(self):
         check_for_error(error_code)
 
     @property
-    def expir_trig_dig_edge_src(self):
+    def dig_edge_watchdog_expir_trig_src(self):
```
zhindes, line 143: `expiration_trigger_dig_edge_src`

generated/nidaqmx/system/watchdog.py
```diff
@@ -188,8 +186,8 @@ def expir_trig_dig_edge_src(self, val):
             self._handle, val)
         check_for_error(error_code)
 
-    @expir_trig_dig_edge_src.deleter
-    def expir_trig_dig_edge_src(self):
+    @dig_edge_watchdog_expir_trig_src.deleter
+    def dig_edge_watchdog_expir_trig_src(self):
         cfunc = lib_importer.windll.DAQmxResetDigEdgeWatchdogExpirTrigSrc
         if cfunc.argtypes is None:
             with cfunc.arglock:
```
zhindes, line 193: I can't make a comment on the network_conn_loss, but I think we should change it...

`expiration_trigger_trig_on_network_conn_loss`

zhindes (changes requested):
OK, this one is a mess. The existing names are bad and repetitive. Here is an example I wrote up using the existing API.

```python
from nidaqmx.system.watchdog import WatchdogTask
from nidaqmx.types import AOExpirationState
from nidaqmx.constants import WatchdogAOExpirState

def print_expir_state(phys_chan):
    ao_state = watchdog_task.expiration_states[phys_chan].expir_states_ao_state
    ao_type = watchdog_task.expiration_states[phys_chan].expir_states_ao_type
    print(f"state: {ao_state}")
    print(f"type: {ao_type}")


with WatchdogTask("cDAQ1") as watchdog_task:
    print("original state")
    print_expir_state("cDAQ1Mod1/ao0")

    new_state = AOExpirationState("cDAQ1Mod1/ao0", 1.0, WatchdogAOExpirState.VOLTAGE)
    watchdog_task.cfg_watchdog_ao_expir_states([new_state])

    print("new state")
    print_expir_state("cDAQ1Mod1/ao0")

    watchdog_task.expir_trig_trig_on_network_conn_loss = True
    watchdog_task.start()
    watchdog_task.stop()
```

My suggestions are an attempt to clean that up a bit. I would appreciate thoughts from @bkeryan .

zhindes (approved):
lets see if brad has any recommendations on naming

bkeryan (approved):
Please update the names in the PR description before committing.

src/codegen/utilities/attribute_helpers.py
```diff
@@ -94,12 +94,27 @@
...
+    "expir_states_ao_type": {"new_name": "ao_output_type", "deprecated_in": "0.6.6"},
+    "expir_states_co_state": {"new_name": "co_state", "deprecated_in": "0.6.6"},
+    "expir_states_do_state": {"new_name": "do_state", "deprecated_in": "0.6.6"},
+    "expir_states_ao_state": {"new_name": "ao_state", "deprecated_in": "0.6.6"},
+    "expir_trig_trig_type": {"new_name": "expiration_trigger_type", "deprecated_in": "0.6.6"},
+    "expir_trig_dig_edge_edge": {
+        "new_name": "expiration_trigger_dig_edge_edge",
+        "deprecated_in": "0.6.6",
+    },
+    "expir_trig_dig_edge_src": {"new_name": "expiration_trigger_type", "deprecated_in": "0.6.6"},
```
bkeryan, line 106: The new name is incorrect.
```suggestion
    "expir_trig_dig_edge_src": {"new_name": "expiration_trigger_dig_edge_src", "deprecated_in": "0.6.6"},
```

bkeryan (changes requested):

generated/nidaqmx/system/watchdog.py
```diff
@@ -16,8 +17,6 @@
 from nidaqmx.utils import flatten_channel_string
 from nidaqmx.constants import (
     Edge, TriggerType, WDTTaskAction)
-from nidaqmx.types import (
```
bkeryan, line 19: Removing this broke compatibility with the tests. I will re-add it.

PR #232, opened 2023-03-24
Add mako templates for handwritten files

subash-suresh:
Copies the handwritten files in nidaqmx to the generated folder.

- Created a new directory `src\handwritten` and added the handwritten files here.
- Updated the `generator.py` logic to copy the handwritten files to the generated folder.
- Added logic in the toml file to exclude lint error checks for the handwritten files

note: The `__init__.py files are also considered handwritten.


subash-suresh:
The source path of the handwritten file is currently the original nidaqmx code, but this can be changed based on where we decide to keep these files.

src/codegen/metadata/script_info.py
```diff
@@ -28,5 +28,123 @@
             "relativeOutputPath": "_task_modules\\channels\\do_channel.py", 
             "templateFile": "_task_modules\\channels\\do_channel.py.mako",
         },
+    ],
+    "modulesToCopy":[
```
keer1111, line 32: instead can we copy all files and skip files that are already present?

src/codegen/generator.py
```diff
@@ -41,9 +52,12 @@ def generate(dest):
...
 
-    for info in codegen_metadata["script_info"]["modules"]:
+    for info in codegen_metadata["script_info"]["modulesToGenerate"]:
         _generate_file(codegen_metadata, info["templateFile"], dest / info["relativeOutputPath"])
 
     _generate_file(codegen_metadata["enums"], "error_codes.mako", dest / "error_codes.py")
 
     _generate_file(codegen_metadata, "constants.mako", dest / "constants.py")
+
+    for info in codegen_metadata["script_info"]["modulesToCopy"]:
```
keer1111, line 62: the for can be inside the copy method and finding parent directory can be outside of for loop

keer1111 (changes requested):

src/codegen/metadata/script_info.py
```diff
@@ -28,5 +28,123 @@
...
+        {
+            "relativeOutputPath": "__init__.py",
```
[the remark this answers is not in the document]
subash-suresh: That's right both are same now, but I just have that since I wanted to get some suggestions on where the handwritten files can be placed in the src folder.  Based on this we can plan to remove the repeated field.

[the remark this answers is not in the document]
subash-suresh: The logic has been updated to have mako files for handwritten files

src/codegen/metadata/script_info.py
subash-suresh: updated the code to use mako files

zhindes (changes requested):
Not sure how I feel about a bunch of mako files with no generation... what do we think of creating `src/handwritten` directory with all these files as simply .py files that we copy as part of the generation process? @bkeryan ?

src/handwritten/__init__.py
```diff
@@ -0,0 +1,13 @@
+from nidaqmx.errors import DaqError, DaqReadError, DaqWriteError, DaqWarning, DaqResourceWarning
```
keer1111, line 1: should the files in the other folder "nidaqmx" be removed? So it shows as moved, but not additions?

keer1111 (approved):

src/handwritten/__init__.py
subash-suresh: I think the removal of the original nidaqmx code from the source folder can be done once we have generated all nidaqmx files and the existing tests in the repo is passed.

zhindes: Yep, we're gonna wait to do that until all the codegen is done

bkeryan:
Looks good to me.
image

bkeryan (approved):

PR #236, opened 2023-03-29
[Internal] Addition of tests for properties in persisted scale

Sakthi-SM:
Added test cases for adding `string` and `boolean` properties in persisted scale.

This implements [Task 2330444](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330444): Add or verify tests for Persisted Scale property and related data types

### DIFF
--- a/tests/test_persisted_scale_property.py
+++ b/tests/test_persisted_scale_property.py
@@ -1,0 +1,23 @@
+"Contains a collection of pytest tests that validates the persisted scale properties."
+
+import nidaqmx
+import nidaqmx.system.system
+from nidaqmx.tests.test_read_write import TestDAQmxIOBase
+
+
+class TestPersistedScaleProperty(TestDAQmxIOBase):
+    """Contains a collection of pytest tests that validates the persisted scale properties."""
+
+    def test_bool_property(self):
+        """Test for validating bool attributes in persisted scale."""
+
+        system = nidaqmx.system.System.local()
+        persisted_scale = system.scales[0]
+        assert persisted_scale.allow_interactive_editing
+
+    def test_string_property(self):
+        """Test for validating string attributes in persisted scale."""
+
+        system = nidaqmx.system.System.local()
+        persisted_scale = system.scales[0]
+        persisted_scale.author


tests/test_persisted_scale_property.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class TestPersistedScaleProperty(TestDAQmxIOBase):
+    """Contains a collection of pytest tests that validates the persisted scale properties."""
+
+    def test_bool_property(self):
+        """Test for validating bool attributes in persisted scale."""
+        system = nidaqmx.system.System.local()
+        persisted_scale = system.scales[0]
+        assert persisted_scale.allow_interactive_editing
```
Sakthi-SM, line 15: @zhindes Will `allow_interactive_editing` be always true, irrespective of the instrument tested?

tests/test_persisted_scale_property.py
```diff
@@ -0,0 +1,21 @@
...
+from nidaqmx.tests.test_read_write import TestDAQmxIOBase
+
+
+class TestPersistedScaleProperty(TestDAQmxIOBase):
+    """Contains a collection of pytest tests that validates the persisted scale properties."""
+
+    def test_bool_property(self):
+        """Test for validating bool attributes in persisted scale."""
+        system = nidaqmx.system.System.local()
+        persisted_scale = system.scales[0]
```
subash-suresh, line 14: Is it fine to take the first scale from the ni max configuration or would it be better to take the scale based on name, so that we could make sure we are not taking any unknown scale from NI max

subash-suresh (approved):

tests/test_persisted_scale_property.py
```diff
@@ -0,0 +1,21 @@
...
+import nidaqmx.system.system
+from nidaqmx.tests.test_read_write import TestDAQmxIOBase
+
+
+class TestPersistedScaleProperty(TestDAQmxIOBase):
+    """Contains a collection of pytest tests that validates the persisted scale properties."""
+
+    def test_bool_property(self):
+        """Test for validating bool attributes in persisted scale."""
+        system = nidaqmx.system.System.local()
```
subash-suresh, line 13: This could be moved into a fixture to avoid code repetition.

tests/test_persisted_scale_property.py
bkeryan: This property is based on one of the inputs to `DAQmx Save Scale`: https://www.ni.com/docs/en-US/bundle/ni-daqmx-labview-api-ref/page/lvdaqmx/mxsavescale.html

tests/test_persisted_scale_property.py
```diff
@@ -0,0 +1,21 @@
...
+        persisted_scale = system.scales[0]
+        assert persisted_scale.allow_interactive_editing
+
+    def test_string_property(self):
+        """Test for validating string attributes in persisted scale."""
+        system = nidaqmx.system.System.local()
+        persisted_scale = system.scales[0]
+        persisted_scale.author
```
bkeryan, line 21: Here is how to validate this attribute:
- Add the `Author` attribute to one of the scales in the `.ini` file. You can edit the scale in MAX and export the `.ini` to see what the correct format is.
- Change this test to create a scale object for a specific scale by name, e.g. `nidaqmx.system.Scale("polynomial_scale")`
- Assert that the `author` property matches the `.ini` file's `Author` attribute.

Likewise, you can test a float property by checking `scale.lin_slope` for the scale named "double_gain_scale" and you can test a list[float] property by checking `scale.poly_forward_coeff` for the scale named "polynomial_scale".

bkeryan (changes requested):

tests/conftest.py
```diff
@@ -178,3 +178,12 @@ def multi_threading_test_devices():
...
     return None
+
+
+@pytest.fixture(scope="module")
+def persisted_scale(request):
+    """Gets the device information based on the device name."""
+    system = nidaqmx.system.System.local()
+    if request.param in system.scales:
+        return system.scales[request.param]
+    return None
```
bkeryan, line 189: If the scale is not found, skip the test with a descriptive message like the other fixtures.

tests/test_persisted_scale_property.py
bkeryan: Re: `lin_scope` and `poly_forward_coeff`: I see those properties are in the `Scale` class not the `PersistedScale` class, so they would go in a different test.

bkeryan (changes requested):

tests/test_persisted_scale_property.py
```diff
@@ -0,0 +1,18 @@
...
+import pytest
+
+from nidaqmx.tests.test_read_write import TestDAQmxIOBase
+
+
+class TestPersistedScaleProperty(TestDAQmxIOBase):
+    """Contains a collection of pytest tests that validates the persisted scale properties."""
+
+    @pytest.mark.parametrize("persisted_scale", ["double_gain_scale"], indirect=True)
+    def test_bool_property(self, persisted_scale):
```
bkeryan, line 11: I'm sorry, I forgot to bring up NI's regression testing conventions: https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions

Rename to something like `test__persisted_scale__get_bool_property__returns_value`

tests/test_persisted_scale_property.py
```diff
@@ -0,0 +1,18 @@
+"Contains a collection of pytest tests that validates the persisted scale properties."
+import pytest
+
+from nidaqmx.tests.test_read_write import TestDAQmxIOBase
+
+
+class TestPersistedScaleProperty(TestDAQmxIOBase):
+    """Contains a collection of pytest tests that validates the persisted scale properties."""
```
bkeryan, line 8: Remove the test class. Test functions should be sufficient.

bkeryan (changes requested):

zhindes (approved):
brad covered it, once his issues are resolved, we're good!

bkeryan (approved):

PR #237, opened 2023-03-29
Addtion of Tests for device property

subash-suresh:
- Adds Tests for testing the different datatypes of device properties.
- The tests added only test the getters of the properties as all the device properties are read-only
- The device utilized for this testing is the `PXIe-4331` which was already provided by the `bridge_device` fixture`. All the tests rely on the `PXIe-4331's` device properties.
    -  Added a new file `test_device_properties.py`.
    - Modified the `nidaqmxMaxConfig.ini` to include an accessary device under the `bridgeTester` device. This was required for testing one of the properties.


tests/test_device_properties.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def test_string_property(self, bridge_device):
+        """Test for validating a device property of string type."""
+        assert bridge_device.product_type == "PXIe-4331"
+
+    def test_enum_property(self, bridge_device):
+        """Test for validating a device property of enum type."""
+        assert bridge_device.bus_type == BusType.PXIE
+
+    def test_list_of_float_property(self, bridge_device):
```
subash-suresh, line 33: The properties which returned a list of float values where mostly properties which returned a range like current/voltage ranges and when attempting to read such properties the value returned was always an empty array. Hence the test also checks the same. 
Please let me know if there is a way to actual values with a simulated device.

tests/max_config/nidaqmxMaxConfig.ini
```diff
@@ -29,8 +29,10 @@ DevSerialNum = 0x0
 DevIsSimulated = 1
 ProductNum = 0x74A9C4C4
 BusType = PXIe
-PXI.ChassisNum = 1
-PXI.SlotNum = 2
+PXI.ChassisNum = 4294967295
+PXI.SlotNum = 4294967295
```
zhindes, line 33: please revert this change. you can just do it manually, the import should still work

tests/test_device_properties.py
```diff
@@ -0,0 +1,56 @@
...
+        assert bridge_device.anlg_trig_supported
+
+    def test_float_property(self, bridge_device):
+        """Test for validating the device property of float type."""
+        # The ai max single channel rate of the NI PXIe-4331 device is 102400.0
+        assert bridge_device.ai_max_single_chan_rate == 102400.0
+
+    def test_uint_property(self, bridge_device):
+        """Test for validating the device property of uint type."""
+        assert bridge_device.product_num == 1957283012
```
zhindes, line 24: Product Number constants are typically hex: `0x74A9C4C4`

tests/test_device_properties.py
```diff
@@ -0,0 +1,56 @@
...
+        assert len(ai_trigger_usage) == 2
+        assert ai_trigger_usage == [TriggerUsage.REFERENCE, TriggerUsage.START]
+
+    def test_list_of_uint_property(self, bridge_device):
+        """Test for validating the device property of uint array type."""
+        accessory_product_numbers = bridge_device.accessory_product_nums
+
+        assert isinstance(accessory_product_numbers, list)
+        assert len(accessory_product_numbers) == 1
+        assert accessory_product_numbers == [31122]
```
zhindes, line 56: 0x7992

zhindes (changes requested):

tests/test_device_properties.py
```diff
@@ -0,0 +1,56 @@
...
+        # The ai max single channel rate of the NI PXIe-4331 device is 102400.0
+        assert bridge_device.ai_max_single_chan_rate == 102400.0
+
+    def test_uint_property(self, bridge_device):
+        """Test for validating the device property of uint type."""
+        assert bridge_device.product_num == 1957283012
+
+    def test_string_property(self, bridge_device):
+        """Test for validating a device property of string type."""
+        assert bridge_device.product_type == "PXIe-4331"
```
bkeryan, line 28: The `bridge_device` fixture matches against the first device that supports `UsageTypeAI.BRIDGE`: https://github.com/ni/nidaqmx-python/blob/master/tests/conftest.py#L77

If you have another bridge device in your system, these tests may use the wrong device and fail.

Some ideas:
- Write a local fixture that returns `nidaqmx.system.Device("bridgeTester")`
- Skip if the device's name isn't "bridgeTester"

tests/test_device_properties.py
```diff
@@ -0,0 +1,56 @@
...
+        assert len(ai_bridge_ranges) == 4
+        assert ai_bridge_ranges == [-0.025, 0.025, -0.1, 0.1]
+
+    def test_list_of_enums_property(self, bridge_device):
+        """Test for validating the device property of a enum array type."""
+        ai_trigger_usage = bridge_device.ai_trig_usage
+
+        assert isinstance(ai_trigger_usage, list)
+        assert len(ai_trigger_usage) == 2
+        assert ai_trigger_usage == [TriggerUsage.REFERENCE, TriggerUsage.START]
```
bkeryan, line 48: I think the length comparisons in these tests are unnecessary. If the list is the wrong length, the `==` operator should return False.

bkeryan (changes requested):

tests/max_config/nidaqmxMaxConfig.ini
subash-suresh: Noted. I have reverted the two lines

tests/test_device_properties.py
subash-suresh: updated

tests/test_device_properties.py
subash-suresh: I have included a common fixture `test_device`, which could get the device based on name.
This fixture is now used for all device tests to get the bridgeTester device based on name.
I added a common fixture, as I feel this might be required by other tests that needs to get a device based on the name.

tests/test_device_properties.py
subash-suresh: updated

tests/test_device_properties.py
bkeryan: Good, the new fixture will provide helpful errors when the specified device is missing.

bkeryan (approved):

tests/test_device_properties.py
```diff
@@ -0,0 +1,63 @@
...
+
+from nidaqmx.constants import BusType, TriggerUsage
+
+
+class TestDevicePropertyDataTypes(object):
+    """Contains a collection of pytest tests.
+
+    This validates the property getter,setter and deleter methods for different
+    data types of device properties.
+    """
```
bkeryan, line 12: Remove the test class. Test functions should be sufficient.

bkeryan (changes requested):

zhindes (approved):
Once Brad's issues are addressed, I'm happy.

tests/conftest.py
```diff
@@ -178,3 +178,20 @@ def multi_threading_test_devices():
         "nidaqmx\\tests\\max_config\\nidaqmxMaxConfig.ini to create these devices."
     )
     return None
+
+
+@pytest.fixture(scope="module")
+def test_device(request):
```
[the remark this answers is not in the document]
bkeryan: It could make sense to rename `bridge_device` to `any_bridge_device`, but it looks like only one test (another property test) uses it, so I agree that it would be okay to change that test to use this new fixture.

bkeryan, line 184: Prefixing the fixture with `test_` does not make a meaningful distinction because all test fixtures are for tests.

Suggestion: something like `named_device` or `device_by_name`

bkeryan (changes requested):

tests/test_device_properties.py
subash-suresh: removed the class

tests/test_device_properties.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @pytest.mark.parametrize("test_device", ["bridgeTester"], indirect=True)
+    def test_boolean_property(self, test_device):
```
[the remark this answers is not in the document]
subash-suresh: updated the method names to follow conventions

tests/conftest.py
subash-suresh: updated the name of the fixture to `device_by_name`

tests/legacy/test_properties.py
```diff
@@ -169,17 +169,18 @@ def test_list_of_enums_property(self, any_x_series_device):
...
     @pytest.mark.parametrize("seed", [generate_random_seed()])
-    def test_list_of_floats_property(self, bridge_device, seed):
+    @pytest.mark.parametrize("device_by_name", ["bridgeTester"], indirect=True)
+    def test_list_of_floats_property(self, device_by_name, seed):
         """Test for validating list of float property."""
-        if bridge_device is None:
+        if device_by_name is None:
             pytest.skip("requires bridge device")
 
         # Reset the pseudorandom number generator with seed.
```
bkeryan, line 178: I think this check is unnecessary because the fixture should skip the test.

bkeryan (approved):
Approved with suggestions

tests/legacy/test_properties.py
subash-suresh: I have removed the checks

tests/component/test_device_properties.py
```diff
@@ -0,0 +1,63 @@
+"""Tests for validating device properties for different basic data types."""
+import pytest
+
+from nidaqmx.constants import BusType, TriggerUsage
+
+
+@pytest.mark.parametrize("device_by_name", ["bridgeTester"], indirect=True)
+def test__device__bool_property__returns_value(device_by_name):
```
bkeryan, line 8: FYI, "given device" does make sense, because that is an object that the test is given.

subash-suresh: Do you mean that the test name could be changed to `test__given_device__bool_property__return_value`?

Issue #209, opened 2023-02-16
Support type annotations


bkeryan:
Python 3 supports optional type annotations using the [typing](https://docs.python.org/3/library/typing.html) module. 

Benefits of specifying type annotations:
- Clients can use external tools like [mypy](https://www.mypy-lang.org/) to perform static type-checking.
- Editors such as Visual Studio Code use type annotations to provide auto-complete suggestions.

nidaqmx-python already documents types in docstrings, so the information is already there. It would only need to be output in a different format.

PR #238, opened 2023-03-29
Generate base_interpreter.py file

samchris007:

- Generates Interface `BaseInterpreter` in nidaqmx-python using scrapigen metadata. 
- The generated `BaseInterpreter` has functions' signature for all functions in the scrapigen metadata.
- Reused existing logic from [function_template.py.mako](https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/function_template.py.mako) in `base_interpreter.py.mako`

Implements [Task 2330448](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330448): Generate a base interpreter class to support all the c method calls from the scrapigen metadata


generated/nidaqmx/base_interpreter.py
```diff
@@ -0,0 +1,1630 @@
...
+    GpsSignalType, LVDTSensitivityUnits, LengthUnits, Level, LineGrouping,
+    Polarity, PressureUnits, RTDType, RVDTSensitivityUnits,
+    ResistanceConfiguration, ResistanceUnits, Signal, SignalModifiers, Slope,
+    SoundPressureUnits, StrainGageBridgeType, StrainGageRosetteType,
+    StrainUnits, TEDSUnits, TemperatureUnits, TerminalConfiguration,
+    ThermocoupleType, TimeUnits, Timescale, TorqueUnits, UnitsPreScaled,
+    VelocityIEPESensorSensitivityUnits, VelocityUnits, VoltageUnits,
+    WindowTriggerCondition1, WriteBasicTEDSOptions)
+
+class BaseInterpreter:
```
zhindes, line 20: its common to explicitly inherit from `object` even though its not strictly required in Python 3

generated/nidaqmx/base_interpreter.py
```diff
@@ -0,0 +1,1630 @@
...
+    Contains signature of functions for all DAQmx APIs.
+    """
+    def add_ai_accel_4_wire_dc_voltage_chan(
+            self, physical_channel, name_to_assign_to_channel="",
+            terminal_config=TerminalConfiguration.DEFAULT, min_val=-5.0,
+            max_val=5.0, units=AccelUnits.G, sensitivity=1000.0,
+            sensitivity_units=AccelSensitivityUnits.MILLIVOLTS_PER_G,
+            voltage_excit_source=ExcitationSource.INTERNAL,
+            voltage_excit_val=0.0, use_excit_for_scaling=False,
+            custom_scale_name=""):
```
zhindes, line 31: the interpreter shouldn't have default value parameters, any functions calling them will have defaults, but they should pass them in here

zhindes (changes requested):

generated/nidaqmx/base_interpreter.py
bkeryan: FYI, I started running [pyupgrade](https://pypi.org/project/pyupgrade/) on some of our other GitHub repos and it removes explicit inheritance from `object`.

bkeryan, line 20: Would there be value in making this an abstract base class? https://peps.python.org/pep-3119/

nimi-python doesn't have a base interpreter class and uses duck typing.

I think that a base interpreter class **with type annotations** would add value compared to duck typing.

@mshafer-NI Do you have any suggestions?

generated/nidaqmx/base_interpreter.py
bkeryan, line 1: Interpreters are a private implementation detail, so add a leading underscore to the module name.

generated/nidaqmx/base_interpreter.py
```diff
@@ -0,0 +1,1630 @@
...
+        pass
+
+    def delete_saved_scale(self):
+        pass
+
+    def delete_saved_task(self):
+        pass
+
+    def device_supports_cal(self):
+        pass
```
bkeryan, line 824: Functions that operate on named objects are missing the object name:
- `delete_network_device` and `device_supports_cal` should take a device name.
- `delete_saved_global_chan` should take a global channel name.
- `delete_saved_scale` should take a scale name.
- `delete_saved_task` should take a task name. (Saved tasks are referred to by name, not handle.)

We agreed that the interpreter class will have a field for the task handle (local) or task name (gRPC), so the logic for removing taskHandle parameters is still valid here. However, the logic for removing device names, scale names, etc. does not apply to interpreter methods.

generated/nidaqmx/base_interpreter.py
bkeryan, line 31: Consider generating type annotations.

generated/nidaqmx/base_interpreter.py
```diff
@@ -0,0 +1,1630 @@
...
+
+    @staticmethod
+    def add_table_scale(
+            self, name, prescaled_vals, scaled_vals,
+            pre_scaled_units=UnitsPreScaled.VOLTS, scaled_units=None):
+        pass
+
+    @staticmethod
+    def add_task(self):
+        pass
```
bkeryan, line 529: As discussed previously, I think it would make more sense for the interpreter method names to follow the C API, like `create_task`: https://github.com/ni/nidaqmx-python/pull/210/files#r1111067299

I see this class has `clear_task` instead of `close`, but it still changes `create` to `add`.

generated/nidaqmx/base_interpreter.py
```diff
@@ -0,0 +1,1630 @@
...
+
+    def read_binary_i_32(self):
+        pass
+
+    def read_binary_u_16(self):
+        pass
+
+    def read_binary_u_32(self):
+        pass
```
bkeryan, line 1196: Snake case conversion of `Int32`, `UInt32`, `F64`, etc. resulted in extra underscores.

bkeryan (changes requested):

generated/nidaqmx/base_interpreter.py
samchris007: Sure. I have removed leading underscores in function names that has datatypes.

generated/nidaqmx/base_interpreter.py
samchris007: Got it. I have added a new function `get_all_functions()` which uses the `function_name` from the metadata directly.

generated/nidaqmx/base_interpreter.py
samchris007: Sure. I have added all the parameters from the metadata in the functions signature (Though I have filtered some parameters that doesn't have `type` in the metadata itself).

generated/nidaqmx/base_interpreter.py
samchris007: Sure. Removed.

generated/nidaqmx/base_interpreter.py
samchris007: Updated.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1621 @@
+# Do not edit this file; it was automatically generated.
+
+class BaseInterpreter(object):
+    """
+    Contains signature of functions for all DAQmx APIs.
+    """
+    def add_cdaq_sync_connection(self, port_list):
+        pass
+
+    def add_global_chans_to_task(self, task, channel_names):
```
bkeryan, line 10: Does this mean that the interpreter isn't going to store the task handle as previously discussed? That might be okay. The type of `task` would be interpreter-dependent, but it would be stored in the `Task` object, not the interpreter object. 

If `BaseInterpreter` had type annotations, it should use `task: object` not `task: Any`: https://mypy.readthedocs.io/en/latest/dynamic_typing.html#any-vs-object

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1621 @@
...
+        pass
+
+    def get_chan_attribute_double_array(
+            self, task, channel, attribute, value, size):
+        pass
+
+    def get_chan_attribute_int32(self, task, channel, attribute, value, size):
+        pass
+
+    def get_chan_attribute_string(self, task, channel, attribute, value, size):
```
bkeryan, line 732: Python strings are immutable, so the interpreter should represent string out-parameters as return values. The library interpreter should do the "double call" or "IVI dance". The gRPC interpreter should just return strings.

```
def get_chan_attribute_string(self, task, channel, attribute) -> str:
    pass
```

(It's also possible to box string out-parameters in a list or something, but that's kind of hacky. `ctypes` also boxes strings, using string buffer objects, but we don't want to use `ctypes` for the gRPC interpreter.)

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1621 @@
...
+
+    def get_chan_attribute_bool(self, task, channel, attribute, value, size):
+        pass
+
+    def get_chan_attribute_double(self, task, channel, attribute, value, size):
+        pass
+
+    def get_chan_attribute_double_array(
+            self, task, channel, attribute, value, size):
+        pass
```
bkeryan, line 727: Python lists and numpy arrays are mutable, so I don't think it's critical to represent array out-parameters as return values. However, we should remove the `size` parameter. The library interpreter should do the "double call" or "IVI dance".

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1621 @@
...
+    def get_digital_logic_family_power_up_state(
+            self, device_name, logic_family):
+        pass
+
+    def get_digital_power_up_states(self, device_name, channel_name, state):
+        pass
+
+    def get_digital_pull_up_pull_down_states(
+            self, device_name, channel_name, state):
+        pass
```
bkeryan, line 774: Power-up state functions uses varargs in the C API, so they will require a special case. Think of them as getting/setting either an array of (channel, state) tuples or parallel arrays of channels and states.

I think the way you call them in C is something like this:
```
int32 states[4];
DAQmxGetDigitalPowerUpStates("Dev1", "port0/line0", &states[0], "port0/line1", &states[1], "port0/line2", &states[2], "port0/line3", &states[3], (const char*)NULL);
DAQmxSetDigitalPowerUpStates("Dev1", "port0/line0", states[0], "port0/line1", states[1], "port0/line2", states[2], "port0/line3", states[3], (const char*)NULL);
```

bkeryan (changes requested):

generated/nidaqmx/base_interpreter.py
bkeryan: Ok, if we don't have the needed metadata, then let's not add type annotations now.

I think type annotations should be added at some point, but it doesn't have to be part of this project. https://github.com/ni/nidaqmx-python/issues/209

generated/nidaqmx/base_interpreter.py
bkeryan: If @mshafer-NI doesn't have a different suggestion, let's do this:
- Inherit from `collections.abc.ABC`
- Change the base class method bodies from `pass` to `raise NotImplementedError` (https://docs.python.org/3/library/exceptions.html#NotImplementedError)

bkeryan: I chatted with @mshafer-NI offline and he said a base class with ABC is fine as long as the shared interface (public methods) constitutes > 90% of each class.

zhindes (approved):

generated/nidaqmx/base_interpreter.py
bkeryan: I almost forgot:
- Decorate each base class method with `@abc.abstractmethod`

generated/nidaqmx/_base_interpreter.py
bkeryan, line 1: Let's release nidaqmx-python with the updated codegen before completing this PR.

src/codegen/functions/parameter.py
```diff
@@ -11,11 +11,15 @@ def __init__(self, parameter_metadata):
...
-        self._python_type_annotation = parameter_metadata["python_type_annotation"]
+        self._type = parameter_metadata.get("type", None)
+        if "ctypes_data_type" in parameter_metadata:
+            self._ctypes_data_type = parameter_metadata["ctypes_data_type"]
+        if "python_data_type" in parameter_metadata:
+            self._python_data_type = parameter_metadata["python_data_type"]
+        if "python_description" in parameter_metadata:
+            self._description = parameter_metadata["python_description"]
+        if "python_type_annotation" in parameter_metadata:
+            self._python_type_annotation = parameter_metadata["python_type_annotation"]
```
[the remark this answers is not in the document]
samchris007: Ok. Updated.

src/codegen/utilities/helpers.py
```diff
@@ -11,10 +11,13 @@
...
 
-def camel_to_snake_case(camel_case_string):
+def camel_to_snake_case(camel_case_string, custom_regexes=[]):
     """Converts a camelCase string to a snake_case string."""
+    regexes = CAMEL_TO_SNAKE_CASE_REGEXES
+    if custom_regexes:
+        regexes = custom_regexes
     partial = camel_case_string
-    for regex in CAMEL_TO_SNAKE_CASE_REGEXES:
+    for regex in regexes:
```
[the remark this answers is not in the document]
samchris007: Updated.

src/codegen/utilities/function_helpers.py
```diff
@@ -78,6 +103,18 @@ def get_parameter_signature(is_python_factory, sorted_params):
     return ", ".join(params_with_defaults)
 
 
+def get_all_parameter_signature(is_python_factory, params):
```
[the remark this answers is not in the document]
samchris007: Renamed to `get_interpreter_parameter_signature`

generated/nidaqmx/_base_interpreter.py
samchris007: Sure. Updated.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1621 @@
...
+
+    def get_exported_signal_attribute_string(
+            self, task, attribute, value, size):
+        pass
+
+    def get_exported_signal_attribute_uint32(
+            self, task, attribute, value, size):
+        pass
+
+    def get_extended_error_info(self, error_string, buffer_size):
```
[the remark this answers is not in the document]
samchris007: I have added a List `INTERPRETER_IGNORED_FUNCTIONS` which is used to ignore these types of functions when generating base interpreter.

src/codegen/utilities/function_helpers.py
```diff
@@ -11,6 +12,12 @@
     "CIGPS": "CI_GPS",
 }
 
+CUSTOM_CAMEL_TO_SNAKE_CASE_REGEXES = [
```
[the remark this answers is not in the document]
samchris007: Added comment `# This custom regex list doesn't split the string before the number.`

generated/nidaqmx/_base_interpreter.py
samchris007: Ok. But some of the functions has multiple out parameters for a single function. which include multiple arrays as return values.
- `GetAnalogPowerUpStatesWithOutputType`
- `ReadCtrTicks`

In that case, should we need to form a `Tuple(List[int], List[float], ...)` as return value for the functions?
However, I have removed the size parameter as suggested.

generated/nidaqmx/_base_interpreter.py
samchris007: Yes. Interpreter will not have functions like 'set_task_handle'.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1574 @@
+# Do not edit this file; it was automatically generated.
+from typing import List
+
+class BaseInterpreter(object):
+    """
+    Contains signature of functions for all DAQmx APIs.
+    """
+    def add_cdaq_sync_connection(self, port_list):
+        pass
```
mshafer-NI, line 9: If someone imports this class and decides to write their on `Interpreter` class (or we in future decide to add another one), they should implement all of these method, right?

To document this, rather than putting "pass", have these default implementations do a `raise NotImplemented()`, or (if using abc.ABC, mark them as [`@abstractmethod`](https://docs.python.org/3.9/library/abc.html#abc.abstractmethod)

generated/nidaqmx/_base_interpreter.py
mshafer-NI: > In that case, should we need to form a Tuple(List[int], List[float], ...) as return value for the functions?

If returning, yes. Conveniently, this still turns into
```python
    return foo, bar, # ...
```

If *mutating the passed list(s)*, the input parameter should ask for a `List`, AND the docstring should note that it will add to the parameter object.

generated/nidaqmx/base_interpreter.py
samchris007: Sure. I have inherited `abc.ABC` into `BaseInterpreter` and added decorator `@abc.abstractmethod` for all the functions.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1574 @@
+# Do not edit this file; it was automatically generated.
+from typing import List
+
+class BaseInterpreter(object):
```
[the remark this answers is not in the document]
samchris007: Sure. I have inherited `abc.ABC` into `BaseInterpreter` and added decorator `@abc.abstractmethod` for all the functions.

generated/nidaqmx/_base_interpreter.py
samchris007: Sure. But should we need to type annotate the return values in `BaseInterpreter` like this,

```
def get_analog_power_up_states_with_output_type(
            self, channel_names, array_size): -> Tuple(List[Float], List[int], ...):
    raise NotImplementedError
``

Comment (mshafer-NI) [reply generated/nidaqmx/_base_interpreter.py]: If *returning* the lists, yes. they should be documents as a `) -> Tuple(` return.

The exact value of that I would need to know more about the intended api. If it's only a list of floats, and a list of ints, then `) -> Tuple(List[float], List[int]):`.

File: generated/nidaqmx/_base_interpreter.py [hunk 17]
```diff
@@ -0,0 +1,1621 @@
...
+        pass
+
+    def get_extended_error_info(self, error_string, buffer_size):
+        pass
+
+    def get_first_samp_clk_when(self, task, data):
+        pass
+
+    def get_first_samp_timestamp_val(self, task, data):
+        pass
```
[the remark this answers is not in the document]
bkeryan: I think we should add these to `INTERPRETER_IGNORED_FUNCTIONS` and try to use the generic versions.

src/codegen/utilities/function_helpers.py
```diff
@@ -11,22 +12,62 @@
     "CIGPS": "CI_GPS",
 }
 
+FUNCTION_RETURN_TYPE_MAP_SET = {
+    "char[]": "str",
+}
+
+# This custom regex list doesn't split the string before the number.
+CUSTOM_CAMEL_TO_SNAKE_CASE_REGEXES = [
```
bkeryan, line 20: `INTERPRETER_IGNORED_FUNCTIONS` made me realize it would make sense to change this constant's prefix from `CUSTOM_` to `INTERPRETER_`.

src/codegen/templates/_base_interpreter.py.mako
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Contains signature of functions for all DAQmx APIs.
+    """
+################################################################################
+## Script function signature.
+################################################################################
+% for func in functions:
+    %if func.is_python_factory:
+    @staticmethod
+%endif
```
bkeryan, line 19: Indent the `%endif`

src/codegen/functions/parameter.py
```diff
@@ -11,11 +11,11 @@ def __init__(self, parameter_metadata):
...
-        self._type = parameter_metadata["type"]
-        self._ctypes_data_type = parameter_metadata["ctypes_data_type"]
-        self._python_data_type = parameter_metadata["python_data_type"]
-        self._description = parameter_metadata["python_description"]
-        self._python_type_annotation = parameter_metadata["python_type_annotation"]
+        self._type = parameter_metadata.get("type", None)
+        self._ctypes_data_type = parameter_metadata.get("ctypes_data_type", None)
+        self._python_data_type = parameter_metadata.get("python_data_type", None)
+        self._description = parameter_metadata.get("python_description", None)
+        self._python_type_annotation = parameter_metadata.get("python_type_annotation", None)
```
bkeryan, line 18: FYI: https://docs.python.org/3/library/stdtypes.html#dict.get

> get(key[, default])
> Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError).

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1963 @@
...
+    @abc.abstractmethod
+    def get_arm_start_trig_timestamp_val(self, task):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_arm_start_trig_trig_when(self, task):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_auto_configured_cdaq_sync_connections(self) -> str:
```
bkeryan, line 827: Ok, well, it's a start 😅

Partial type annotations are better than nothing (as long as they're correct).

generated/nidaqmx/_base_interpreter.py
bkeryan: Current status:
- This codebase has no type annotations at all.
- I [tried running mypy yesterday](https://github.com/ni/nidaqmx-python/tree/users/bkeryan/mypy/), and `check_untyped_defs` produces a lot of errors. 
- Type annotations are not part of the acceptance criteria for the gRPC feature.

So I don't think we should require type annotations. 

If we can *easily* generate type annotations, they are nice to have. However, they would have more value for public interfaces than for internal stuff like the interpreter.

Maybe I will try generating `.pyi` files at some point.

src/codegen/utilities/function_helpers.py
```diff
@@ -205,3 +258,26 @@ def instantiate_explicit_output_param(param):
...
+def get_input_params(func):
+    """Gets input parameters for the function."""
+    return (p for p in func.base_parameters if p.direction == "in")
+
+
+def get_skippable_param_for_func(func):
+    """Gets parameter name that needs to be skipped for the function."""
+    for param in func["parameters"]:
+        size = param.get("size", dict())
+        if size.get("mechanism", None) == "ivi-dance":
```
bkeryan, line 282: Simplify: `if param.get("size", {}).get("mechanism") == "ivi-dance"`

bkeryan (changes requested):

generated/nidaqmx/_base_interpreter.py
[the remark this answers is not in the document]
reckenro: Sorry for the late reply, I don't remember many aspects of DAQmx. I know Alex was more involved so he might be able to recall if there was an explicit decision made around those functions.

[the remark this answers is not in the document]
samchris007: Sure. I have removed the generation of below listed functions in **_base_interpreter.py**
- GetArmStartTrigTimestampVal
- GetFirstSampTimestampVal
- GetRefTrigTimestampVal
- GetStartTrigTimestampVal
- GetTimingAttributeExTimestamp
- GetTimingAttributeTimestamp
- GetTrigAttributeTimestamp
- SetTimingAttributeExTimestamp
- SetTimingAttributeTimestamp
- SetTrigAttributeTimestamp

src/codegen/utilities/function_helpers.py
samchris007: Updated.

src/codegen/functions/parameter.py
samchris007: Understood. Removed setting default value as `None`.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1963 @@
...
+
+    @abc.abstractmethod
+    def load_task(self, session_name, initialization_behavior):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def read_analog_f64(
+            self, task, num_samps_per_chan, timeout, fill_mode,
+            array_size_in_samps, reserved):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
samchris007: Ok. I have removed all `reserved` parameters using `include_in_proto` key.

src/codegen/templates/_base_interpreter.py.mako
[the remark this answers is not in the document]
samchris007: Sure. Removed `@staticmethod`

src/codegen/utilities/function_helpers.py
```diff
@@ -205,3 +258,26 @@ def instantiate_explicit_output_param(param):
         )
     elif param.ctypes_data_type == "ctypes.c_char_p":
         return f"{param.parameter_name} = ctypes.create_string_buffer(temp_size)"
+
+
+def get_return_value_for_func(func):
```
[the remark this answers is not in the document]
samchris007: Sure. I have moved those logic to a separate py file **interpreter_helpers.py**. and renamed accordingly.

src/codegen/utilities/function_helpers.py
samchris007: But `size` object is required within the if check. So, using it within the if check alone might make multiple get calls for `size` dict. 
However, I have optimized the code to some extent.

src/codegen/utilities/function_helpers.py
```diff
@@ -205,3 +258,26 @@ def instantiate_explicit_output_param(param):
...
+    """Gets input parameters for the function."""
+    return (p for p in func.base_parameters if p.direction == "in")
+
+
+def get_skippable_param_for_func(func):
+    """Gets parameter name that needs to be skipped for the function."""
+    for param in func["parameters"]:
+        size = param.get("size", dict())
+        if size.get("mechanism", None) == "ivi-dance":
+            return size["value"]
```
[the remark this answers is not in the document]
samchris007: Sure. Added `is_skippable_param()` and skipped `reserved` parameters.

generated/nidaqmx/_base_interpreter.py
samchris007: Sure. I have removed the function `get_return_value_for_func()` which was previously used to set return value (type annotation) as `str` for functions.
As suggested, type annotations for individual parameters along with the return value for functions will not be handled as part of this feature.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1963 @@
...
+    def get_chan_attribute_double(self, task, channel, attribute, size):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_chan_attribute_double_array(self, task, channel, attribute):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_chan_attribute_int32(self, task, channel, attribute, size):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
samchris007: I have used key `include_in_proto` along with parameter names that are to be ignored `["size", "reserved"]` to identify and remove `size` and `reserved` parameters.

generated/nidaqmx/_base_interpreter.py
samchris007: Ok sure. Are there any plans on doing the release for the updated codegen any time soon? 
If so, could you please provide some information regarding the same?

generated/nidaqmx/_base_interpreter.py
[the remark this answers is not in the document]
astarche: We had specific tasks to implement `CVIAbsoluteTime` and then add the methods that used it. I suspect that ended up undoing the removal of specific attribute accessors for those methods. I don't know of any other reason those would have been included.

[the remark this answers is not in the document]
astarche: In short: yes, I think it was a mistake.

src/codegen/utilities/function_helpers.py
bkeryan: Ok, that's fine.

bkeryan (approved):
Approved with suggestions:
- Exclude more attribute-specific functions
- Wait until Zach has made a release of 0.7.0.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1885 @@
...
+
+    @abc.abstractmethod
+    def load_task(self, session_name):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def read_analog_f64(
+            self, task, num_samps_per_chan, timeout, fill_mode,
+            array_size_in_samps):
+        raise NotImplementedError
```
bkeryan, line 1225: Compare to `_read_analog_f_64`: https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/read_functions.py

`_read_analog_f_64` takes a numpy array parameter called `read_array`:
- This allows it to use a preallocated array.
- The numpy array knows its size, so `_read_analog_f_64` doesn't take `array_size_in_samps`.

bkeryan: Here is where the preallocated array is used: `task.in_stream.readinto()`: https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/in_stream.py#L2371

generated/nidaqmx/_base_interpreter.py
samchris007: Sure. Thanks for the info.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -0,0 +1,1885 @@
...
+            self, device_name, channel_name, channel_type):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_analog_power_up_states_with_output_type(
+            self, channel_names, array_size):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_arm_start_trig_trig_when(self, task):
```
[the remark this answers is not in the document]
samchris007: Ok. Removed the generation of below mentioned functions in **_base_interpreter.py**
- GetArmStartTrigTrigWhen
- GetFirstSampClkWhen
- GetStartTrigTrigWhen
- GetSyncPulseTimeWhen
- SetArmStartTrigTrigWhen
- SetFirstSampClkWhen
- SetStartTrigTrigWhen
- SetSyncPulseTimeWhen

generated/nidaqmx/_base_interpreter.py
samchris007: Ok. I have added the array parameters for functions that has mechanism - `passed-in`.
Also, removed the size parameter `array_size_in_samps` in `read_analog_f64` using **mechanism** of the array parameter.

samchris007:
@zhindes This PR is ready to be merged.

PR #239, opened 2023-03-29
[Internal] Addition of tests for persisted task properties

subash-suresh:
- Adds Tests for testing the different datatypes of persisted task properties.
    - Added a new test file `test_persisted_task_properties.py` which contains the tests in this PR.
    - Added a persisted task `VoltageTesterTask` to the `nidaqmxMaxConfig.ini` file.
    - Added a test fixture in `test_persisted_task_properties.py`, which picks the `VoltageTesterTask` object.

Implements [Task 2330442](https://ni.visualstudio.com/DevCentral/_workitems/edit/2330442): Add or verify tests for Persisted Task property and related data types


tests/test_persisted_task_properties.py
```diff
@@ -0,0 +1,30 @@
...
+    @pytest.fixture(scope="class")
+    def voltage_tester_task(self):
+        """Gets the voltage tester task from the persisted tasks in the system."""
+        system = nidaqmx.system.System.local()
+        voltage_tester_task = system.tasks[0]
+        return voltage_tester_task
+
+    def test_boolean_property(self, voltage_tester_task):
+        """Test for validating boolean properties in persisted task."""
+        assert voltage_tester_task.allow_interactive_editing
```
subash-suresh, line 24: Will the `allow_interactive_editing ` be always true, irrespective of the instrument? 
This is similar to the question in the [persisted scale PR](https://github.com/ni/nidaqmx-python/pull/236#discussion_r1151811614)

tests/test_persisted_task_properties.py
```diff
@@ -0,0 +1,30 @@
...
+
+    This validates the property getter,setter and deleter methods for
+    persisted task properties for different data types.
+    """
+
+    @pytest.fixture(scope="class")
+    def voltage_tester_task(self):
+        """Gets the voltage tester task from the persisted tasks in the system."""
+        system = nidaqmx.system.System.local()
+        voltage_tester_task = system.tasks[0]
```
zhindes, line 19: this isn't very safe. if someone changes the imports it might break. the `PersistedTaskCollection` allows for indexing by string, so you can ask for it by name

tests/test_persisted_task_properties.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_boolean_property(self, voltage_tester_task):
+        """Test for validating boolean properties in persisted task."""
+        assert voltage_tester_task.allow_interactive_editing
+
+    def test_string_property(self, voltage_tester_task):
+        """Test for validating boolean properties in persisted task."""
+        author = voltage_tester_task.author
+
+        assert isinstance(author, str)
```
zhindes, line 30: I think this should be able to be persisted in the ini file, so we can validate it is a specific value

tests/test_persisted_task_properties.py
zhindes: I think we should be able to persist a specific value here in the ini file?

zhindes (changes requested):

tests/test_persisted_task_properties.py
subash-suresh: Added the required test data to the NI max config to persist the required information.

tests/test_persisted_task_properties.py
subash-suresh: Added the required test data to the NI max config to persist the required information.

tests/test_persisted_task_properties.py
subash-suresh: Updated to take the persisted task by name

tests/test_persisted_task_properties.py
```diff
@@ -0,0 +1,28 @@
...
+    This validates the property getter,setter and deleter methods for
+    persisted task properties for different data types.
+    """
+
+    @pytest.fixture(scope="class")
+    def voltage_tester_task(self):
+        """Gets the voltage tester task from the persisted tasks in the system."""
+        system = nidaqmx.system.System.local()
+        voltage_tester_task = system.tasks["VoltageTesterTask"]
+        return voltage_tester_task
```
bkeryan, line 20: Please add a `persisted_task` fixture similar to `persisted_scale` in https://github.com/ni/nidaqmx-python/pull/236, plus the suggested skip logic. Then we can use it in other tests that need persisted tasks, by calling `task = persisted_task.load()`.

tests/test_persisted_task_properties.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def test_boolean_property(self, voltage_tester_task):
```
bkeryan, line 22: NI's regression testing conventions: https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions

Rename to something like test__persisted_task__bool_property__returns_value

tests/test_persisted_task_properties.py
```diff
@@ -0,0 +1,28 @@
...
+import nidaqmx
+import nidaqmx.system
+
+
+class TestPersistedTaskPropertyDataTypes(object):
+    """Contains a collection of pytest tests.
+
+    This validates the property getter,setter and deleter methods for
+    persisted task properties for different data types.
+    """
```
bkeryan, line 13: Remove the test class and use test functions.

bkeryan (approved):
Approved with suggestions

tests/test_persisted_task_properties.py
subash-suresh: Added a fixture for generating persisted tasks as suggested

tests/test_persisted_task_properties.py
subash-suresh: Updated the method name and remove class names

tests/test_persisted_task_properties.py
subash-suresh: removed

tests/component/test_persisted_task_properties.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@pytest.mark.parametrize("persisted_task", ["VoltageTesterTask"], indirect=True)
+def test__persisted_task__get_bool_property__returns_value(persisted_task):
+    """Test for validating boolean properties in persisted task."""
+    assert persisted_task.allow_interactive_editing
+
+
+@pytest.mark.parametrize("persisted_task", ["VoltageTesterTask"], indirect=True)
+def test__persisted_scale__get_string_property__returns_value(persisted_task):
```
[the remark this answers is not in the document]
subash-suresh: I have updated the then clause as per suggestions

PR #240, opened 2023-03-30
[Internal] Addition of tests for scale properties 

MounikaBattu17:
- Adds Tests for testing the different datatypes of `scale` properties.
     - Added a new test file `test_scale_properties.py` which contains the tests.
     - Used scales (`double_gain_scale`, `polynomial_scale`) defined in `nidaqmxMaxConfig.ini` file and also created a linear scale to test error code scenario.
     - Datatypes covered:
         - float64(Double)
         - float64[](Double array)
         - char[] -String
         -  int32(Enum)

- Implements the [Task 2330428](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330428): Add or verify tests for Scale property and related data types


subash-suresh (approved):

bkeryan (changes requested):

zhindes (approved):
Once Brad's feedback is addressed, I'll be happy

tests/test_scale_properties.py
```diff
@@ -0,0 +1,66 @@
...
+from nidaqmx.errors import DaqError
+from nidaqmx.scale import Scale
+
+
+class TestScaleProperty(object):
+    """Contains a collection of pytest tests that validates the scale properties."""
+
+    def test_float64_property(self):
+        """Test for validating float property."""
+        scale = Scale("double_gain_scale")
```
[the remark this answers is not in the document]
MounikaBattu17: Used `persisted_scale` fixture from https://github.com/ni/nidaqmx-python/pull/236 and loaded the `scale`.

tests/test_scale_properties.py
```diff
@@ -0,0 +1,66 @@
...
+        coeff_list = [1.0, 2.0]
+        scale.poly_forward_coeff = coeff_list
+        assert scale.poly_forward_coeff == coeff_list
+
+        # Test Incorrect property usage. Reading this property value throws DaqError.
+        linear_scale = Scale.create_lin_scale("custom_linear_scale", 5)
+        try:
+            _ = linear_scale.poly_forward_coeff
+        except DaqError as e:
+            assert e.error_code == -200601
```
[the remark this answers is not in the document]
MounikaBattu17: Moved as a separate test case.

tests/test_scale_properties.py
```diff
@@ -0,0 +1,66 @@
...
+
+from nidaqmx.constants import UnitsPreScaled
+from nidaqmx.errors import DaqError
+from nidaqmx.scale import Scale
+
+
+class TestScaleProperty(object):
+    """Contains a collection of pytest tests that validates the scale properties."""
+
+    def test_float64_property(self):
```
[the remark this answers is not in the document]
MounikaBattu17: Thank you. Followed given/when/then in function name and arrange/act/assert in function body.

tests/test_scale_properties.py
```diff
@@ -1,66 +1,94 @@
...
+    """Test for validating setter for float list property."""
+    scale = persisted_scale.load()
+
+    coeff_list = [1.0, 2.0]
+    scale.poly_forward_coeff = coeff_list
+
+    assert scale.poly_forward_coeff == coeff_list
+
+
+def test__scale__get_Invalid_float64_list_property__throws_DaqError():
```
bkeryan, line 49: Naming: use lowercase snake_case.

Optional: I think it's slightly clearer if you use the given and when clauses to describe what is invalid, like "given a linear scale, when you get a poly scale property, it throws DaqError." i.e. `test__linear_scale__get_poly_scale_property__throws_daqerror`. I'm assuming we aren't trying to test the error case for each data type. However, your proposed function name is okay too (except the capitalization).

tests/test_scale_properties.py
```diff
@@ -1,66 +1,94 @@
...
-            "custom_linear_scale", 5, y_intercept=1, pre_scaled_units=UnitsPreScaled.VOLTS
-        )
+@pytest.mark.parametrize("persisted_scale", ["double_gain_scale"], indirect=True)
+def test__scale__string_property__returns_default_value(persisted_scale):
+    """Test for validating getter for string property."""
+    scale = persisted_scale.load()
 
-        # Test property default value.
-        assert scale.pre_scaled_units == UnitsPreScaled.VOLTS
+    assert scale.description == ""
```
bkeryan, line 83: Please test a more interesting property, or set the persisted `Description` to something more interesting.

bkeryan (changes requested):

tests/test_scale_properties.py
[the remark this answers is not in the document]
bkeryan: Good. I have specific comments about some of the split-up test cases, but you've got the basic idea.

tests/test_scale_properties.py
```diff
@@ -0,0 +1,94 @@
...
+from nidaqmx.scale import Scale
+
+
+@pytest.mark.parametrize("persisted_scale", ["double_gain_scale"], indirect=True)
+def test__scale__float64_property__returns_default_value(persisted_scale):
+    """Test for validating getter for float property."""
+    scale = persisted_scale.load()
+    default_value = scale.lin_slope
+
+    assert scale.lin_slope == default_value
```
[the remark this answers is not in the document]
MounikaBattu17: Thanks for the suggestion. 
1) Updated the test function names.
2) Removed the `default_value` and hardcoded expected default values in the `assert`.
3) Yes, `Scale` properties are not resettable.

tests/test_scale_properties.py
MounikaBattu17: Changed the function name to lower snake_case and specified the 'given' in name.

tests/test_scale_properties.py
MounikaBattu17: Instead of getting `description` string property, I created a custom `polynomial_scale` and read the `scaled_units` property value from the create scale.

tests/test_scale_properties.py
```diff
@@ -1,66 +1,97 @@
...
+
+    assert scale.lin_slope == 5
+
+
+@pytest.mark.parametrize("persisted_scale", ["polynomial_scale"], indirect=True)
+def test__persisted_scale__get_float64_list_property__returns_persisted_value(persisted_scale):
+    """Test for validating getter for float list property."""
+    scale = persisted_scale.load()
+
+    assert scale.poly_forward_coeff == [-1.0785280415e-14, 1.0]
```
bkeryan, line 33: I'm a bit surprised that coefficient 0 isn't exactly 0.0. If you compare to `[0.0, 1.0]`, does the test fail?

This property doesn't seem to return -1.0785280415e-14 on my test machine.
```
PS X:\nidaqmx-python> poetry run python
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
>>> import nidaqmx
>>> s = nidaqmx.Scale('polynomial_scale')
>>> s.poly_forward_coeff
[0.0, 1.0]
>>> s.poly_forward_coeff == [0.0, 1.0]
True
```
image
image

If we need to do a fuzzy comparison, I think we should use `pytest.approx`: https://docs.pytest.org/en/4.6.x/reference.html#pytest-approx
```
>>> [-1.0785280415e-14, 1.0] == [0.0, 1.0]
False
>>> [-1.0785280415e-14, 1.0] == [approx(0.0), approx(1.0)]
True
```

But double precision floating point numbers can accurately reproduce integers up to about 2^53, so I wonder why this is happening.

tests/test_scale_properties.py
bkeryan: Okay, but it's no longer a persisted scale then, so update the test name.

tests/test_scale_properties.py
```diff
@@ -0,0 +1,97 @@
...
+def test__linear_scale__get_poly_scale_property__throws_daqerror():
+    """Test for validating getter for polynomial scale float list property in linear scale."""
+    linear_scale = Scale.create_lin_scale("custom_linear_scale", 5)
+    try:
+        _ = linear_scale.poly_forward_coeff
+    except DaqError as e:
+        assert e.error_code == -200601
+
+
+def test__persisted_scale__enum_property__returns_persisted_value():
```
bkeryan, line 56: This is no longer a persisted scale.

```suggestion
def test__scale__enum_property__returns_assigned_value():
```

tests/test_scale_properties.py
```diff
@@ -0,0 +1,97 @@
...
+    scale = Scale.create_lin_scale(
+        "custom_linear_scale", 5, y_intercept=1, pre_scaled_units=UnitsPreScaled.VOLTS
+    )
+
+    scale.pre_scaled_units = UnitsPreScaled.AMPS
+
+    assert scale.pre_scaled_units == UnitsPreScaled.AMPS
+
+
+def test__persisted_scale__string_property__returns_persisted_value():
```
bkeryan, line 76: This is no longer a persisted scale.

```suggestion
def test__scale__string_property__returns_assigned_value():
```

bkeryan (changes requested):
Almost done, but we need to figure out whether the floating point rounding issue is an actual bug.

tests/test_scale_properties.py
```diff
@@ -0,0 +1,97 @@
...
+    assert scale.poly_forward_coeff == coeff_list
+
+
+def test__linear_scale__get_poly_scale_property__throws_daqerror():
+    """Test for validating getter for polynomial scale float list property in linear scale."""
+    linear_scale = Scale.create_lin_scale("custom_linear_scale", 5)
+    try:
+        _ = linear_scale.poly_forward_coeff
+    except DaqError as e:
+        assert e.error_code == -200601
```
bkeryan, line 53: As @zhindes suggested in another PR, let's use the constant: https://github.com/ni/nidaqmx-python/blob/master/src/nidaqmx/error_codes.py#L1001

MounikaBattu17: Updated the constant in place of value.

tests/test_scale_properties.py
MounikaBattu17: Updated the suggested name.

tests/test_scale_properties.py
MounikaBattu17: Updated the suggested name.

tests/test_scale_properties.py
MounikaBattu17: > I'm a bit surprised that coefficient 0 isn't exactly 0.0. If you compare to `[0.0, 1.0]`, does the test fail?
> Yes, the **test failed** when I compared with [0.0, 1.0], so I changed the comparing value to `[-1.0785280415e-14, 1.0]`

`pytest.approx` did the magic!!! Now the test is passing when comparing with `[approx(0.0), approx(1.0)]`

bkeryan: I still don't understand why this fails on your machine and not my machine. I used `poetry run tox` to run the tests with Python 3.7 through 3.11 and it didn't fail.

What Python version do you have installed?

Is it 32-bit or 64-bit?

If you delete the scales from MAX and re-import the INI file, does the test pass without `approx`?

What NI-DAQmx version do you have installed?

MounikaBattu17: > I still don't understand why this fails on your machine and not my machine. I used `poetry run tox` to run the tests with Python 3.7 through 3.11 and it didn't fail.
> I use `poetry run pytest`
> What Python version do you have installed?
> Python 3.8.10 
> Is it 32-bit or 64-bit?
> 64 bit
> If you delete the scales from MAX and re-import the INI file, does the test pass without `approx`?
> I deleted the scales from MAX and re-imported again. The tests pass with `approx `
`assert scale.poly_forward_coeff == [0.0, 1.0]`  -> **PASS**
> What NI-DAQmx version do you have installed? I have NI-DAQmx 2023 Q1.
NI-DAQmx 21.8

MounikaBattu17: When I deleted and reimported the Scales in MAX, that passed the test without `approx`. Does that mean any issue with previous import in MAX or should the re-import happen for everyone using these tests?

bkeryan: Okay, if re-importing the scale fixed it, then please remove `approx`. If it fails again, we'll have to revisit this.

bkeryan (approved):

tests/test_scale_properties.py
MounikaBattu17: Sure Brad.

tests/test_scale_properties.py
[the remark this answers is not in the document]
zhindes: I'm fine with these tests failing briefly since its easier than trying to keep track of all the balls in the air at the moment. I'll complete as-is.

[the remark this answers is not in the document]
MounikaBattu17: @zhindes There is a dislocation in the test file `test_scale_properties.py`. Currently this is in `tests\test_scale_properties.py` but the correct location is `tests\component\test_scale_properties.py`. Since this PR was dependent on another PR as Brad mentioned, I thought of doing it later. Can we please track this file location change in this PR https://github.com/ni/nidaqmx-python/pull/236 or any other solution?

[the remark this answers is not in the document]
zhindes: Doing the move later is fine. It doesn't seem relevant to #236, but once that one is in, feel free to post a new PR to move it to the right location.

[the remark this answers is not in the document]
MounikaBattu17: Sure. Sounds great! Once https://github.com/ni/nidaqmx-python/pull/236 goes in, I'll create a new PR for location change. Thank you.

tests/test_scale_properties.py
```diff
@@ -0,0 +1,98 @@
...
+@pytest.mark.parametrize("persisted_scale", ["double_gain_scale"], indirect=True)
+def test__persisted_scale__get_float64_property__returns_persisted_value(persisted_scale):
+    """Test for validating getter for float property."""
+    scale = persisted_scale.load()
+
+    assert scale.lin_slope == 2.0
+
+
+@pytest.mark.parametrize("persisted_scale", ["double_gain_scale"], indirect=True)
+def test__persisted_scale__set_float64_property__returns_assigned_value(persisted_scale):
```
bkeryan, line 20: Oops, setting properties on persisted scales breaks other tests that use this scale.

DAQmx caches persisted scales in each process. `persisted_scale.load()` doesn't invalidate the cache because persisted scales only have a load operation in Python. In the C API, there is no load function for scales.

I'll post a follow-up PR to fix this.

bkeryan: https://github.com/ni/nidaqmx-python/pull/258

MounikaBattu17: Thank you Brad!

PR #241, opened 2023-03-30
Addition of tests for channel properties

vigkre:
- Adds Tests for testing the different datatypes of channel properties.
    - Added a new test file `test_channel_properties.py` which contains the tests
    -  Datatypes covered:
         - `bool`
         - `int32`
         - `float64`(Double)
         - `char[]`(String)
         - `uint32`
- The tests for int32 properties were covered by the tests written for enum properties. There was no property the just returned direct int values, all of them represented enums.
- Verified that `test_properties.py` already contains test for `float64[]`(Double array) --  https://github.com/ni/nidaqmx-python/blob/master/tests/legacy/test_properties.py#L172

- Implements [Task 2330424](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330424): Add or verify tests for Channel property and related data types

What testing has been made?


tests/test_channel_properties.py
```diff
@@ -0,0 +1,197 @@
...
+)
+from nidaqmx.tests.helpers import generate_random_seed
+
+
+class TestChannelPropertyDataTypes(object):
+    """Contains a collection of pytest tests.
+
+    This validates the property getter,setter and deleter methods for different
+    data types of channel properties.
+    """
```
bkeryan, line 29: Remove the test class and use test functions.

tests/test_channel_properties.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @pytest.mark.parametrize("seed", [generate_random_seed()])
+    def test_boolean_property(self, any_x_series_device, seed):
```
bkeryan, line 32: NI's regression testing conventions: https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions

Rename using given/when/then syntax. This may require splitting the test cases and adding local fixtures to share init code.

bkeryan (changes requested):

tests/test_channel_properties.py
```diff
@@ -0,0 +1,197 @@
...
+                terminal_config=TerminalConfiguration.NRSE,
+                min_val=-10.0,
+                max_val=10.0,
+                units=VoltageUnits.VOLTS,
+                bridge_config=BridgeConfiguration.NO_BRIDGE,
+                voltage_excit_source=ExcitationSource.EXTERNAL,
+                voltage_excit_val=0.1,
+                use_excit_for_scaling=False,
+                custom_scale_name="",
+            )
```
bkeryan, line 52: Consider using default values for more add_*_channel parameters.

zhindes: Yeah, if you aren't testing the resulting attribute value, then leveraging the defaults is great

tests/test_channel_properties.py
```diff
@@ -0,0 +1,197 @@
...
+    @pytest.mark.parametrize("seed", [generate_random_seed()])
+    def test_enum_property(self, sim_ts_power_device, seed):
+        """Test for validating channel property of enum type."""
+        # Reset the pseudorandom number generator with seed.
+        random.seed(seed)
+
+        pwr_phys_chan = f"{sim_ts_power_device.name}/power"
+        voltage_setpoint = random.random() * 6.0
+        current_setpoint = random.random() * 3.0
+        output_enable = random.choice([True, False])
```
zhindes, line 74: these random values aren't valuable, just use a constant or even leverage the function default, if it exists

tests/test_channel_properties.py
```diff
@@ -0,0 +1,197 @@
...
+        ai_phys_chan = random.choice(any_x_series_device.ai_physical_chans).name
+
+        with nidaqmx.Task() as task:
+            ai_channel = task.ai_channels.add_ai_rtd_chan(
+                ai_phys_chan,
+                name_to_assign_to_channel="RTDChannel",
+                min_val=28.0,
+                max_val=120.0,
+                units=TemperatureUnits.K,
+                rtd_type=RTDType.PT_3750,
```
zhindes, line 117: I think the `rtd_type` is the only thing relevant to the `ai_rtd_a` value. defaults for everything else is fine.

tests/test_channel_properties.py
```diff
@@ -0,0 +1,197 @@
...
+            value_to_test = random.random()
+            ai_channel.ai_rtd_a = value_to_test
+            assert ai_channel.ai_rtd_a == value_to_test
+
+            # Test property deleter.
+            del ai_channel.ai_rtd_a
+            assert ai_channel.ai_rtd_a == 0.00381
+
+    @pytest.mark.parametrize("seed", [generate_random_seed()])
+    def test_string_property(self, any_x_series_device, seed):
```
zhindes, line 137: you're not using the seed in this function, so I would just remove it. Same is true for most of these

zhindes (changes requested):

tests/test_channel_properties.py
vigkre: Removed unnecessary usage of seed in the tests

tests/test_channel_properties.py
vigkre: Added value assignment to `rtd_type` only and removed other assignments

tests/test_channel_properties.py
vigkre: Replaced random values with default.

tests/test_channel_properties.py
vigkre: Removed unnecessary value assignments for parameters.

tests/test_channel_properties.py
vigkre: Renamed the tests with the conventional syntax. Added fixtures to share the init code that is useful for tests

tests/test_channel_properties.py
vigkre: Removed test class

tests/component/test_channel_properties.py
```diff
@@ -0,0 +1,116 @@
...
+    """Test to validate setter for channel property of enum type."""
+    ai_power_chan.pwr_idle_output_behavior = PowerIdleOutputBehavior.OUTPUT_DISABLED
+    assert ai_power_chan.pwr_idle_output_behavior == PowerIdleOutputBehavior.OUTPUT_DISABLED
+
+
+def test__channel__reset_enum_property__returns_initial_value(ai_power_chan: AIChannel):
+    """Test to validate resetting channel property of enum type."""
+    # Default value is to maintain the existing value after the task is uncommitted
+    # in NI TS-15200 device
+    del ai_power_chan.pwr_idle_output_behavior
```
zhindes, line 46: set to disabled before reset

tests/component/test_channel_properties.py
```diff
@@ -0,0 +1,116 @@
...
+    value_to_set = 0.058388
+    ai_rtd_chan.ai_rtd_a = value_to_set
+    assert ai_rtd_chan.ai_rtd_a == value_to_set
+
+
+def test__channel__reset_float_property__returns_initial_value(ai_rtd_chan: AIChannel):
+    """Test to validate resetting channel property of float type."""
+    # Default 'A' constant of the Callendar-Van Dusen equation
+    # in the NI PCIe-6363 device is "0.00381"
+    del ai_rtd_chan.ai_rtd_a
```
zhindes, line 68: set to some other value before reset

zhindes (approved):
only trivial changes, I think. I really like the new fixtures, great idea :)

tests/conftest.py
```diff
@@ -202,3 +202,61 @@ def persisted_task(request):
...
     return pathlib.Path(__file__).parent / "test_assets"
+
+
+@pytest.fixture(scope="module")
+def ai_voltage_chan_with_excit(any_x_series_device):
+    """Creates AI Channel object to measure Voltage.
+
+    Used in testing channel property of uint32 and boolean data types.
+    """
+    task = nidaqmx.Task()
```
zhindes, line 213: Wait, we're leaking these tasks. It looks like using a [yield fixture](https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization) is the right way to clean this up. You can put this in a `with` block and yield the `ai_channel`

zhindes (changes requested):
I realized one major issue with these fixtures, but I think there is a simple fix.

tests/component/test_channel_properties.py
```diff
@@ -0,0 +1,116 @@
...
+    """Test to validate setter for channel property of boolean type."""
+    ai_voltage_chan_with_excit.ai_excit_use_for_scaling = True
+    assert ai_voltage_chan_with_excit.ai_excit_use_for_scaling
+
+
+def test__channel__reset_boolean_property__returns_initial_value(
+    ai_voltage_chan_with_excit: AIChannel,
+):
+    """Test to validate resetting channel property of boolean type."""
+    del ai_voltage_chan_with_excit.ai_excit_use_for_scaling
```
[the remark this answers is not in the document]
vigkre: Set non-default value before resetting.

tests/component/test_channel_properties.py
vigkre: Set non-default value before resetting

tests/conftest.py
vigkre: Updated the fixture accordingly. Thanks for the suggestion Zach :)

zhindes (approved):

tests/component/test_channel_properties.py
```diff
@@ -0,0 +1,136 @@
...
+def test__channel__set_boolean_property__returns_assigned_value(
+    ai_voltage_chan_with_excit: AIChannel,
+):
+    """Test to validate setter for channel property of boolean type."""
+    ai_voltage_chan_with_excit.ai_excit_use_for_scaling = True
+
+    assert ai_voltage_chan_with_excit.ai_excit_use_for_scaling
+
+
+def test__channel__reset_boolean_property__returns_initial_value(
```
bkeryan, line 22: Change all references to "initial_value" to "default_value".

Reason: "default value" is the standard DAQmx terminology, and "initial value" could refer to a value assigned by the Create Channel function.

tests/conftest.py
```diff
@@ -202,3 +202,61 @@ def persisted_task(request):
 def test_assets_directory() -> pathlib.Path:
     """Gets path to test_assets directory."""
     return pathlib.Path(__file__).parent / "test_assets"
+
+
+@pytest.fixture(scope="module")
+def ai_voltage_chan_with_excit(any_x_series_device):
```
bkeryan, line 208: FYI, if we don't expect to reuse a fixture in other tests, we don't have to define it in conftest.py. We can define it in test_channel_properties.py.

bkeryan (approved):
Approved with suggestions.

tests/component/test_channel_properties.py
```diff
@@ -1,11 +1,74 @@
...
 from nidaqmx._task_modules.channels.ai_channel import AIChannel
 from nidaqmx._task_modules.channels.ci_channel import CIChannel
-from nidaqmx.constants import PowerIdleOutputBehavior
-
-
-def test__channel__get_boolean_property__returns_value(ai_voltage_chan_with_excit: AIChannel):
+from nidaqmx.constants import ExcitationSource, PowerIdleOutputBehavior, RTDType
+
+
+@pytest.fixture(scope="module")
```
bkeryan, line 11: Scope=module reuses the same task for multiple test cases. This may make test failures harder to debug. For example, if you added an error test that sets an invalid value and verifies the task, then other test cases would raise an error when they verify the task.

vigkre: [Scope function added to fixture to destroy task at end of test](https://github.com/ni/nidaqmx-python/pull/241/commits/672297e0087f313b262de4a85ea76663e8fd3921)

bkeryan: FYI, `scope='function'` is the default.

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.fixture

bkeryan (approved):

PR #242, opened 2023-03-30
Addition of tests for properties in physical channel

Sakthi-SM:
* Adds tests for various types of properties present in `physical_channel.py`.
* The tests added covers the following properties in physical channel: 
`uint32`,`uint32[]`,`str`,`bytes`,`int32[]`

This implements [Task 2330441](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330441): Add or verify tests for Physical Channel property and related data types


tests/test_physical_channel_property.py
```diff
@@ -0,0 +1,65 @@
...
+from nidaqmx.tests.test_read_write import TestDAQmxIOBase
+
+
+class TestPhysicalChannelProperty(TestDAQmxIOBase):
+    """Contains a collection of pytest tests that validates the physical channel properties."""
+
+    def test_bool_property(self, any_x_series_device):
+        """Test for validating boolean attributes in physical channel."""
+        phys_chans = any_x_series_device.di_lines
+        assert phys_chans[0].di_change_detect_supported
```
Sakthi-SM, line 14: @bkeryan  does this value remain constant for all the device ? If not, can we set it from `ini` file?

bkeryan: I think all X Series devices support change detection on all DIO lines, so this is fine.

tests/test_physical_channel_property.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_byte_property(self, any_x_series_device):
+        """Test for validating byte attributes in physical channel."""
+        phys_chans = any_x_series_device.ai_physical_chans
+
+        # Generate path to a virtual TEDS file.
+        teds_file_path = os.path.join(os.path.dirname(__file__), "teds", "Voltage.ted")
+        phys_chans["ai0"].configure_teds(teds_file_path)
+
+        phys_chans["ai0"].teds_bit_stream
```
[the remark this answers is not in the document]
bkeryan: Try reading the file and comparing the bytes. If it doesn't match, then it might require additional parsing or conversions.

tests/test_physical_channel_property.py
```diff
@@ -0,0 +1,65 @@
+"Contains a collection of pytest tests that validates the physical channel properties."
+import os
+
+from nidaqmx.constants import TerminalConfiguration
+from nidaqmx.tests.test_read_write import TestDAQmxIOBase
+
+
+class TestPhysicalChannelProperty(TestDAQmxIOBase):
+    """Contains a collection of pytest tests that validates the physical channel properties."""
```
bkeryan, line 9: Remove the test class and use test functions. I don't think this test is using any of the base class's methods.

tests/test_physical_channel_property.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_bool_property(self, any_x_series_device):
```
bkeryan, line 11: NI's regression testing conventions: https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions

Rename using given/when/then syntax, like `test__phys_chan__bool_property__returns_value`

bkeryan (changes requested):

zhindes (approved):
Once Brad's feedback is addressed, I'll be happy

tests/test_physical_channel_property.py
```diff
@@ -0,0 +1,65 @@
...
+        """Test for validating boolean attributes in physical channel."""
+        phys_chans = any_x_series_device.di_lines
+        assert phys_chans[0].di_change_detect_supported
+
+    def test_byte_property(self, any_x_series_device):
+        """Test for validating byte attributes in physical channel."""
+        phys_chans = any_x_series_device.ai_physical_chans
+
+        # Generate path to a virtual TEDS file.
+        teds_file_path = os.path.join(os.path.dirname(__file__), "teds", "Voltage.ted")
```
bkeryan, line 21: FYI, I'm moving this to `test/test_assets/teds/Voltage.ted` and adding a `test_assets_directory` fixture. 

https://github.com/ni/nidaqmx-python/pull/245

tests/test_physical_channel_property.py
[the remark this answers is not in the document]
Sakthi-SM: I have hardcoded the expected value in the test files.

tests/test_physical_channel_property.py
```diff
@@ -1,65 +1,107 @@
...
-        phys_chans["ai0"].configure_teds(teds_file_path)
-
-        assert phys_chans["ai0"].teds_mfg_id == 17
+def test__physical_channel__get_bool_property__returns_value(any_x_series_device):
+    """Test for validating boolean attributes in physical channel."""
+    phys_chans = any_x_series_device.di_lines
+    assert phys_chans[0].di_change_detect_supported
+
+
+def test__physical_channel__get_bit_stream__returns_assigned_value(
```
bkeryan, line 13: If the focus of the test is on the property, then call `configure_teds` in the arrange section, not the act section, and change the "given" clause to "physical_channel_with_teds".

Also, "configured_value" is more appropriate than "assigned_value" for the result of "configure_teds".

This applies to every test case that calls `configure_teds`.

tests/test_physical_channel_property.py
```diff
@@ -1,65 +1,107 @@
...
+    any_x_series_device, test_assets_directory
+):
+    """Test for validating byte attributes in physical channel."""
+    phys_chans = any_x_series_device.ai_physical_chans
+    expected_value = numpy.array(VALUES_IN_TED, dtype=numpy.uint8)
+
+    # Generate path to a virtual TEDS file.
+    teds_file_path = str(test_assets_directory / "teds" / "Voltage.ted")
```
bkeryan, line 21: Replace this with a local fixture (in this file, not in conftest.py) called `teds_file_path`

tests/test_physical_channel_property.py
```diff
@@ -0,0 +1,110 @@
...
+
+def test__physical_channel_with_teds__get_bit_stream__returns_configured_value(
+    any_x_series_device, teds_file_path
+):
+    """Test for validating byte attributes in physical channel."""
+    phys_chans = any_x_series_device.ai_physical_chans
+    expected_value = numpy.array(VALUES_IN_TED, dtype=numpy.uint8)
+
+    # Generate path to a virtual TEDS file.
+    phys_chans["ai0"].configure_teds(teds_file_path)
```
bkeryan, line 22: TEDS is part of the given clause ("physical_channel_with_teds"), so `configure_teds` should be part of the arrange section. Remove the blank line between `expected_value = ...` and `configure_teds`.

Same for all of these TEDS test cases.

tests/test_physical_channel_property.py
```diff
@@ -0,0 +1,110 @@
+"Contains a collection of pytest tests that validates the physical channel properties."
+import numpy
+import pytest
+
+from nidaqmx.constants import TerminalConfiguration
+
+
+def test__physical_channel__get_bool_property__returns_value(any_x_series_device):
```
bkeryan, line 8: FYI, "given physical channel" does make sense, because that is an object that the test is given.

bkeryan (approved):

PR #250, opened 2023-04-03
Addition of tests for trigger properties

vigkre:
- Adds Tests for testing the different datatypes of channel properties.
    - Added a new test file `test_trigger_properties.py` which contains the tests
    -  Datatypes covered:
         - `int32`
         - `float64`(Double)
         - `char[]`(String)
         - `uint32`
- Verified that `test_properties.py` already contains test for `bool` -- https://github.com/ni/nidaqmx-python/blob/master/tests/legacy/test_properties.py#L20

- Implements [Task 2330427](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330427): Add or verify tests for Trigger property and related data types

What testing has been made?


tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,136 @@
...
+import nidaqmx
+from nidaqmx.constants import TriggerType
+from nidaqmx.error_codes import DAQmxErrors
+from nidaqmx.errors import DaqError
+
+
+def test__trigger__get_float_property__returns_value(any_x_series_device):
+    """Test to validate getter for trigger property of float type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
```
zhindes, line 12: #241 added a fixture for channels, which I like. There is one issue with their impl that I believe is resolvable, but I think its a good idea to do the same for these tests to remove a bunch of duplicate code!

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,136 @@
...
+        assert task.triggers.start_trigger.dig_edge_dig_fltr_timebase_rate == value_to_test
+
+
+def test__trigger__reset_float_property__returns_initial_value(any_x_series_device):
+    """Test to validate resetting trigger property of float type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        # Default Rate of the pulse width filter timebase in NI PCIe-6363 device is 0.0
+        del task.triggers.start_trigger.dig_edge_dig_fltr_timebase_rate
```
zhindes, line 34: this test currently could be passing if reset is a noop - set to non-default (you don't need to validate that worked since another test does that) before reset

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,136 @@
...
+        task.triggers.start_trigger.dig_edge_dig_fltr_timebase_src = value_to_test
+        assert task.triggers.start_trigger.dig_edge_dig_fltr_timebase_src == value_to_test
+
+
+def test__trigger__reset_string_property__returns_initial_value(any_x_series_device):
+    """Test to validate resetting trigger property of string type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        del task.triggers.start_trigger.dig_edge_dig_fltr_timebase_src
```
zhindes, line 62: set to non-default before reset

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,136 @@
...
+            e.error_type = DAQmxErrors.TRIG_WHEN_ON_DEMAND_SAMP_TIMING
+
+
+def test__trigger__reset_enum_property__returns_initial_value(any_x_series_device):
+    """Test to validate resetting trigger property of enum type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        # The default trig_type value in NI PCIe-6363 device is "NONE"
+        del task.triggers.start_trigger.trig_type
```
zhindes, line 105: set to non-default before reset

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,136 @@
...
+        assert task.triggers.reference_trigger.pretrig_samples == value_to_test
+
+
+def test__trigger__reset_uint32_property__returns_initial_value(any_x_series_device):
+    """Test to validate resetting trigger property of uint32 type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        # The default pretrig_samples value in NI PCIe-6363 device is 2
+        del task.triggers.reference_trigger.pretrig_samples
```
zhindes, line 135: set to non-default before reset

zhindes (changes requested):

tests/component/test_trigger_properties.py
vigkre: Added fixture for this block of code.

tests/component/test_trigger_properties.py
vigkre: Resolved. Set non-default value before resetting.

zhindes (approved):

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,115 @@
...
+    assert ai_voltage_chan_task.triggers.start_trigger.trig_type == value_to_test
+
+
+def test__trigger__set_trig_type_without_cfg_samp_clk__throws_daqerror(ai_voltage_chan_task: Task):
+    """Test to validate error while setting trigger type without configuring sample clock."""
+    try:
+        value_to_test = TriggerType.ANALOG_EDGE
+        ai_voltage_chan_task.triggers.start_trigger.trig_type = value_to_test
+    except DaqError as e:
+        e.error_type = DAQmxErrors.TRIG_WHEN_ON_DEMAND_SAMP_TIMING
```
bkeryan, line 83: This passes if no exception is thrown. Use [pytest.raises](https://docs.pytest.org/en/4.6.x/reference.html#pytest-raises). See https://github.com/ni/nidaqmx-python/pull/259 for an example.

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,115 @@
...
+    """Test to validate setter for trigger property of float type."""
+    value_to_test = 2.505
+    ai_voltage_chan_task.triggers.start_trigger.dig_edge_dig_fltr_timebase_rate = value_to_test
+
+    assert (
+        ai_voltage_chan_task.triggers.start_trigger.dig_edge_dig_fltr_timebase_rate == value_to_test
+    )
+
+
+def test__trigger__reset_float_property__returns_initial_value(ai_voltage_chan_task: Task):
```
bkeryan, line 25: Change all references to "initial_value" to "default_value".

Reason: "default value" is the standard DAQmx terminology, and "initial value" could refer to a value assigned by the Create Channel function.

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,115 @@
+"""Contains a collection of pytest tests that validates trigger properties."""
+
+from nidaqmx.constants import TriggerType
+from nidaqmx.error_codes import DAQmxErrors
+from nidaqmx.errors import DaqError
+from nidaqmx.task import Task
+
+
+def test__trigger__get_float_property__returns_value(ai_voltage_chan_task: Task):
```
bkeryan, line 9: Rename to ..._returns_default_value. Same for the other tests that validate default values.

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,115 @@
...
+    assert (
+        ai_voltage_chan_task.triggers.start_trigger.dig_edge_dig_fltr_timebase_src == value_to_test
+    )
+
+
+def test__trigger__reset_string_property__returns_initial_value(ai_voltage_chan_task: Task):
+    """Test to validate resetting trigger property of string type."""
+    ai_voltage_chan_task.triggers.start_trigger.dig_edge_dig_fltr_timebase_src = (
+        "Test Value for Digital Edge Digital Filter Timebase Source"
+    )
```
bkeryan, line 54: Use a more realistic value like "PFI3"

tests/conftest.py
```diff
@@ -202,3 +202,11 @@ def persisted_task(request):
 def test_assets_directory() -> pathlib.Path:
     """Gets path to test_assets directory."""
     return pathlib.Path(__file__).parent / "test_assets"
+
+
+@pytest.fixture(scope="module")
+def ai_voltage_chan_task(any_x_series_device):
```
bkeryan, line 208: All tasks have channels, so `ai_voltage_task` is just as descriptive.

Not all AI voltage tasks are on X Series devices. Consider moving your fixture from conftest.py to test_trigger_properties.py so that it can make assumptions about the device model.

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,115 @@
...
+
+from nidaqmx.constants import TriggerType
+from nidaqmx.error_codes import DAQmxErrors
+from nidaqmx.errors import DaqError
+from nidaqmx.task import Task
+
+
+def test__trigger__get_float_property__returns_value(ai_voltage_chan_task: Task):
+    """Test to validate getter for trigger property of float type."""
+    # Default Rate of the pulse width filter timebase in NI PCIe-6363 device is 0.0
```
bkeryan, line 11: If the test name says this is the default value, is this comment really needed? Same for the other tests that validate default values.

If you ask for an `ai_voltage_chan`, you shouldn't assume it comes from a PCIe-6363. 

If you make `ai_voltage_chan` a local fixture, it's okay to make more assumptions about what device it uses. However, `any_x_series_device` isn't guaranteed to give you a PCIe-6363.

Anyway, remove this comment, change the test name, and consider using a local fixture.

bkeryan (changes requested):

tests/conftest.py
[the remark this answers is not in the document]
vigkre: Moved fixture to local for tests

tests/component/test_trigger_properties.py
vigkre: Updated

tests/component/test_trigger_properties.py
vigkre: Renamed to _returns_default_value

tests/component/test_trigger_properties.py
vigkre: Renamed to _returns_default_value

tests/component/test_trigger_properties.py
vigkre: Removed the comment

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,118 @@
...
+
+@pytest.fixture(scope="module")
+def ai_voltage_task(any_x_series_device):
+    """Gets AI voltage task."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+        yield task
+
+
+def test__trigger__get_float_property__returns_default_value(ai_voltage_task: Task):
```
bkeryan, line 20: I don't think "given trigger" makes sense, because the test is not given a trigger. Rename all test cases to `test__ai_task__...`

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,118 @@
...
+
+def test__trigger__set_trig_type_without_cfg_samp_clk__throws_daqerror(ai_voltage_task: Task):
+    """Test to validate error while setting trigger type without configuring sample clock."""
+    value_to_test = TriggerType.ANALOG_EDGE
+    ai_voltage_task.triggers.start_trigger.trig_type = value_to_test
+
+    with pytest.raises(DaqError) as exc_info:
+        _ = ai_voltage_task.triggers.start_trigger.trig_type
+
+    assert exc_info.value.error_type == DAQmxErrors.TRIG_WHEN_ON_DEMAND_SAMP_TIMING
```
bkeryan, line 77: The test name has "set_trig_type" in the "when" clause, but the test actually sets it in the "arrange" section and gets it in the "act" section.

Suggestion:
- Move the set call to the "act" section. 
- The error is really caused by the set call, but it isn't reported until you verify the task. Instead of using a get call to verify the task, use `task.control(TaskMode.VERIFY)`
- Move "without_cfg_samp_clk" to the "given" clause: `test__ai_task_without_cfg_samp_clk__set_trig_type__throws_daqerror`

bkeryan (changes requested):

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,118 @@
...
+import pytest
+
+import nidaqmx
+from nidaqmx.constants import TriggerType
+from nidaqmx.error_codes import DAQmxErrors
+from nidaqmx.errors import DaqError
+from nidaqmx.task import Task
+
+
+@pytest.fixture(scope="module")
```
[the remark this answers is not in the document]
vigkre: [Scope function added to fixture to destroy task at end of test](https://github.com/ni/nidaqmx-python/pull/250/commits/e2104c12472c79f532444f69dd92c2f7646bfbd3)

tests/component/test_trigger_properties.py
vigkre: Renamed all the test cases to `test__ai_task__...`

tests/component/test_trigger_properties.py
vigkre: Used  `ai_voltage_task.control(TaskMode.TASK_VERIFY)` instead of set call.

Updated the test name to `test__ai_task_without_cfg_samp_clk__set_trig_type__throws_daqerror`

tests/component/test_trigger_properties.py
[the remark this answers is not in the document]
bkeryan: FYI, scope=function is the default

tests/component/test_trigger_properties.py
bkeryan: "Move the set call to the "act" section" means put it in the middle group of lines.

```
def test__trigger__set_trig_type_without_cfg_samp_clk__throws_daqerror(ai_voltage_task: Task):
    """Test to validate error while setting trigger type without configuring sample clock."""
    value_to_test = TriggerType.ANALOG_EDGE

    ai_voltage_task.triggers.start_trigger.trig_type = value_to_test
    with pytest.raises(DaqError) as exc_info:
        _ = ai_voltage_task.triggers.start_trigger.trig_type

    assert exc_info.value.error_type == DAQmxErrors.TRIG_WHEN_ON_DEMAND_SAMP_TIMING
```

bkeryan (approved):

tests/component/test_trigger_properties.py
```diff
@@ -0,0 +1,118 @@
...
+    assert ai_voltage_task.triggers.start_trigger.trig_type == TriggerType.NONE
+
+
+def test__ai_task_without_cfg_samp_clk__set_trig_type__throws_daqerror(ai_voltage_task: Task):
+    """Test to validate error while setting trigger type without configuring sample clock."""
+    value_to_test = TriggerType.ANALOG_EDGE
+    ai_voltage_task.triggers.start_trigger.trig_type = value_to_test
+
+    with pytest.raises(DaqError) as exc_info:
+        _ = ai_voltage_task.control(TaskMode.TASK_VERIFY)
```
bkeryan, line 75: control() doesn't return anything interesting.

tests/component/test_trigger_properties.py
vigkre: Updated as suggested

tests/component/test_trigger_properties.py
vigkre: Resolved

vigkre:
@zhindes Please merge this PR as we got required approvals.

PR #251, opened 2023-04-03
Addition of tests for export signal properties

vigkre:
- Adds Tests for testing the different datatypes of export signal properties.
    - Added a new test file `test_export_signal_properties.py` which contains the tests
    -  Datatypes covered:
         - `int32`
         - `char[]`(String)

- Implements[Task 2330440](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330440): Add or verify tests for Export Signal property and related data types

What testing has been made?

### DIFF
--- a/tests/component/test_export_signal_properties.py
+++ b/tests/component/test_export_signal_properties.py
@@ -1,0 +1,81 @@
+"""Contains a collection of pytest tests that validates export signal properties."""
+
+import pytest
+
+import nidaqmx
+from nidaqmx.constants import ExportAction
+from nidaqmx.error_codes import DAQmxErrors
+from nidaqmx.errors import DaqError
+from nidaqmx.system.device import Device
+
+
+def test__export_signal__get_enum_property__returns_value(any_x_series_device: Device):
+    """Test to validate getter for export signal property of enum type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        assert task.export_signals.samp_clk_output_behavior == ExportAction.PULSE
+
+
+def test__export_signal__set_enum_property__returns_assigned_value(any_x_series_device: Device):
+    """Test to validate setter for export signal property of enum type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        value_to_set = ExportAction.LEVEL
+        task.export_signals.samp_clk_output_behavior = value_to_set
+
+        assert task.export_signals.samp_clk_output_behavior == value_to_set
+
+
+def test__export_signal__reset_enum_property__returns_initial_value(any_x_series_device: Device):
+    """Test to validate resetting export signal property of enum type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        del task.export_signals.samp_clk_output_behavior
+
+        assert task.export_signals.samp_clk_output_behavior == ExportAction.PULSE
+
+
+def test__export_signal__get_string_property__returns_value(any_x_series_device: Device):
+    """Test to validate getter for export signal property of string type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        assert task.export_signals.start_trig_output_term == ""
+
+
+def test__export_signal__set_invalid_routing_destination__throws_daqerror(
+    any_x_series_device: Device,
+):
+    """Test to validate setter for export signal property of string type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        task.export_signals.start_trig_output_term = "RTSI"
+
+        with pytest.raises(DaqError) as e:
+            _ = task.export_signals.start_trig_output_term
+        assert e.value.error_type == DAQmxErrors.INVALID_ROUTING_DESTINATION_TERMINAL_NAME_ROUTING
+
+
+def test__export_signal__set_string_property__returns_assigned_value(any_x_series_device: Device):
+    """Test to validate setter for export signal property of string type."""
+    with nidaqmx.Task() as task:
+        task.ao_channels.add_ao_voltage_chan(any_x_series_device.ao_physical_chans[0].name)
+
+        value_to_set = "RSE"
+        task.export_signals.start_trig_output_term = value_to_set
+
+        assert task.export_signals.start_trig_output_term == value_to_set
+
+
+def test__export_signal__reset_string_property__returns_initial_value(any_x_series_device: Device):
+    """Test to validate resetting export signal property of string type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        del task.export_signals.start_trig_output_term
+
+        assert task.export_signals.start_trig_output_term == ""
--- a/tests/component/test_export_signal_properties.py
+++ b/tests/component/test_export_signal_properties.py
@@ -9,7 +9,7 @@ from nidaqmx.errors import DaqError
 from nidaqmx.task import Task
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="function")
 def ai_voltage_task(any_x_series_device):
     """Gets AI voltage task."""
     with nidaqmx.Task() as task:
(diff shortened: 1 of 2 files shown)


tests/component/test_export_signal_properties.py
```diff
@@ -0,0 +1,81 @@
...
+import nidaqmx
+from nidaqmx.constants import ExportAction
+from nidaqmx.error_codes import DAQmxErrors
+from nidaqmx.errors import DaqError
+from nidaqmx.system.device import Device
+
+
+def test__export_signal__get_enum_property__returns_value(any_x_series_device: Device):
+    """Test to validate getter for export signal property of enum type."""
+    with nidaqmx.Task() as task:
```
zhindes, line 14: #241 added a fixture for channels, which I like. There is one issue with their impl that I believe is resolvable, but I think its a good idea to do the same for these tests to remove a bunch of duplicate code!

tests/component/test_export_signal_properties.py
```diff
@@ -0,0 +1,81 @@
...
+
+        assert task.export_signals.samp_clk_output_behavior == value_to_set
+
+
+def test__export_signal__reset_enum_property__returns_initial_value(any_x_series_device: Device):
+    """Test to validate resetting export signal property of enum type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        del task.export_signals.samp_clk_output_behavior
```
zhindes, line 36: this test currently could be passing if reset is a noop - set to non-default (you don't need to validate that worked since another test does that) before reset

tests/component/test_export_signal_properties.py
```diff
@@ -0,0 +1,81 @@
...
+
+        assert task.export_signals.start_trig_output_term == value_to_set
+
+
+def test__export_signal__reset_string_property__returns_initial_value(any_x_series_device: Device):
+    """Test to validate resetting export signal property of string type."""
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+
+        del task.export_signals.start_trig_output_term
```
zhindes, line 79: set to non-default before reset

zhindes (changes requested):

tests/component/test_export_signal_properties.py
vigkre: Added fixtures for this block of code.

tests/component/test_export_signal_properties.py
vigkre: Resolved. Set non-default value before resetting.

zhindes (approved):

tests/component/test_export_signal_properties.py
```diff
@@ -0,0 +1,63 @@
...
+    assert ai_voltage_chan_task.export_signals.start_trig_output_term == ""
+
+
+def test__export_signal__set_invalid_routing_destination__throws_daqerror(
+    ai_voltage_chan_task: Task,
+):
+    """Test to validate setter for export signal property of string type."""
+    ai_voltage_chan_task.export_signals.start_trig_output_term = "RTSI"
+
+    with pytest.raises(DaqError) as e:
```
bkeryan, line 44: Style: rename to `exc_info` because `e` is usually the exception itself

tests/component/test_export_signal_properties.py
```diff
@@ -0,0 +1,63 @@
...
+
+import pytest
+
+from nidaqmx.constants import ExportAction
+from nidaqmx.error_codes import DAQmxErrors
+from nidaqmx.errors import DaqError
+from nidaqmx.task import Task
+
+
+def test__export_signal__get_enum_property__returns_value(ai_voltage_chan_task: Task):
```
bkeryan, line 11: Rename to ..._returns_default_value. Same for the other tests that validate default values.

tests/component/test_export_signal_properties.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test__export_signal__set_string_property__returns_assigned_value(ao_voltage_chan_task: Task):
+    """Test to validate setter for export signal property of string type."""
+    value_to_set = "RSE"
+    ao_voltage_chan_task.export_signals.start_trig_output_term = value_to_set
+
+    assert ao_voltage_chan_task.export_signals.start_trig_output_term == value_to_set
+
+
+def test__export_signal__reset_string_property__returns_initial_value(ao_voltage_chan_task: Task):
```
bkeryan, line 57: Change all references to "initial_value" to "default_value".

Reason: "default value" is the standard DAQmx terminology

bkeryan (changes requested):

tests/component/test_export_signal_properties.py
vigkre: Renamed to exc_info

tests/component/test_export_signal_properties.py
vigkre: Renamed get and reset test cases to _returns_default_value

tests/component/test_export_signal_properties.py
```diff
@@ -2,62 +2,81 @@
...
 
+import nidaqmx
 from nidaqmx.constants import ExportAction
 from nidaqmx.error_codes import DAQmxErrors
 from nidaqmx.errors import DaqError
 from nidaqmx.task import Task
 
 
-def test__export_signal__get_enum_property__returns_value(ai_voltage_chan_task: Task):
+@pytest.fixture(scope="module")
```
bkeryan, line 12: Same as the other PR: scope=module reuses the same task for multiple test cases. This may make test failures harder to debug. For example, if you added an error test that sets an invalid value and verifies the task, then other test cases would raise an error when they verify the task.

tests/component/test_export_signal_properties.py
```diff
@@ -2,62 +2,81 @@
...
+
+@pytest.fixture(scope="module")
+def ao_voltage_task(any_x_series_device):
+    """Gets AO voltage task."""
+    with nidaqmx.Task() as task:
+        task.ao_channels.add_ao_voltage_chan(any_x_series_device.ao_physical_chans[0].name)
+        yield task
+
+
+def test__export_signal__get_enum_property__returns_default_value(ai_voltage_task: Task):
```
bkeryan, line 28: These tests aren't given an "export signal". Rename to `test__ai_task__...`

tests/component/test_export_signal_properties.py
```diff
@@ -2,62 +2,81 @@
...
+    ai_voltage_task.export_signals.start_trig_output_term = "RTSI"
 
-    with pytest.raises(DaqError) as e:
-        _ = ai_voltage_chan_task.export_signals.start_trig_output_term
-    assert e.value.error_type == DAQmxErrors.INVALID_ROUTING_DESTINATION_TERMINAL_NAME_ROUTING
+    with pytest.raises(DaqError) as exc_info:
+        _ = ai_voltage_task.export_signals.start_trig_output_term
+    assert (
+        exc_info.value.error_type == DAQmxErrors.INVALID_ROUTING_DESTINATION_TERMINAL_NAME_ROUTING
+    )
```
bkeryan: The error is really caused by the set call, but it isn't reported until you verify the task. Instead of using a get call to verify the task, use task.control(TaskMode.VERIFY)

I think the set and verify should both be in the act section.

bkeryan (changes requested):

tests/component/test_export_signal_properties.py
vigkre: Renamed all the test cases to `test__ai_task__...`

tests/component/test_export_signal_properties.py
vigkre: Used `ai_voltage_task.control(TaskMode.TASK_VERIFY)` instead of set call.

tests/component/test_export_signal_properties.py
```diff
@@ -3,42 +3,42 @@
...
+@pytest.fixture(scope="function")
 def ai_voltage_task(any_x_series_device):
     """Gets AI voltage task."""
     with nidaqmx.Task() as task:
         task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
         yield task
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="function")
```
bkeryan, line 20: FYI, scope=function is the default

tests/component/test_export_signal_properties.py
```diff
@@ -47,33 +47,33 @@ def test__export_signal__reset_enum_property__returns_default_value(ai_voltage_t
...
+def test__ai_task__set_invalid_routing_destination__throws_daqerror(
     ai_voltage_task: Task,
 ):
     """Test to validate setter for export signal property of string type."""
-    ai_voltage_task.export_signals.start_trig_output_term = "RTSI"
-
     with pytest.raises(DaqError) as exc_info:
-        _ = ai_voltage_task.export_signals.start_trig_output_term
+        ai_voltage_task.export_signals.start_trig_output_term = "RTSI"
+        _ = ai_voltage_task.control(TaskMode.TASK_VERIFY)
```
bkeryan, line 61: control() doesn't return anything interesting.

bkeryan (approved):

tests/component/test_export_signal_properties.py
vigkre: Resolved

vigkre:
@zhindes Please merge this PR as we got required approvals

PR #256, opened 2023-04-03
nidaqmx: Remove Python 2.7 workarounds

bkeryan:

Run `pyupgrade --py37-plus`
- [pyupgrade](https://pypi.org/project/pyupgrade/) is intended to be run from a `pre-commit` hook, but I did a one-time run with PowerShell: `get-childitem -recurse "*.py" | foreach-object { poetry run pyupgrade --py37-plus $_ }`
- For `.py.mako` files, I ran `pyupgrade` on the output and then updated the templates accordingly.
- Here's what it did:
  - Remove refererences to `six.int_types`, etc. `int`/`long` and `str`/`unicode` have been unified in Python 3.x.
  - Remove explicit inheritance from `object`. Python 3.x does not support old-style classes, so explicitly inheriting from `object` is no longer necessary to opt-in to new-style classes.
  - Remove unnecessary parameter numbers from format strings. Parameter numbers like the `0` in `{0}` are only needed when repeating a format parameter.
  - Convert simple `.format` expressions to f-strings.
  - Simplify calls to `super()`

Remove other Python 2.7 workarounds:
- `_lib.py`: Don't encode lib name as ASCII
- `errors.py`: Remove `_ResourceWarning`

Remove dependency on `six` module.

Clean up vestigial Python 2.7 support.

Remove a dependency (`six`).

Ran `poetry run pytest` with Python 3.9.
Ran tests with Tox.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -159,7 +159,7 @@ Please include **all** of the following information when opening an issue:
 
   $ python -c "import sys; print(sys.version)"
 
-- The versions of the **nidaqmx**, numpy, six and enum34 packages used::
+- The versions of the **nidaqmx** and numpy packages used::
 
   $ python -m pip list
 
--- a/generated/nidaqmx/_lib.py
+++ b/generated/nidaqmx/_lib.py
@@ -2,7 +2,6 @@ from ctypes.util import find_library
 import ctypes
 from numpy.ctypeslib import ndpointer
 import platform
-import six
 import sys
 import threading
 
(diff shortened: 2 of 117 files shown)


zhindes (approved):
let me know when you've run the testes on the other python versions, and I can merge

bkeryan:
With test fixes from main:
```
  clean: commands succeeded
  py37: commands succeeded
  py38: commands succeeded
  py39: commands succeeded
  py310: commands succeeded
  py311: commands succeeded
  report: commands succeeded
  congratulations :)
```

PR #266, opened 2023-04-05
[DO NOT SUBMIT] Diff src/nidaqmx/ against generated/nidaqmx/

bkeryan:


src/nidaqmx/constants.py
```diff
@@ -62,11 +62,6 @@ class AcquisitionType(Enum):
     HW_TIMED_SINGLE_POINT = 12522  #: Acquire or generate samples continuously using hardware timing without a buffer. Hardware timed single point sample mode is supported only for the sample clock and change detection timing types.
 
 
-class Action(Enum):
-    COMMIT = 0  #: Commit
-    CANCEL = 1  #: Cancel
-
```
bkeryan, line 68: We should mention the removed enums in CHANGELOG.md

bkeryan: https://github.com/ni/nidaqmx-python/pull/268/commits/acc1f1458280479c691b30a8a014e7537ec2aaca

PR #267, opened 2023-04-05
Library Interpreter | PR1 |  Generate library Interpreter with simple c function calls

subash-suresh:
What does this PR accomplish?
This PR adds the required mako templates to generate the library interpreter with the required functions.

This PR consists of the following changes,
- Updated the `get_interpreter_functions` to have an option to filter out functions with the `grpc_only` tag.
- Created a mako file for the library interpreter generation. This template will call other mako templates based on the type of the function generated. 
- To get the mako template to be used for the c function call in the library interpreter, a helper method called `get_c_function_call_template` has been added.  Currently this method always returns the `default_c_function_call.py.mako`, but as we add logic for other types of c function calls (like calling the c functions twice in case of ivi dance scenario or calling c function for the register event functions) we could update this logic.
- The `default_c_function_call.py.mako` has been added to make simple c function calls.

Note: The list of functions filtered using the `grpc-only` tag in the metadata, might lead to missing out the property related functions and some of the handwritten functions, efforts are being taken parallelly to include all the required functions by updating the logic in scrapigen and in the `get_interpreter_functions` method in the intepreter helpers.


subash-suresh:
@bkeryan  and @zhindes 
This is work in progress PR based out of #238.
I would like to get some thoughts and ideas on the direction we are taking with the implementation in this PR.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -51,6 +53,37 @@ def get_interpreter_functions(metadata):
...
+                function_call_args.append(f"len({param.parameter_name})")
+        else:
+            if param.has_explicit_buffer_size:
+                function_call_args.append(param.parameter_name)
+                function_call_args.append("temp_size")
+            else:
+                function_call_args.append(f"ctypes.byref({param.parameter_name})")
+
+    if function_metadata.calling_convention == "Cdecl":
+        function_call_args.append("None")
```
bkeryan, line 82: This is sometimes correct but it's not correct in the general case.

In general, the DAQmx C API uses `__cdecl` for varargs functions (with a few exceptions). There are a few common patterns for designing varargs functions:
- [`printf`-style](https://man7.org/linux/man-pages/man3/printf.3.html): a format string parameter specifies how to unpack the variable arguments
- [`exec`-style](https://man7.org/linux/man-pages/man3/exec.3.html): the variable arguments specify array elements, with a null parameter to terminate the array.
- [`ioctl`-style](https://man7.org/linux/man-pages/man3/ioctl.3p.html): a numeric parameter specifies which operation to execute and how to interpret the variable arguments.

Most DAQmx programmable power-up state functions are execl-style. These need a null argument as a terminator. (At some point, we started using arrays instead of varargs, to make it easier to choose the number of arguments at run time. I guess that would be execv-style.)

Generic attribute get/set functions like `DAQmxGetChanAttribute`, `DAQmxSetChanAttribute` are ioctl-style. They do not need a null argument as a terminator.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -0,0 +1,2148 @@
...
+            with cfunc.arglock:
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        lib_importer.task_handle, ctypes_byte_str,
+                        ctypes.c_int, ctypes.c_double, ctypes.c_uint]
+
+        error_code = cfunc(
+            task, trigger_source, trigger_slope, trigger_level,
+            pretrigger_samples)
+        check_for_error(error_code)
```
bkeryan, line 55: FYI, one notable difference from nimi-python is that DAQmx uses `ctypes_byte_str` to encode all string inputs as ASCII, but nimi-python uses `create_string_buffer` to convert to a specific encoding: https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/_library_interpreter.py#L98

Fixing this is outside the scope of this project. If one did fix it, the best fix would be different from nimi-python: load the UTF-8 version of the DAQmx C API (`nicai_utf8.dll`) and encode all string inputs as UTF-8.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -27,11 +27,13 @@
...
 
 
-def get_interpreter_functions(metadata):
+def get_interpreter_functions(metadata, exclude_grpc_only_functions=False):
     """Converts the scrapigen metadata into a list of functions."""
     all_functions = deepcopy(metadata["functions"])
     functions_metadata = []
     for function_name, function_data in all_functions.items():
+        if exclude_grpc_only_functions and is_grpc_only_function(function_data):
+            continue
```
bkeryan, line 36: All of the read/write functions are marked grpc-only. What is the plan for supporting those? Add hand-written functions that redirect to https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/read_functions.py ?

src/codegen/templates/library_interpreter_templates/default_c_function_call.py.mako
```diff
@@ -0,0 +1,25 @@
...
+<%
+    from codegen.utilities.interpreter_helpers import generate_interpreter_function_call_args
+    from codegen.utilities.function_helpers import get_arguments_type
+    from codegen.utilities.text_wrappers import wrap, docstring_wrap
+%>\
+
+        cfunc = lib_importer.${'windll' if function.calling_convention == 'StdCall' else 'cdll'}.DAQmx${function.c_function_name}
+\
+## Create argument ctypes types list.
+    %if function.calling_convention == 'StdCall':
```
bkeryan, line 11: I see code for handling `Cdecl` but this if-statement skips those functions.

I guess they will have different locking, though. Instead of using double-checked locking to lazily initialize `cfunc.argtypes`, they always initialize it and they hold the lock while calling into the DAQmx C API.

bkeryan (changes requested):

src/codegen/utilities/interpreter_helpers.py
keer1111: Brad, the current plan considered for read/write functions to remove the "grpc-only" mark so it will be considered for code generation. Since the read/write functions does not belong to a class name, it will be only considered for library interpreter generation & hence can be removed from the hand written wrapper `read_functions.py` as you have mentioned. Still out of these functions, two are not present in DAQmxAPISharp.json. Hence, those might have to be handwritten without the details metadata. 

This is the plan and the metadata is yet to be updated in grpc-device-scrapigen repo and tested. 

Meanwhile we have started the update of metadata for other function cases here - [in this PR](https://github.com/ni/grpc-device-scrapigen/pull/156/files). This PR covers update for standard Cdll & windll calls (Excluding read/write) and Event calls.

src/codegen/utilities/interpreter_helpers.py
subash-suresh: Thanks for the information, based on this we have decided to take up the logic for handling cdecl calls in a separate PR. Currently this is being tracked by [Task 2357886](https://ni.visualstudio.com/DevCentral/_workitems/edit/2357886): Handle the cdecl calls in library interpreter

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -34,12 +34,24 @@
     "SetSyncPulseTimeWhen",
 ]
 
+LIBRARY_INTERPRETER_IGNORED_FUNCTIONS = [
```
subash-suresh, line 37: This is a temporary structure, used to handle these functions. This would be removed once logic has been added to handle these function calls

src/codegen/templates/_base_interpreter.py.mako
```diff
@@ -2,7 +2,7 @@
     from codegen.utilities.interpreter_helpers import get_interpreter_functions,get_interpreter_params,get_interpreter_parameter_signature
     from codegen.utilities.function_helpers import order_function_parameters_by_optional
     from codegen.utilities.text_wrappers import wrap, docstring_wrap
-    functions = get_interpreter_functions(data)
+    functions = get_interpreter_functions(data, True)
```
zhindes, line 5: can you use a named argument here so I can tell what `True` means?

zhindes (approved):

src/codegen/templates/library_interpreter_templates/default_c_function_call.py.mako
```diff
@@ -0,0 +1,25 @@
+<%page args="function"/>\
+<%
+    from codegen.utilities.interpreter_helpers import generate_interpreter_function_call_args
+    from codegen.utilities.function_helpers import get_arguments_type
+    from codegen.utilities.text_wrappers import wrap, docstring_wrap
+%>\
+
+        cfunc = lib_importer.${'windll' if function.calling_convention == 'StdCall' else 'cdll'}.DAQmx${function.c_function_name}
```
bkeryan, line 8: Please remove the extra blank line above `cfunc = ...` where the docstring would go

bkeryan (approved):
Approved with suggestions:
- Remove extra blank line

bkeryan (approved):

PR #271, opened 2023-04-06
gRPC Interpreter | PR1 | Add protos and generate stubs for gRPC client

Sakthi-SM:
* Adds the necessary gRPC device protos for generating the stubs.
* Added the gRPC dependencies to the poetry
* Generated the stubs using gRPC compiler

This implements [Task 2351280](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351280): Add proto and gRPC stubs into nidaqmx-python


src/codegen/protos/session.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
```
bkeryan, line 1: Because we are duplicating the `session.proto` codegen in nidaqmx-python, nimi-python, and measurementlink-python, we must use exactly the same version of `session.proto` for all three, or else users will get errors when using nidaqmx-python with either nimi-python or measurementlink-python. 

For now, the solution is to copy `session.proto` from nimi-python or measurementlink-python.

As a manual test, run `poetry run pip install ni-measurementlink-service` and `poetry run pip install nidcpower`, then `poetry run python`, `import nidaqmx.stubs.nidaqmx_pb2_grpc`, `import ni_measurementlink_service`, and `import nidcpower.nidcpower_pb2_grpc`.

The long term solution will probably involve packaging shared protos in separate Python modules and is tracked by AB#2250775: session.proto versioning is extremely brittle in Python and **will** break when MeasurementLink or nimi-python upgrades

pyproject.toml
```diff
@@ -59,6 +59,7 @@ click = "^8.1"
 Mako = "^1.2"
 ni-python-styleguide = "^0.4"
 sphinx_rtd_theme = "1.0"
+poetry-grpc-plugin = "^0.1.0"
```
bkeryan, line 62: We should be more careful about gRPC/protobuf versioning. Please add grpcio, grpcio-tools, and protobuf so that we can control their versions.

For measurementlink-python, we are synchronizing the main package's oldest compatible grpcio with the version of grpcio-tools used for codegen: https://github.com/ni/measurementlink-python/blob/main/pyproject.toml
However, the nidaqmx package won't depend on grpcio, so we will have to rely on runtime errors to notify users about mismatches.

bkeryan, line 62: poetry-grpc-plugin requires poetry >=1.2, but the build-system section below says this package requires poetry >= 1.1.12. I have been using 1.1.15 because most internal NI components haven't upgraded past that yet.

generated/nidaqmx/stubs/nidaqmx_pb2.py
```diff
@@ -0,0 +1,1978 @@
+# -*- coding: utf-8 -*-
```
bkeryan, line 1: Please update CONTRIBUTING.md to say what command line to run to regenerate the gRPC stubs. 

See the note about copying measurementlink-python's generate_grpc_stubs.py before you do this.

bkeryan (changes requested):

generated/nidaqmx/stubs/nidaqmx_pb2.py
bkeryan, line 1: We are likely to move our gRPC stubs to a separate Python package, so please make them private by renaming `nidaqmx/stubs/` to `nidaqmx/_stubs/`

pyproject.toml
bkeryan: Update: let's start using the "extras" feature to capture the run-time dependencies.
https://python-poetry.org/docs/pyproject/#extras

That way, clients can explicitly enable gRPC support: 
https://github.com/ni/measurementlink-python/issues/239

```
grpcio = { version = "^1.49.0", optional = true }

[tool.poetry.extras]
grpc = ["grpcio", "protobuf"]
```

pyproject.toml
```diff
@@ -59,6 +60,9 @@ click = "^8.1"
...
+grpcio = { version = "^1.49.0", optional = true }
+protobuf = "^3.20"
```
Sakthi-SM, line 65: I had to downgrade this version because I was getting the error shown below and one of the suggestions was to downgrade this. I hope this is the right way to resolve this. Please share your comments on this @bkeryan.
image

pyproject.toml
Sakthi-SM: Do we need to specifically resolve this by mentioning the expected version number?

zhindes (approved):

pyproject.toml
bkeryan: No, this is not the right way to resolve this. I'll look into it.

bkeryan:
Here are the correct packages and versions to use in `pyproject.toml`.

```
diff --git a/pyproject.toml b/pyproject.toml
index 69b0999..2744a3c 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,8 @@ importlib_metadata = {version="^4.2", python="~3.7"}
 # See https://github.com/readthedocs/readthedocs.org/issues/4912.
 Sphinx = {version="^4.3", optional=true}
 sphinx_rtd_theme = {version="^1.0", optional=true}
+grpcio = { version = "^1.49.0", optional = true }
+protobuf = { version = "^4.21", optional = true }

 [tool.poetry.extras]
 docs = ["sphinx", "sphinx_rtd_theme"]
@@ -60,9 +62,10 @@ click = "^8.1"
 Mako = "^1.2"
 ni-python-styleguide = "^0.4"
 sphinx_rtd_theme = "1.0"
-poetry-grpc-plugin = "^0.1.0"
-grpcio = { version = "^1.49.0", optional = true }
-protobuf = "^3.20"
+grpcio-tools = "1.49.1"
+mypy = ">=1.0"
+mypy-protobuf = ">=3.4"
+types-protobuf = "^4.21"

 [tool.black]
 line-length = 100
```

bkeryan (changes requested):

CONTRIBUTING.md
```diff
@@ -107,6 +107,15 @@ can be verified once that has been completed.
    * If updating a minor or major version, update `version` and `release` in `docs/conf.py`.
    * Add a section to `CHANGELOG.md` for the new version with empty subsections.
 
+# Updating gRPC stubs when the .proto file is modified
+
+The `generated\__internal\_stubs` directory contains the auto-generated Python files based on NI DAQmx protobuf (`.proto`) file.
```
bkeryan, line 112: The path is incorrect.
```suggestion
The `generated\nidaqmx\_stubs` directory contains the auto-generated Python files based on NI DAQmx protobuf (`.proto`) file.
```

CONTRIBUTING.md
```diff
@@ -107,6 +107,15 @@ can be verified once that has been completed.
...
+
+The latest NI DAQmx .proto file is available in [Azure Repo](https://github.com/ni/grpc-device/blob/main/generated/nidaqmx/nidaqmx.proto). From the Azure Repo manually download and overwrite the `.proto` file under the location `codegen\protos\nidaqmx.proto`.
```
bkeryan, line 114: NI-DAQmx should have a dash.
```suggestion
The `generated\__internal\_stubs` directory contains the auto-generated Python files based on NI-DAQmx protobuf (`.proto`) file.

The latest NI-DAQmx .proto file is available in [Azure Repo](https://github.com/ni/grpc-device/blob/main/generated/nidaqmx/nidaqmx.proto). From the Azure Repo manually download and overwrite the `.proto` file under the location `codegen\protos\nidaqmx.proto`.
```

CONTRIBUTING.md
bkeryan, line 112: This is also missing the word "the".
```suggestion
The `generated\__internal\_stubs` directory contains the auto-generated Python files based on the NI DAQmx protobuf (`.proto`) file.
```

pyproject.toml
```diff
@@ -48,6 +48,7 @@ sphinx_rtd_theme = {version="^1.0", optional=true}
 
 [tool.poetry.extras]
 docs = ["sphinx", "sphinx_rtd_theme"]
+grpc = ["grpcio", "protobuf"]
```
bkeryan, line 51: We need to run tests both with and without extras. `tox.ini` uses `poetry install -v`, which doesn't install extras, but it installs dev-dependencies, which means it still installs `grpcio` and `protobuf`.

Perhaps we should split our dev-dependencies into dependency groups `test` (e.g. pytest) and `codegen` (e.g. grpcio-tools, ...)? https://python-poetry.org/docs/pyproject/#dependencies-and-dependency-groups

@zhindes @mshafer-NI Do you have any suggestions?

bkeryan (changes requested):

pyproject.toml
bkeryan: We should also add a tox environment called `py39_no_grpc` or something like that. I don't think we need to test each Python version for the no-gRPC case.

bkeryan: First draft of dependency groups: keep dev group, convert to Poetry 1.2 syntax, and add codegen group.

```
diff --git a/pyproject.toml b/pyproject.toml
index 69b0999..98c8eea 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,24 +45,28 @@ importlib_metadata = {version="^4.2", python="~3.7"}
 # See https://github.com/readthedocs/readthedocs.org/issues/4912.
 Sphinx = {version="^4.3", optional=true}
 sphinx_rtd_theme = {version="^1.0", optional=true}
+grpcio = { version = "^1.49.0", optional = true }
+protobuf = { version = "^4.21", optional = true }

 [tool.poetry.extras]
 docs = ["sphinx", "sphinx_rtd_theme"]
 grpc = ["grpcio", "protobuf"]

-[tool.poetry.dev-dependencies]
-# Test
+[tool.poetry.group.dev.dependencies]
 pytest = ">=7.2"
 pytest-cov = ">=4.0"
 pykka = "^3.0"
 tox = "^3.24"
-click = "^8.1"
-Mako = "^1.2"
 ni-python-styleguide = "^0.4"
 sphinx_rtd_theme = "1.0"
-poetry-grpc-plugin = "^0.1.0"
-grpcio = { version = "^1.49.0", optional = true }
-protobuf = "^3.20"
+mypy = ">=1.0"
+types-protobuf = "^4.21"
+
+[tool.poetry.group.codegen.dependencies]
+click = "^8.1"
+Mako = "^1.2"
+grpcio-tools = "1.49.1"
+mypy-protobuf = ">=3.4"

 [tool.black]
 line-length = 100
@@ -72,5 +76,5 @@ extend_exclude = "docs,generated,src/codegen/metadata,src/codegen/templates"
 extend_exclude = "docs,generated,src/codegen/metadata,src/codegen/templates,src/handwritten"

 [build-system]
-requires = ["poetry>=1.1.12"]
+requires = ["poetry>=1.2"]
 build-backend = "poetry.masonry.api"
\ No newline at end of file
```

I verified that `poetry1.2 install --without codegen` does not install grpcio, protobuf, click, etc., and that `poetry1.2 install --without codegen --extras grpc` installs grpcio and protobuf.

bkeryan:
On further thought, this project already has a codegen tool, and we are adding a second one.

> Once the necessary changes are done, update the auto-generated code using `poetry run python src/codegen --dest generated/nidaqmx`

It would make sense to update `src/codegen/__main__.py` to call `generate_grpc_stubs.main()`.

src/codegen/generate_grpc_stubs.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def blacken_code(stubs_path: pathlib.Path):
+    """Run black on generated files."""
+    print("Running black")
+    for py_path in itertools.chain(stubs_path.rglob("*.py"), stubs_path.rglob("*.pyi")):
+        if black.format_file_in_place(
+            src=py_path, fast=False, mode=Mode(line_length=100), write_back=black.WriteBack.YES
+        ):
+            print(f"reformatted {py_path}")
```
bkeryan, line 107: > It runs protoc, fixes up imports, creates __init__.py files, and even runs black (however it's probably better to just exclude the protoc output in the black config in pyproject.toml).

Reformatting the stubs is slow and does not have much value. Let's delete this and exclude the stubs instead.

src/codegen/protos/session.proto
Sakthi-SM: Updated the `session.proto` from nimi-python.

pyproject.toml
Sakthi-SM: As suggested, I have added the dependency groups and updated the poetry version. Do we need to create a tox environment as part of this PR or can we take that up while updating the tests?

generated/nidaqmx/_stubs/nidaqmx_pb2_grpc.py
```diff
@@ -0,0 +1,12870 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
+"""Client and server classes corresponding to protobuf-defined services."""
+import grpc
+
+from nidaqmx._stubs import nidaqmx_pb2 as nidaqmx__pb2
```
bkeryan, line 5: Ok, this looks correct.

pyproject.toml
bkeryan: You can add the tox environment when updating the tests.

bkeryan (approved):
Approved with suggestions:
- Delete Poetry 1.1 dev-dependencies block.
- Rename `generate_grpc_stubs.py` to `stub_generator.py`

src/codegen/stub_generator.py
```diff
@@ -10,14 +9,15 @@
...
+STUBS_PATH = (
+    pathlib.Path(__file__).parent.parent.parent / "generated" / STUBS_NAMESPACE.replace(".", "/")
+)
 PROTO_PATH = pathlib.Path(__file__).parent.parent.parent / "src" / "codegen" / "protos"
 PROTO_FILES = list(PROTO_PATH.rglob("*.proto"))
 
 
 def generate_stubs():
     """Generate and fixup gRPC Python stubs."""
-    remove_generated_files(STUBS_PATH, PROTO_PATH)
```
bkeryan: Why delete this? Is it because [the PR workflow](https://github.com/ni/nidaqmx-python/blob/master/.github/workflows/PR.yml) already deletes generated/nidaqmx?

PR #281, opened 2023-04-13
Library Interpreter | PR0 | Update the codegen to work with the latest metadata files

subash-suresh:
- Updated the scrapigen metadata files from the latest metadata changes in scrapigen
- Added a check in the `get_functions` method to filter the functions based upon the presence of `python_codegen_method` key.
- Added a logic to exclude certain functions from being generated.

Note: The metadata used for this update, is based upon https://github.com/ni/grpc-device-scrapigen/pull/159

This implements [Task 2356609](https://ni.visualstudio.com/DevCentral/_workitems/edit/2356609): Fix nidaqmx-python logic with latest change(adding metadata for handwritten functions) in scrapigen metadata


src/codegen/utilities/function_helpers.py
```diff
@@ -11,12 +11,37 @@
...
+    "SetDigitalPullUpPullDownStates",
+]
+
 
 def get_functions(metadata, class_name=""):
     """Converts the scrapigen metadata into a list of functions."""
     all_functions = deepcopy(metadata["functions"])
     functions_metadata = []
     for function_name, function_data in all_functions.items():
+        if "pythod_codegen_method" in function_data or function_name in EXCLUDED_FUNCTIONS:
```
zhindes, line 43: Is `pythod` a typo? Also I think it should be `"python_codegen_method" not in function_data` just going by the name.

zhindes (changes requested):

src/codegen/functions/function.py
```diff
@@ -38,7 +39,7 @@ def __init__(self, function_name, function_metadata):
                     self._parameters.append(FunctionParameter(parameter))
 
                     if parameter["direction"] == "out":
-                        self._output_parameters.append(self._parameters)
+                        self._output_parameters.extend(self._parameters)
```
bkeryan, line 42: This adds all of the parameters to `self._output_parameters`. I think you want just the last one, right?
```
self._output_parameters.append(self._parameters[-1])
```

bkeryan (changes requested):

src/codegen/functions/function.py
keer1111: The intention for using extend was because append was creating list of lists, because _parameters is a list and the whole list got added as list element instead of the only one parameter.

Your suggestion works better, will add the lastly added parameter if its 'out'

bkeryan (approved):

PR #282, opened 2023-04-13
Addition of Read Write functions to Library Interpreter

MounikaBattu17:

- Added handwritten Read/Write functions from `read_functions.py`, write_functions.py to `_library_interpreter.py` through `_library_interpreter.py.mako`
-  Removed complex datatypes like `enum` `FillMode` and hardcoded its value
- Made all read/write functions `public`
- Updated function names like `_read_analog_f_64` to `read_analog_f64` to align with `_base_interpreter.py`
- Removed the types conversion  `value = CtrFreq(freq.value, duty_cycle.value)` logic and updated to return the tuple `freq.value, duty_cycle.value` alone, the calling function will have to make the conversion.

Implements [Task 2351399](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351399): Generate the Library Interpreter inheriting the base - Handling Error methods for Read/Write Case

### DIFF
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -13,6 +13,12 @@ class LibraryInterpreter(BaseInterpreter):
     This class is responsible for interpreting the Library's C API.
 
     """
+    fill_mode = ctypes.c_int()
+    fill_mode = 0
+    interleaved = ctypes.c_int()
+    interleaved = 0
+    data_layout = ctypes.c_int()
+    data_layout = 0
 
     def calculate_reverse_poly_coeff(
             self, forward_coeffs, num_forward_coeffs_in, reverse_coeffs,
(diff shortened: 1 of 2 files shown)


generated/nidaqmx/_library_interpreter.py
```diff
@@ -13,6 +13,12 @@ class LibraryInterpreter(BaseInterpreter):
     This class is responsible for interpreting the Library's C API.
 
     """
+    fill_mode = ctypes.c_int()
```
MounikaBattu17, line 16: Please provide a suggestion if it could be simplified or not.

MounikaBattu17:
WRT to this comment https://github.com/ni/nidaqmx-python/pull/238#discussion_r1152595049 updated the function names from `read_analog_f_64` to `read_analog_f64`

MounikaBattu17:
The types conversion (`value = CtrFreq(freq.value, duty_cycle.value)`) logic will be updated in the calling functions in the upcoming PRs

PR #283, opened 2023-04-13
[Internal] Added Interpreter instance to the constructor of the generated Task and task related py files

MounikaBattu17:
- Created function `_select_interpreter` to get Interpreter object in `utils.py`
- Added interpreter in constructor of `task.py`, `stream_readers.py`, `stream_writters.py`
- Passed on the instance via constructor in all other generated files.

Implements [Task 2330456](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330456): Update generation of Task related classes accepting an Interpreter Instance via Constructor


keer1111 (changes requested):

src/codegen/templates/_task_modules/channels/ai_channel.py.mako
```diff
@@ -16,7 +16,7 @@ from nidaqmx._lib import (
 from nidaqmx.scale import Scale
 from nidaqmx.errors import (
     check_for_error, is_string_buffer_too_small, is_array_buffer_too_small)
-from nidaqmx._task_modules.channels.channel import Channel
+from generated.nidaqmx._task_modules.channels.channel import Channel
```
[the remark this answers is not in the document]
bkeryan: That's right. "generated" is not a Python module and should not have an `__init__.py` file, so it should be impossible to import `generated.nidaqmx.

zhindes (approved):

src/codegen/templates/_task_modules/channels/ai_channel.py.mako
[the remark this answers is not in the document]
MounikaBattu17: Removed Generated in import.

src/handwritten/stream_readers.py
```diff
@@ -35,6 +36,7 @@ def __init__(self, task_in_stream):
         self._in_stream = task_in_stream
         self._task = task_in_stream._task
         self._handle = task_in_stream._task._handle
+        self._interpreter = utils._select_interpreter()
```
bkeryan, line 39: Stream readers and writers should reuse the task's interpreter.

generated/nidaqmx/_task_modules/channels/channel.py
```diff
@@ -19,16 +19,18 @@ class Channel:
...
 
-    def __init__(self, task_handle, virtual_or_physical_name):
+    def __init__(self, task_handle, virtual_or_physical_name, interpreter):
         """
         Args:
             task_handle (TaskHandle): Specifies the handle of the task that
                 this channel is associated with.
             virtual_or_physical_name (str): Specifies the flattened virtual or
                 physical name of a channel.
+            interpreter: Specifies the interpreter instance assigned.
```
bkeryan, line 29: In all of these docstrings, when referring to the interpreter:
- Specify the type: `interpreter (BaseInterpreter): ...`
- Remove "assigned". "Specifies the interpreter instance." is perfectly clear without it.

src/handwritten/task.py
```diff
@@ -358,7 +359,7 @@ def triggers(self):
         """
         return self._triggers
 
-    def _initialize(self, task_handle):
+    def _initialize(self, task_handle, interpreter):
```
bkeryan, line 362: You need to update the "alternate task constructor" accordingly: https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/system/storage/_alternate_task_constructor.py

Also, please plan to add test coverage for it:
image

bkeryan (changes requested):

src/handwritten/task.py
keer1111: Sure Brad, will now update the AlternateTaskConstructor's usage of Task's initialize method by using a utils generated interpreter.

Since this class is part of System, we will take up during next [feature](https://dev.azure.com/ni/DevCentral/_workitems/edit/2270240). Is that okay?

Also Tracking the test coverage for same [here](https://dev.azure.com/ni/DevCentral/_workitems/edit/2365741) in the same feature.

bkeryan (approved):

keer1111 (approved):

PR #286, opened 2023-04-18
[Internal] Library Interpreter | PR0 | Handle parameters with explicit buffer size definitions

subash-suresh:
This PR handles the parameters which have explicit buffer size definitions associated with it. These sizes can be defined during the c function call in multiple ways and this is denoted by the different mechanisms defined in the functions metadata. 

The following changes have been made:
- The `has_explicit_buffer_size` is assigned a value of `True` if the `size` field in defined in the parameter metadata.
- The logic to generate the function call arguments has been updated to handle the different mechanisms of the size parameters
- The logic to generate the argument types for the c function calls has also been updated for the same.
- removed the `passed-in` value from the ignored mechanisms list, as this value is required to be considered as a parameter.
- Added new functions to work with the `ivi-dance` mechanism.
- Added a new library interpreter template called `double_cfunc_call.py.mako` for the function with parameters having the `ivi-dance` mechanism.


generated/nidaqmx/_base_interpreter.py
```diff
@@ -1204,7 +1204,8 @@ def load_task(self, session_name):
 
     @abc.abstractmethod
     def read_analog_f64(
-            self, task, num_samps_per_chan, timeout, fill_mode, read_array):
+            self, task, num_samps_per_chan, timeout, fill_mode,
```
subash-suresh, line 1207: This signature is in line with the grpc signature of this function and hence this is being followed.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2285,7 +2319,8 @@ def read_analog_f64(
 
         error_code = cfunc(
             task, num_samps_per_chan, timeout, fill_mode,
-            ctypes.byref(read_array), ctypes.byref(samps_per_chan_read))
+            ctypes.byref(read_array), array_size_in_samps,
+            ctypes.byref(samps_per_chan_read))
         check_for_error(error_code)
         return read_array, samps_per_chan_read
```
bkeryan, line 2325: We need to handle numbers and strings differently from arrays.
- Python numbers are immutable, so the interpreter should always return them by value.
- Python strings are immutable, so the interpreter should always return them by value. However, in ctypes we use a mutable string buffer to interface with the DAQmx C API.
- Numpy arrays are mutable, and nidaqmx-python supports a number of functions that fill in preallocated arrays: https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/stream_readers.py . So we must pass them by reference, not by value.

This means:
- The interpreter signature for `read_analog_f64` should have a `read_array` parameter. It shouldn't return `read_array`.
- The interpreter signature for `read_analog_scalar_f64` should not have a `value` parameter. It should initialize `value = ctypes.c_double()`, pass `ctypes.byref(value)` to the C function, and return `value.value` like https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/channels/ai_channel.py#L5821
- The interpreter signature for `add_network_device` should not have a `device_name_out` parameter. It should initialize `device_name_out = ctypes.create_string_buffer(...)` and pass it to the C function like you are doing here, and it should return `device_name_out.value.encode('ascii')` like https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/channels/channel.py#L218

generated/nidaqmx/_library_interpreter.py
```diff
@@ -48,10 +48,21 @@ def add_network_device(
...
+            if is_string_buffer_too_small(size_or_code):
+                # Buffer size must have changed between calls; check again.
+                temp_size = 0
+            elif size_or_code > 0 and temp_size == 0:
+                # Buffer size obtained, use to retrieve data.
+                temp_size = size_or_code
+            else:
+                break
+        check_for_error(size_or_code)
         return device_name_out
```
bkeryan, line 66: This returns a raw ctypes string buffer. It should convert to a string: `return device_name_out.value.encode('ascii')`

bkeryan (changes requested):

generated/nidaqmx/_library_interpreter.py
bkeryan: BTW `read_array` isn't assigned before you call `ctypes.byref(read_array)`, so this probably returns `NameError: name 'read_array' is not defined`

bkeryan:
Please do some basic testing of the library interpreter with a test script such as this, but converted to use the library interpreter:
```
import pprint
from collections import namedtuple
import grpc
import nidaqmx
import nidaqmx.constants
import nidaqmx._grpc_interpreter
pp = pprint.PrettyPrinter(indent=4)
GrpcSessionOptions = namedtuple("GrpcSessionOptions", ["grpc_channel"])
grpc_channel = grpc.insecure_channel("localhost:31763")
interpreter = nidaqmx._grpc_interpreter.GrpcStubInterpreter(GrpcSessionOptions(grpc_channel))
task = interpreter.create_task("MyTask")
interpreter.create_ai_voltage_chan(task, "Dev1/ai0", "", nidaqmx.constants.TerminalConfiguration.DEFAULT.value, -10.0, 10.0, nidaqmx.constants.VoltageUnits.VOLTS.value, "")
value = interpreter.read_analog_scalar_f64(task, 10.0)
pp.pprint(value)
interpreter.clear_task(task)
```

generated/nidaqmx/_library_interpreter.py
subash-suresh: > BTW `read_array` isn't assigned before you call `ctypes.byref(read_array)`, so this probably returns `NameError: name 'read_array' is not defined`

I have raised another #287  on adding initialization logic for the output parameters used in the interpreter functions.

zhindes (approved):

generated/nidaqmx/_library_interpreter.py
bkeryan: > So is this ok to change the interpreter to accept the read array input

Yes.

The LibraryInterpreter should pass the `read_array` parameter to the C DLL.
The GrpcStubInterpreter should assign `read_array[:] = reponse.read_array`.

subash-suresh: Sure brad, i have created a task to track this [Task 2371633](https://ni.visualstudio.com/DevCentral/_workitems/edit/2371633): Handle the read functions with read array inputs in library interpreter

keer1111:
Tried out the testing Brad has suggested above, still the creation of task is failing because of missing initialization of the return variable. This is addressed in the next PR - #287 . Hence merging this PR.

PR #287, opened 2023-04-18
Library Interpreter | PR1| Add instantiation lines for output parameters

subash-suresh:
Currently the library interpreter functions that return values are using certain variables to pass in by reference in the c function call to get the return values. These variables are not initialized, and this PR adds logic to initialize them.

- The scrapigen metadata has been updated based on https://github.com/ni/grpc-device-scrapigen/pull/163
- Added a new method in `interpreter_helpers` called `get_instantiation_lines_for_output_params` to get the lines for initialization of output parameters.
- Added a new method called `get_output_parameters` to get the output parameters related to interpreters.
- Added two new properties to `functions.py`
    -  ` interpreter_parameters`: gets all parameters used in the interpreters
    - ` is_init_method`: identifies if the method used is an init method or not. The task object instantiation is done based on this.
    - `new_session_initialized` parameter has been handled separately as this is a unique case and this parameter has always been set to `True` in case of the library interpreter.
- Added logic in mako templates to only generate code for `StdCall` calling convention and throw Not Implemented exceptions for other calling conventions
- Updated the library interpreter generate the lines for initialization.


src/codegen/utilities/interpreter_helpers.py
```diff
@@ -108,6 +108,22 @@ def get_interpreter_parameter_signature(is_python_factory, params):
     return ", ".join(params_with_defaults)
 
 
+def get_instantiation_lines_for_output_params(output_parameters):
```
keer1111, line 111: is the output a parameter passed by reference? or just the return value/output. Please update the name accordingly

keer1111 (approved):

src/codegen/utilities/interpreter_helpers.py
subash-suresh: updated as per suggestion

subash-suresh:
The scrapigen metadata from https://github.com/ni/grpc-device-scrapigen/pull/162 has been used to generate the files in this PR

zhindes (approved):

generated/nidaqmx/_library_interpreter.py
```diff
@@ -106,6 +108,9 @@ def auto_configure_cdaq_sync_connections(
     def calculate_reverse_poly_coeff(
             self, forward_coeffs, num_forward_coeffs_in, min_val_x, max_val_x,
             num_points_to_compute, reverse_poly_order):
+        size = len(forward_coeffs) if (reverse_poly_order < 0) else reverse_poly_order + 1
```
bkeryan, line 111: If we are adding a special case for this function, then we should also add some test coverage.

It's an easy function to test because it doesn't require any hardware or tasks. It just does some math and returns the result.

bkeryan (changes requested):

src/codegen/metadata/functions.py
```diff
@@ -281,6 +281,7 @@
                 'python_type_annotation': 'List[float]',
                 'size': {
                     'mechanism': 'custom-code',
+                    'python_value': 'len(forward_coeffs) if (reverse_poly_order < 0) else reverse_poly_order + 1',
```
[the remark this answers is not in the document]
subash-suresh: updated

generated/nidaqmx/_library_interpreter.py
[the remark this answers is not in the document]
bkeryan: This should be fixed by https://github.com/ni/nidaqmx-python/pull/294

generated/nidaqmx/_library_interpreter.py
```diff
@@ -1774,7 +1781,6 @@ def create_task(self, session_name):
         error_code = cfunc(
             session_name, ctypes.byref(task))
         check_for_error(error_code)
-        return task
```
bkeryan, line 1777: Changing to use func.output_parameters removed `return task` because `Function.__init__` filters out task handles:
https://github.com/ni/nidaqmx-python/blob/master/src/codegen/functions/function.py#L35

I think we need to be more specific about what we mean by "output parameters". The Python API, C DLL, and gRPC proto file all have completely different ideas of what "output parameters" means.

Maybe the answer is that the interpreter codegen should continue to use base_parameters and do its own filtering.

bkeryan (changes requested):

generated/nidaqmx/_library_interpreter.py
```diff
@@ -106,6 +108,9 @@ def auto_configure_cdaq_sync_connections(
     def calculate_reverse_poly_coeff(
             self, forward_coeffs, num_forward_coeffs_in, min_val_x, max_val_x,
             num_points_to_compute, reverse_poly_order):
+        size = len(forward_coeffs) if (reverse_poly_order < 0) else reverse_poly_order + 1
+        reverse_coeffs = numpy.zeros(size, dtype=numpy.numpy.float64)
```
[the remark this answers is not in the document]
bkeryan: The byref is fixed and the function parameters look correct, but argtypes still has an extra ctypes_byte_str that doesn't belong.

```
>>> interpreter.calculate_reverse_poly_coeff([1.0, 2.0], -10.0, 10.0, 1000, 2)
  File "D:\dev\nidaqmx-python\generated\nidaqmx\_library_interpreter.py", line 126, in calculate_reverse_poly_coeff
    error_code = cfunc(
TypeError: this function takes 8 arguments (7 given)
```

I removed the ctypes_byte_str in my copy.

```
>>> interpreter.calculate_reverse_poly_coeff([1.0, 2.0], -10.0, 10.0, 1000, 2)
  File "D:\dev\nidaqmx-python\generated\nidaqmx\_library_interpreter.py", line 126, in calculate_reverse_poly_coeff
    error_code = cfunc(
ctypes.ArgumentError: argument 1: <class 'TypeError'>: argument must be an ndarray
```

Ok, it wants a numpy array as input but returns a list. That's a little strange, but this function is sort of unique.

```
>>> interpreter.calculate_reverse_poly_coeff(numpy.asarray([1.0,2.0]), -10.0, 10.0, 1000, 2)
[-0.49999999999999883, 0.5000000000000009, -1.6306154203060172e-17]
>>> interpreter.calculate_reverse_poly_coeff(numpy.asarray([1.0,2.0]), -10.0, 10.0, 1000, 1)
[-0.500000000000001, 0.5000000000000009]
```

This looks like it's working.
y = 2x + 1
y - 1 = 2x
1/2 * y - 1/2 = x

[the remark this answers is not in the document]
subash-suresh: I have added a new test case with the above mentioned values

generated/nidaqmx/_library_interpreter.py
subash-suresh: Thanks for the suggestion, i have added the tests

generated/nidaqmx/_library_interpreter.py
subash-suresh: I have added a separate method to find the output parameters for interpreters and this issue is now fixed

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -62,22 +62,25 @@ def get_interpreter_functions(metadata):
...
         if param.direction == "in":
             function_call_args.append(param.parameter_name)
             if param.has_explicit_buffer_size:
                 function_call_args.append(f"len({param.parameter_name})")
+                size_value = param.size.value
+                size_param_index = len(function_call_args) - 1
         else:
             if param.has_explicit_buffer_size:
+                function_call_args.append(param.parameter_name)
                 if param.size.mechanism == "ivi-dance":
```
bkeryan, line 82: I think deleting the repeated parameter makes this function hard to follow. It would be clearer if the function didn't add a duplicate in the first place. 

Initializing `size_value` to a list and comparing it to a string doesn't help matters.

```suggestion
    params_to_skip = []
    for param in function_metadata.interpreter_parameters:
        if param.parameter_name in params_to_skip:
            continue
        if param.direction == "in":
            function_call_args.append(param.parameter_name)
            if param.has_explicit_buffer_size:
                function_call_args.append(f"len({param.parameter_name})")
                params_to_skip.append(param.size.value)
        else:
            if param.has_explicit_buffer_size:
                function_call_args.append(param.parameter_name)
                if param.size.mechanism == "ivi-dance":
```

BTW, this still assumes the array is immediately followed by the array size. If this assumption is valid, that's okay. Avoiding this assumption would be more complicated: add another loop to pre-scan for array parameters and change the main loop to add `len(array)` when processing the size parameter, not when processing the array parameter.

tests/component/test_scale_operations.py
```diff
@@ -0,0 +1,21 @@
+"""Test for validating different scale related operations."""
+import pytest
+
+import nidaqmx
+
+
+def test__polynomial__calcualte_reverse_coeff__returns_expected_reverse_coeff():
```
bkeryan, line 7: The "then" clause always describes the expected behavior, so using the word "expected" is redundant.

Also, the "when" clause doesn't match the spelling of the function name.

```suggestion
def test__polynomial__calculate_reverse_poly_coeff__returns_reverse_coeff():
```

tests/component/test_scale_operations.py
```diff
@@ -0,0 +1,21 @@
...
+    max_val_x = 10.0
+    num_of_points_to_compute = 1000
+    reverse_polynomial_order = 1
+
+    reverse_coeff = nidaqmx.Scale.calculate_reverse_poly_coeff(
+        forward_coeff, min_val_x, max_val_x, num_of_points_to_compute, reverse_polynomial_order
+    )
+
+    # The expected inverted polynomial is 0.5y - 0.5 = x
+    assert pytest.approx(reverse_coeff) == [-0.5, 0.5]
```
bkeryan, line 21: Style: I think it reads better if you specify approx on the expected value ("assert reverse_coeff equals approx [-0.5, 0.5]")

```suggestion
    assert reverse_coeff == pytest.approx([-0.5, 0.5])
```

tests/component/test_scale_operations.py
bkeryan, line 1: I overlooked these guidelines when we added the property tests:

> Component and Unit Tests
> - Organized to parallel production code (and test utility code)
> - The test file name is test_*.py, where * is the name of the entity tested, e.g. the name of the class, optionally followed by flavor or differentiator, or just the name of the module being unit tested. The name would follow snake_case style. Example: if the class is MyClass, the associated test file would be test_my_class.py.

Name this file `test_scale.py`. I don't think we should use "operations" as a flavor/differentiator because it is too vague.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -1758,12 +1754,14 @@ def create_table_scale(
...
-            name, prescaled_vals, len(prescaled_vals), num_prescaled_vals_in,
-            scaled_vals, len(scaled_vals), num_scaled_vals_in,
-            pre_scaled_units, scaled_units)
+            name, prescaled_vals, len(prescaled_vals), scaled_vals,
+            len(scaled_vals), pre_scaled_units, scaled_units)
         check_for_error(error_code)
 
     def create_task(self, session_name):
+        task = lib_importer.task_handle(0)
+        task = ctypes.TaskHandle()
```
bkeryan, line 1763:
```
>>> task=interpreter.create_task("MyTask")
  File "D:\dev\nidaqmx-python\generated\nidaqmx\_library_interpreter.py", line 1763, in create_task
    task = ctypes.TaskHandle()
AttributeError: module 'ctypes' has no attribute 'TaskHandle'
```

bkeryan (changes requested):

tests/component/test_scale_operations.py
subash-suresh: updated as per suggestions

tests/component/test_scale_operations.py
subash-suresh: updated as per suggestions

generated/nidaqmx/_library_interpreter.py
subash-suresh: This problem has been fixed and the unwanted instantiation has been removed

tests/component/test_scale_operations.py
subash-suresh: Updated as per suggestions

src/codegen/utilities/interpreter_helpers.py
bkeryan: Thanks for explaining.

Adding `len(array)` immediately after the array makes an assumption about the parameter order. This assumption is not valid in the case where multiple arrays share the same size parameter.

Can we eliminate that assumption and generate the size parameter in the right place?
```
    size_values = {}
    for param in function_metadata.interpreter_parameters:
        if param.direction == "in" and param.has_explicit_buffer_size:
            size_values.setdefault(param.size.value, f"len({param.parameter_name})")
        elif param.direction == "out" and param.has_explicit_buffer_size and param.size.mechanism == "ivi-dance":
            size_values.setdefault(param.size.value, "temp_size")

    for param in function_metadata.interpreter_parameters:
        if param.parameter_name in size_values:
            function_call_args.append(size_values[param.parameter_name])
        elif param.direction == "in":
            function_call_args.append(param.parameter_name)
        else:
```

If this won't work either, then:
- Update the comment to document the assumption you are making and the case where the assumption is not valid.
- Initialize size_value to the correct data type.

subash-suresh: Thanks for the logic suggestion, I used a similar implementation with some slight modifications to achieve the expected behavior.

bkeryan: That still appends the size parameter while iterating over the array parameter, which is a big assumption.

My suggestion eliminates this assumption by appending the size parameter while iterating over the size parameter. This ensures that each parameter is appended in the right order, no matter what.

I tried it and it works. Here it is with your style changes:
```
def generate_interpreter_function_call_args(function_metadata):
    """Gets function call arguments."""
    function_call_args = []
    size_values = {}
    for param in function_metadata.interpreter_parameters:
        if param.has_explicit_buffer_size:
            if param.direction == "in":
                size_values[param.size.value] = f"len({param.parameter_name})"
            elif param.direction == "out":
                if param.size.mechanism == "ivi-dance":
                    size_values[param.size.value] = "temp_size"

    for param in function_metadata.interpreter_parameters:
        if param.parameter_name in size_values:
            function_call_args.append(size_values[param.parameter_name])
        elif param.direction == "in":
            function_call_args.append(param.parameter_name)
        elif param.direction == "out":
            if param.has_explicit_buffer_size:
                function_call_args.append(param.parameter_name)
            else:
                function_call_args.append(f"ctypes.byref({param.parameter_name})")
    return function_call_args
```

subash-suresh: Here in the logic the keys in the `size_value` dictionary are representing the `parameter.size.value`, but when accessing the dictionary the logic tries to access it through param.parameter_name and hence the `if param.parameter_name in size_values` is never true and none of the size params are getting added in my case.

subash-suresh: Thanks for the clarification, I have added a code comment as part of this PR to explain my assumptions, I will take up this logic in a separate PR as it requires some refactoring in the parameter filtering, we do now. Currently we filter out all the size value parameters which run based on `ivi-dance` mechanism and also any parameter with the name `size`.
This logic has to reverted in order for the suggested implementation to completely work for all functions, hence I am planning to take this up in a separate PR.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -1758,12 +1754,14 @@ def create_table_scale(
...
-            name, prescaled_vals, len(prescaled_vals), num_prescaled_vals_in,
-            scaled_vals, len(scaled_vals), num_scaled_vals_in,
-            pre_scaled_units, scaled_units)
+            name, prescaled_vals, len(prescaled_vals), scaled_vals,
+            len(scaled_vals), pre_scaled_units, scaled_units)
         check_for_error(error_code)
 
     def create_task(self, session_name):
+        task = lib_importer.task_handle(0)
+        new_session_initialized = True
```
subash-suresh, line 1763: the `new_session_initialized` parameter has been handled separately in the library interpreter generation and this is always set to `True`

bkeryan (approved):

src/codegen/utilities/interpreter_helpers.py
bkeryan: Okay.

PR #288, opened 2023-04-19
Added gRPC options and interpreter instance to generated files

MounikaBattu17:

- Passed `grpc_options` and `interpreter` to constructor of `system.py` and `scale.py`.
- Passed `grpc_options` to constructor of `task.py` and `watchdog.py` and `_alternate_task_constructor.py`.
- Added `grpc_session_options.py` class for `grpc_options`.
- Passed `interpreter` instance to `ChannelCollection.py`
- Passed `interpreter` instance to all system folder files

`device.py`
`physical_channel.py`
`watchdog.py`
**collections:**
```
device_collection.py
persisted_channel_collection.py
persisted_scale_collection.py
persisted_task_collection.py
physical_channel_collection.py
```
 **watchdog_modules:**
```
expiration_state.py
expiration_states_collection.py
```

**storage:**
```
_alternate_task_constructor.py
persisted_channel.py
persisted_scale.py
persisted_task.py
```

Implements [Task 2351359](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351359): Update all System related classes to init with grpc options
[Task 2351368](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351368): Update all Scale class to init with grpc options


zhindes (approved):

src/handwritten/system/storage/_alternate_task_constructor.py
```diff
@@ -17,14 +17,15 @@ class _TaskAlternateConstructor(Task):
     derived class to 'overload' the constructor.
     """
 
-    def __init__(self, task_handle):
+    def __init__(self, task_handle,*, grpc_options=None, interpreter=None):
         """
         Args:
             task_handle: Specifies the task handle from which to create a
                 Task object.
+            grpc_options: Specifies the gRPC session options.
```
keer1111, line 25: we are intentionally avoiding interpreter docstring, since interpreter will be passed only when it is initialized internally?

src/handwritten/grpc_session_options.py
```diff
@@ -0,0 +1,67 @@
+from enum import IntEnum
+
+
+# This constant specifies the gRPC package and service used by this API.
+# Customers can pass this value to the MeasurementLink discovery service to resolve the server instance that provides this interface.
+GRPC_SERVICE_INTERFACE_NAME = 'nidaqmx_grpc.NiDAQmx'
+
+# This constant specifies the API license key required by the NI gRPC Device Server that comes with
+# MeasurementLink 2023 Q1.
+MEASUREMENTLINK_23Q1_PYTHON_API_KEY = 'DE10751B-3EE0-44EC-A93B-800E6A3C89E4'
```
keer1111, line 10: We have reused the same key from nimi-python, @bkeryan , Please confirm if this is okay

src/codegen/templates/system/physical_channel.py.mako
```diff
@@ -34,12 +34,13 @@ class PhysicalChannel:
     """
     __slots__ = ['_name', '__weakref__']
 
-    def __init__(self, name):
+    def __init__(self, name, interpreter):
         """
         Args:
             name (str): Specifies the name of the physical channel.
         """
         self._name = name
```
keer1111, line 42: update docstring in all places according to new addition of params

generated/nidaqmx/system/system.py
```diff
@@ -35,20 +36,30 @@ class System:
...
     def local():
         """
         nidaqmx.system.system.System: Represents the local DAQmx system.
         """
         return System()
 
+    @staticmethod
+    def remote(grpc_options):
+        """
+        nidaqmx.system.system.System: Represents the remote DAQmx system with interpreter.
```
keer1111, line 52:
```suggestion
        nidaqmx.system.system.System: Represents the remote DAQmx system.
```

keer1111 (changes requested):

generated/nidaqmx/system/system.py
```diff
@@ -35,20 +35,31 @@ class System:
...
         """
-        return System()
+        return System(System._grpc_options, System._interpreter)
+
+    @staticmethod
+    def remote():
+        """
+        nidaqmx.system.system.System._interpreter: Represents the interpreter instance.
+        """
+        return System._interpreter
```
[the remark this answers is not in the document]
MounikaBattu17: Updated to return System object accepting grpc_options.

generated/nidaqmx/task.py
```diff
@@ -63,7 +63,7 @@ class Task:
     Represents a DAQmx Task.
     """
 
-    def __init__(self, new_task_name=''):
+    def __init__(self, new_task_name='',*, grpc_options=None, interpreter=None):
```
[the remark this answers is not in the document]
MounikaBattu17: Removed.

src/codegen/templates/system/physical_channel.py.mako
MounikaBattu17: Updated docstring wherever applicable.

src/codegen/templates/scale.py.mako
```diff
@@ -24,12 +25,13 @@ class Scale:
     """
     __slots__ = ['_name', '__weakref__']
 
-    def __init__(self, name):
+    def __init__(self, name,*, grpc_options=None, interpreter=None):
```
bkeryan, line 28: `interpreter` is a private implementation detail. Instead of adding it to the public interface, let's use a simplified version of the `_TaskAlternateConstructor` approach.

```
class Scale:
    def __init__(self, name, *, grpc_options=None):
        self._name = name
```

In a private submodule:
```
class _ScaleAlternateConstructor(Scale):
    def __init__(self, name, interpreter):
        self._name = name
        self.__class__ = Scale
```

generated/nidaqmx/system/storage/_alternate_task_constructor.py
```diff
@@ -17,14 +17,17 @@ class _TaskAlternateConstructor(Task):
     derived class to 'overload' the constructor.
     """
 
-    def __init__(self, task_handle):
+    def __init__(self, task_handle,*, grpc_options=None, interpreter=None):
```
bkeryan, line 20: This class is used to construct with an existing task handle, so I think you should simplify it to:
```
def __init__(self, task_handle, interpreter):
```

And update all callers to pass in their interpreter.

src/codegen/templates/system/physical_channel.py.mako
```diff
@@ -34,12 +34,14 @@ class PhysicalChannel:
     """
     __slots__ = ['_name', '__weakref__']
 
-    def __init__(self, name):
+    def __init__(self, name, interpreter):
```
bkeryan, line 37: This is a public constructor. Adding a required `interpreter` parameter breaks compatibility with the example code in README.rst:
```
>>> phys_chan = device.ai_physical_chans['ai0']
>>> phys_chan
PhysicalChannel(name=Dev1/ai0)
>>> phys_chan == nidaqmx.system.PhysicalChannel('Dev1/ai0')
True
```

Instead, this class should take an optional, keyword-only `grpc_options`, and you should create a `_PhysicalChannelAlternateConstructor` to allow other code to inject an interpreter. See my suggestion on `scale.py.mako`.

src/handwritten/system/storage/persisted_channel.py
```diff
@@ -16,12 +16,14 @@ class PersistedChannel:
     """
     __slots__ = ['_name', '__weakref__']
 
-    def __init__(self, name):
+    def __init__(self, name, interpreter):
```
bkeryan, line 19: This is a public constructor. Adding a required interpreter parameter breaks compatibility with client code that directly constructs it like `nidaqmx.system.storage.PersistedChannel("MyChannel")`.

See my comments on `scale.py.mako`.

src/handwritten/system/storage/persisted_scale.py
```diff
@@ -17,12 +17,14 @@ class PersistedScale:
     """
     __slots__ = ['_name', '__weakref__']
 
-    def __init__(self, name):
+    def __init__(self, name, interpreter):
```
bkeryan, line 20: This is a public constructor. Adding a required interpreter parameter breaks compatibility with client code that directly constructs it like `nidaqmx.system.storage.PersistedScale("MyScale")`.

See my comments on `scale.py.mako`.

src/handwritten/system/storage/persisted_task.py
```diff
@@ -16,12 +16,14 @@ class PersistedTask:
     """
     __slots__ = ['_name', '__weakref__']
 
-    def __init__(self, name):
+    def __init__(self, name, interpreter):
```
bkeryan, line 19: This is a public constructor. Adding a required interpreter parameter breaks compatibility with client code that directly constructs it like `nidaqmx.system.storage.PersistedTask("MyTask")`.

See my comments on `scale.py.mako`.

src/handwritten/task.py
```diff
@@ -63,7 +63,7 @@ class Task:
     Represents a DAQmx Task.
     """
 
-    def __init__(self, new_task_name=''):
+    def __init__(self, new_task_name='', grpc_options=None):
```
bkeryan, line 66: Make `grpc_options` keyword-only.

src/handwritten/system/storage/_alternate_task_constructor.py
```diff
@@ -17,14 +17,17 @@ class _TaskAlternateConstructor(Task):
     derived class to 'overload' the constructor.
     """
 
-    def __init__(self, task_handle):
+    def __init__(self, task_handle,*, grpc_options=None, interpreter=None):
```
bkeryan, line 20: Formatting: leave a space before `*,`

src/handwritten/system/_collections/device_collection.py
```diff
@@ -15,6 +15,9 @@ class DeviceCollection(Sequence):
     
     This class defines methods that implements a container object.
     """
+    def __init__(self, interpreter):
+        self._interpreter = interpreter
```
bkeryan, line 19: @zhindes Are these collections public APIs? They live in private sub-modules, but they are returned from public classes.

Should we continue to allow constructing a `DeviceCollection()`?

src/handwritten/grpc_session_options.py
```diff
@@ -0,0 +1,67 @@
...
+    ):
+        r'''Collection of options that specifies session behaviors related to gRPC.
+        Creates and returns an object you can pass to a Session constructor.
+        Args:
+            grpc_channel (grpc.Channel): Specifies the channel to the NI gRPC Device Server.
+            session_name (str): User-specified name that identifies the driver session on the NI gRPC Device
+                Server. This is different from the resource name parameter many APIs take as a separate
+                parameter. Specifying a name makes it easy to share sessions across multiple gRPC clients.
+                You can use an empty string if you want to always initialize a new session on the server.
+                To attach to an existing session, you must specify the session name it was initialized with.
```
bkeryan, line 58: FYI, we will have to see whether we can support this.

bkeryan (changes requested):
nidaqmx-python's public API surface is much larger than it may appear. We need to preserve compatibility with client code that directly constructs `Device("Dev1")`, `PhysicalChannel("Dev1/ai0")`, etc.

We didn't know about `_TaskAlternateConstructor` when writing/reviewing the high level design, but I think this is likely to be an important pattern for injecting private implementation details.

generated/nidaqmx/system/storage/_alternate_task_constructor.py
MounikaBattu17: Updated the function signature as suggested. Also passed in `interpreter` from the caller `PersistedTask`.

src/codegen/templates/scale.py.mako
MounikaBattu17: Defining `_ScaleAlternateConstructor` in the same submodule as `Scale` sounds good. Shall we proceed on same?

src/handwritten/system/storage/_alternate_task_constructor.py
MounikaBattu17: Fixed formatting in all applicable places.

MounikaBattu17:
@bkeryan 
1) As suggested I have implemented alternate constructor in the same modules of these classes (`Device`, `Scale`, `PhysicalChannel`, `PersistedChannel`, `PersistedScale`, `PersistedTask`). Kindly review.
2) Also, can we move `_TaskAlternateConstructor` into same module as `Task` to maintain consistency.
3) Updating the callers will be tracked in _[Task 2351360](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351360): Update all System related functions with C calls with interpreter usage - Autogenerated_ and _[Task 2366649](https://dev.azure.com/ni/DevCentral/_workitems/edit/2366649): Update all System related functions with C calls with interpreter usage - Handwritten_. Is that fine?

bkeryan:
> @bkeryan
> 
> 1. As suggested I have implemented alternate constructor in the same modules of these classes (`Device`, `Scale`, `PhysicalChannel`, `PersistedChannel`, `PersistedScale`, `PersistedTask`). Kindly review.

Done

> 2. Also, can we move `_TaskAlternateConstructor` into same module as `Task` to maintain consistency?

Yes. Please update the imports:

```
src\handwritten\system\storage\persisted_task.py
154:        from nidaqmx.system.storage._alternate_task_constructor import (

generated\nidaqmx\system\storage\persisted_task.py
154:        from nidaqmx.system.storage._alternate_task_constructor import (
```

> 3. Updating the callers will be tracked in _[Task 2351360](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351360): Update all System related functions with C calls with interpreter usage - Autogenerated_ and _[Task 2366649](https://dev.azure.com/ni/DevCentral/_workitems/edit/2366649): Update all System related functions with C calls with interpreter usage - Handwritten_. Is that fine?

https://github.com/ni/nidaqmx-python/pull/283 broke the Task class and most of the tests. Please try to restore everything to a working state sooner, rather than later.

Splitting up the updates into separate PRs is okay with me, as long as it doesn't drag out for weeks.

bkeryan (changes requested):

src/handwritten/grpc_session_options.py
bkeryan: I discussed this with @zhindes and we are going to use a different GUID. I used GuidGen to generate one: `'90EDC55C-DE8B-4894-A28C-223BD7171383'`

image

src/codegen/templates/scale.py.mako
```diff
@@ -398,3 +399,29 @@ ${property_template.script_property(attribute)}\
...
+
+    Since we want the user to create a Scale simply by instantiating a
+    Scale object, thus, the Scale object's constructor has a DAQmx Create
+    Scale call.
+
+    Instantiating a Scale object from a Scale with passed in interpreter, 
+    requires that we either change the original constructor's prototype 
+    and add a parameter, or that we create this derived class to 'overload' 
+    the constructor.
+    """
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the docstrings in all alternate constructors.

src/codegen/templates/scale.py.mako
```diff
@@ -398,3 +399,29 @@ ${property_template.script_property(attribute)}\
...
+    def __init__(self, name, interpreter):
+        """
+        Args:
+            name: Specifies the name of the Scale.
+            interpreter: Specifies the interpreter instance.
+            
+        """
+        self._name = name
+        self._interpreter = utils._select_interpreter(interpreter)
+        self.__class__ = Scale
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the suggestion.

src/codegen/templates/scale.py.mako
```diff
@@ -25,13 +25,14 @@ class Scale:
     """
     __slots__ = ['_name', '__weakref__']
 
-    def __init__(self, name,*, grpc_options=None, interpreter=None):
+    def __init__(self, name, *, grpc_options=None):
         """
         Args:
             name (str): Specifies the name of the scale to create.
+            grpc_options(GrpcSessionOptions): Specifies the gRPC session options.
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the docstrig of grpc_options to this `grpc_options (Optional[GrpcSessionOptions]): Specifies the gRPC session options`.

src/handwritten/grpc_session_options.py
MounikaBattu17: Thank you Brad. Used the given api constant and key.

keer1111:
@bkeryan , we will prioritize fixing the test case crashes parallelly and complete it(will not span more than 3 days). The current upcoming PRs will do some fixes, on top of it we will have to update the missing interpreter usages.

src/handwritten/system/_collections/device_collection.py
zhindes: I'm fine if these break compatibility as long as the public classes usage is maintained.

bkeryan: @MounikaBattu17 Let's make sure the documentation is clear about this:
- Add a docstring to the collection classes' `__init__` methods saying do not construct this object directly; instead, call `nidaqmx.system.System.local().devices`, etc.
- Add a note to CHANGELOG.md saying that the signature of the collection classes' `__init__` methods have changed and any code that constructs these objects directly should be updated to get them from the appropriate object (e.g. `System.devices`, `Task.channels`).

src/handwritten/grpc_session_options.py
bkeryan: We discussed this in our conference call today. 
- grpc-device currently uses a single name for both the grpc-device session registration and the DAQmx task. 
- In the future, we could conceivably add a parameter to allow specifying these independently, but we aren't doing that now.
- Let's keep the `GrpcSessionOptions.session_name` field, but validate in `Task.__init__` that it is set to "" or equal to the task name.

bkeryan (changes requested):

src/codegen/templates/scale.py.mako
```diff
@@ -396,3 +399,25 @@ ${property_template.script_property(attribute)}\
...
+
+    def __init__(self, name, interpreter):
+        """
+        Args:
+            name: Specifies the name of the Scale.
+            interpreter: Specifies the interpreter instance.
+            
+        """
+        self._name = name
+        self._interpreter = utils._select_interpreter(interpreter)
```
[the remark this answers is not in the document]
MounikaBattu17: Sure Brad. I have removed `_select_interpreter` from all of the alternate constructors and directly assigned the `interpreter`. Thanks for pointing this out. I too missed to notice this.

src/handwritten/grpc_session_options.py
MounikaBattu17: Sure. Added the validation for `grpc_options.session_name` in `Task.__init__` and raised an `DaqmxError` with error code `DAQmxErrors.UNKNOWN`. Please provide your thoughts on this.

src/handwritten/system/_collections/device_collection.py
MounikaBattu17: Thanks for the detailed comment. Updated the docstrings for all collection classes and added the change in `CHANGELOG.md` in a new version `0.7.2`. Please specify if this version is correct or not.

src/codegen/templates/_task_modules/ai_channel_collection.py.mako
```diff
@@ -26,6 +26,9 @@ class AIChannelCollection(ChannelCollection):
     Contains the collection of analog input channels for a DAQmx Task.
     """
     def __init__(self, task_handle, interpreter):
+        """
+        Do not construct this object directly; instead, call nidaqmx.Task.ai_channels.
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the suggestion.

src/handwritten/system/_collections/physical_channel_collection.py
```diff
@@ -16,6 +16,9 @@ class PhysicalChannelCollection(Sequence):
     This class defines methods that implements a container object.
     """
     def __init__(self, device_name, interpreter):
+        """
+        Do not construct this object directly; instead, call appropriate physical channels from nidaqmx.system.Device.
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the suggestion.

src/handwritten/_task_modules/channel_collection.py
```diff
@@ -17,6 +17,9 @@ class ChannelCollection(Sequence):
     This class defines methods that implements a container object.
     """
     def __init__(self, task_handle, interpreter):
+        """
+        Do not construct this object directly; instead, call appropriate channels from nidaqmx.Task.
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the suggestion.

src/handwritten/task.py
```diff
@@ -78,6 +78,15 @@ def __init__(self, new_task_name='', *, grpc_options=None):
                 results in an error.
         """
         self._handle = lib_importer.task_handle(0)
+
+        if grpc_options and not (
+            grpc_options.session_name == "" or grpc_options.session_name == new_task_name
+        ):
+            raise DaqError(
+                'The given session name is invalid , it should either match with default value or task name',
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the suggestion.

CHANGELOG.md
```diff
@@ -18,6 +18,15 @@
 
 All notable changes to this project will be documented in this file.
 
+## 0.7.2
```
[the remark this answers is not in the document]
MounikaBattu17: Thanks for the information. Added in 0.7.1. Is that fine?

src/handwritten/grpc_session_options.py
bkeryan: Looks good

bkeryan (approved):

CHANGELOG.md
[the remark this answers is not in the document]
bkeryan: Yes

keer1111 (approved):

PR #295, opened 2023-04-21
Library Interpreter | PR0 | Move Read/Write functions to Library interpreter

MounikaBattu17:

-  Moved read/write functions to `_library_interpreter.py`.
- Some logic changes were specific to read/write functions, so to differentiate them, added `functions_addon.py` and overrided `python_codegen_method` to `Custon_Code_Read_Write`.
- Merged `functions` and `functions_override` in `metadata/_init_.py`.
- Handwritten 4 read/write functions
     - `read_power_i16`
     - `read_power_f64`
     - `read_raw`
     - `write_raw`

Implements [Task 2351399](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351399): Generate the Library Interpreter inheriting the base - Handling Error methods for Read/Write Case


zhindes (approved):

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -63,3 +63,95 @@ class LibraryInterpreter(BaseInterpreter):
         raise NotImplementedError
 %endif
 %endfor
+
+    def read_power_i16(
```
bkeryan, line 67: Why are these hand-written? Please add a comment explaining the reason. You can use a Mako comment `##` so it isn't included in the rendered output file.

src/codegen/utilities/function_helpers.py
```diff
@@ -167,6 +176,14 @@ def get_arguments_type(functions_metadata):
                     size_param_info = tuple()
             argtypes.append("ctypes.c_uint")
             size_param_info = param, (len(argtypes) - 1)
+
+    # lib_importer.task_handle is inserted at start index of list .
+    # The argument type for 'reserved' parameter is inserted at
+    # end of list for all read/write functions.
+    if functions_metadata.python_codegen_method == "CustomCode_Read_Write":
+        argtypes.insert(0, "lib_importer.task_handle")
+        argtypes.insert(len(argtypes), "ctypes.POINTER(c_bool32)")
```
bkeryan, line 185: The metadata describes these reserved parameters:

```
            {
                'ctypes_data_type': 'c_bool32',
                'direction': 'in',
                'hardcoded_value': 'nullptr',
                'include_in_proto': False,
                'is_optional_in_python': False,
                'name': 'reserved',
                'pointer': True,
                'python_data_type': 'bool',
                'python_description': '',
                'python_type_annotation': 'bool',
                'type': 'bool32'
            }
```

bkeryan (changes requested):

src/codegen/utilities/function_helpers.py
```diff
@@ -34,6 +34,8 @@
     "SetDigitalPullUpPullDownStates",
 ]
 
+CHANGE_FILLMODE_TYPE_FUNCTIONS = ["read_analog_f64"]
```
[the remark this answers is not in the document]
MounikaBattu17: Removed the replacement of ctype.

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -63,3 +63,95 @@ class LibraryInterpreter(BaseInterpreter):
...
+
+        error_code = cfunc(
+            task, num_samps_per_chan, timeout, fill_mode,
+            read_voltage_array, read_current_array, read_voltage_array.size,
+            ctypes.byref(samps_per_chan_read), None)
+        check_for_error(error_code, samps_per_chan_read=samps_per_chan_read.value)
+
+        return samps_per_chan_read.value
+
+    def read_power_f64(
```
bkeryan, line 92: x

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -63,3 +63,95 @@ class LibraryInterpreter(BaseInterpreter):
...
+            task, read_voltage_array, read_current_array, num_samps_per_chan,
+            timeout, fill_mode):
```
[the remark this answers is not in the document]
bkeryan: This is not resolved.

src/codegen/templates/_library_interpreter.py.mako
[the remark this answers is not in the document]
bkeryan: This is not resolved.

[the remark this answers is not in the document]
bkeryan: This is not resolved.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2619,26 +2619,24 @@ def load_task(self, session_name):
         check_for_error(error_code)
 
     def read_analog_f64(
-            self, task, num_samps_per_chan, timeout, fill_mode,
-            array_size_in_samps):
-        read_array = numpy.zeros(array_size_in_samps, dtype=numpy.numpy.float64)
+            self, task, num_samps_per_chan, timeout, fill_mode, read_array):
```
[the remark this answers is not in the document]
MounikaBattu17: The `functions.py` metadata from Scarpigen defines `fill_mode` as non-optional parameter (Refer metadata [here](https://github.com/ni/nidaqmx-python/blob/users/mounika/move-read-write-func-libint/src/codegen/metadata/functions.py#L17098)) for codegen. That's why signature order is changed.
image

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2619,26 +2619,24 @@ def load_task(self, session_name):
...
 
         error_code = cfunc(
-            task, num_samps_per_chan, timeout, fill_mode,
-            ctypes.byref(read_array), array_size_in_samps,
-            ctypes.byref(samps_per_chan_read))
-        check_for_error(error_code)
+            task, num_samps_per_chan, timeout, fill_mode, read_array,
+            read_array.size, ctypes.byref(samps_per_chan_read), None)
+        check_for_error(error_code, samps_per_chan_read=samps_per_chan_read.value)
         return read_array.tolist(), samps_per_chan_read.value
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the read functions to return directly without any type conversions.

src/codegen/templates/_library_interpreter.py.mako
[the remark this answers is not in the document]
MounikaBattu17: Renamed to `read_power_binary_i16`.

src/codegen/templates/_library_interpreter.py.mako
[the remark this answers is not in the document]
MounikaBattu17: Updated to correct function signature and the parameters are ordered as same as other read/write to maintain consistency.

src/codegen/utilities/function_helpers.py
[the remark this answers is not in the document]
MounikaBattu17: Updated to `append()`.

src/codegen/templates/_library_interpreter.py.mako
MounikaBattu17: The reasons for the functions to be handwritten is that basically we don't have or have incomplete/incorrect data for these functions.
1) `ReadPowerBinaryI16/ReadPowerF64` - there is no data in `daqmxAPISharp.json` file
2) ReadRaw 
   - the datatype for readArray is generic in `daqmxAPISharp.json`
```
               {
               "dotNetDataType": "object[]", 
               "direction": "out", 
               "dotNetModifier": "", 
               "internalDescription": "", 
               "name": "readArray", 
               "cviBufferSizeIsExplicit": true, 
               "default": null, 
               "cviDataType": "object[]", 
               "internalType": "object[]"
            },
```
 - Also, expected c call is to pass `read_array.nbytes` but while autogenerating we get `read_array.size`
    
image

3) WriteRaw - the datatype of writeArray is incorrect in `daqmxAPISharp.json` 
```
            {
               "dotNetDataType": "object", 
               "direction": "in", 
               "dotNetModifier": "", 
               "internalDescription": "", 
               "name": "writeArray", 
               "cviBufferSizeIsExplicit": false, 
               "default": null, 
               "cviDataType": "object", 
               "internalType": "object"
            }, 
```

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -2243,58 +2244,62 @@ def read_analog_scalar_f64(self, task, timeout):
...
+            self, task, num_samps_per_chan, timeout, fill_mode, read_array):
         response = self._invoke(
             self._client.ReadBinaryU16,
             grpc_types.ReadBinaryU16Request(
                 fill_mode_raw=fill_mode,
-                num_samps_per_chan=num_samps_per_chan, task=task,
-                timeout=timeout))
+                num_samps_per_chan=num_samps_per_chan, read_array=read_array,
+                task=task, timeout=timeout))
         return response.read_array, response.samps_per_chan_read
```
bkeryan, line 2275: This introduces new errors into GrpcStubInterpreter: `read_array` does not exist in the request message.

The right output is something like
```
            grpc_types.ReadBinaryU16Request(
                fill_mode_raw=fill_mode,
                num_samps_per_chan=num_samps_per_chan,
                task=task, timeout=timeout))
        read_array[:] = response.read_array
        return response.samps_per_chan_read
```

@Sakthi-SM Do you intend to address gRPC read/write in a separate "gRPC Interpreter" PR?

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2611,25 +2611,26 @@ def load_task(self, session_name):
...
+                        ctypes.POINTER(ctypes.c_int), ctypes.POINTER(c_bool32)]
 
         error_code = cfunc(
             task, num_samps_per_chan, timeout, fill_mode, read_array,
-            ctypes.byref(samps_per_chan_read))
-        check_for_error(error_code)
-        return read_array.tolist(), samps_per_chan_read.value
+            read_array.size, ctypes.byref(samps_per_chan_read), None)
+        check_for_error(error_code, samps_per_chan_read=samps_per_chan_read.value)
+        return read_array, samps_per_chan_read.value
```
bkeryan, line 2633: `read_array` is a mutable in/out parameter, so there is no need to return it.

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -63,3 +63,101 @@ class LibraryInterpreter(BaseInterpreter):
         raise NotImplementedError
 %endif
 %endfor
+
+    ## The metadata for 'read_power_binary_i16' function is not available in daqmxAPSharp.json file.
```
bkeryan, line 67: Spelling: daqmxAPISharp.json

There are at least 4 instances of this misspelling.

src/codegen/templates/_library_interpreter.py.mako
bkeryan: Ok, please create a technical debt item for these.

src/codegen/utilities/function_helpers.py
bkeryan: Yes, a lot of the complexity here is due to mapping between three different versions of the DAQmx API: C, gRPC, and Python. They don't all have the same parameters.

If the metadata's reserved parameter definitions are unavailable because we filtered out non-Python parameters, then okay, leave the special case and move on.

generated/nidaqmx/_library_interpreter.py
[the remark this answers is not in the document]
bkeryan: Ok, I might not have been clear that it shouldn't return the array reference either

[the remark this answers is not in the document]
bkeryan: Nevermind, the hand-written functions did this too.

generated/nidaqmx/_library_interpreter.py
bkeryan: Nevermind, the hand-written functions did this too.

src/codegen/templates/_library_interpreter.py.mako
MounikaBattu17: Updated to correct spelling

generated/nidaqmx/_grpc_interpreter.py
Sakthi-SM: Yes Brad that will be addressed in a different PR.

src/codegen/templates/_library_interpreter.py.mako
MounikaBattu17: Created a [Technical Debt 2377787](https://dev.azure.com/ni/DevCentral/_workitems/edit/2377787): Autogenerate handwritten read/write methods in library interpreter to track this issue.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -34,6 +34,10 @@
     "SetFirstSampClkWhen",
     "SetStartTrigTrigWhen",
     "SetSyncPulseTimeWhen",
+    "ReadPowerBinaryI16",
+    "ReadPowerF64",
+    "ReadRaw",
+    "WriteRaw",
```
bkeryan, line 40: This breaks GrpcStubInterpreter. If the method is defined in BaseInterpreter but not GrpcStubInterpreter, then you can't instantiate GrpcStubInterpreter.

Here's how to fix it:
- Move these to LIBRARY_INTERPRETER_IGNORED_FUNCTIONS
- Remove the hardcoded versions from _base_interpreter.py.mako

bkeryan (changes requested):

src/codegen/utilities/interpreter_helpers.py
MounikaBattu17: Sure Brad. 

- I have added those functions in `LIBRARY_INTERPRETER_IGNORED_FUNCTIONS` and skip their generation in `_library_interpreter.py.mako`. 
- Also removed the hardcoded versions from `_base_interpreter.py.mako` .
- Changed `python_codegen_method` of those functions from '`no`' to '`Custom_Code_Read_Write`.
please say if this is fine or not?

MounikaBattu17: Yes, Completely agree on your thoughts. All these parameters changes will be handled in the upcoming PRs. Thanks for pointing this out.

src/codegen/utilities/function_helpers.py
MounikaBattu17: Yes, Completely agree on your thoughts. All these parameters changes will be handled in the upcoming PRs. Thanks for pointing this out.

src/codegen/utilities/interpreter_helpers.py
bkeryan: Looks good

bkeryan (approved):

PR #285, opened 2023-04-18
gRPC interpreter | PR2 | Generate gRPC interpreter class

Sakthi-SM:
**What does this Pull Request accomplish?**

* Adds the necessary `mako` template for generating gRPC interperter

* Generates `_grpc_interpreter.py`

* Updated the necessary helper functions that required for generating the interpreter.

**Why should this Pull Request be merged?**
This implements [Task 2351313](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351313): Generate gRPC Interpreter using mako

PR #297, opened 2023-04-24
Library Interpreter | Generate functions that uses the Cdecl calling convention

subash-suresh:
# What does this PR accomplish?
This PR generates the functions that is using the cdecl calling convention.

# Why should this PR be merged?
- removed the check in `_library_interpreter.py.mako` that restricted the generation of functions using the cdecl calling convention
- Added a new template called `exec_cdecl_c_function_call.py` that handles the cdecl calls that uses varargs implemented using the exec style. 
- The required helper functions for handling the above scenario has been added.

# What testing has been done?
The functions generated have been manually verified..


generated/nidaqmx/_library_interpreter.py
```diff
@@ -2232,11 +2232,52 @@ def get_ai_chan_cal_exp_date(self, task, channel_name):
 
     def get_analog_power_up_states(
             self, device_name, channel_name, channel_type):
-        raise NotImplementedError
+        state = []
+
+        args = [device_name]
+        argtypes = [ctypes_byte_str]
+
+        for index in range(len(channel_name)):
```
zhindes, line 2240: This isn't right. I gave some specific details in #298 . I think y'all know this, though, so I won't comment on all these issues.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2232,11 +2232,52 @@ def get_ai_chan_cal_exp_date(self, task, channel_name):
...
+            argtypes.append(ctypes.c_int32)
+            
+        args.append(None)
+
+        cfunc = lib_importer.cdll.DAQmxGetAnalogPowerUpStates
+        with cfunc.arglock:
+            cfunc.argtypes = argtypes
+            error_code = cfunc(*args)
+        check_for_error(error_code)
+        return state
```
zhindes, line 2260: Ditto here, this isn't right... this needs to be a constructed list of objects

zhindes (changes requested):

keer1111 (approved):

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -83,10 +99,9 @@ def generate_interpreter_function_call_args(function_metadata):
                     function_call_args.append("temp_size")
             else:
                 function_call_args.append(f"ctypes.byref({param.parameter_name})")
-
+
```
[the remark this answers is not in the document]
subash-suresh: fixed lint errors

src/codegen/functions/function.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, function_name, function_metadata):
                 function_parameter = FunctionParameter(parameter)
                 self._base_parameters.append(function_parameter)
                 if parameter.get("use_in_python_api") is not False:
-                    if parameter["name"] != "task" and "python_data_type" in parameter:
+                    if parameter["name"] != "task" and parameter.get("python_data_type", None) is not None:
```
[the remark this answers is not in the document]
subash-suresh: removed the explicit declaration of None

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -122,6 +140,41 @@ def get_instantiation_lines_for_output(func):
...
+        argument_defininion_lines.append(f"argtypes.append({argtype})")
+        argument_defininion_lines.append("")
+    return argument_defininion_lines
+
+
+def get_varargs_parameters(func):
+    varargs_parameters = []
+    if func.function_name in CDLL_EXEC_STYLE_VARARGS_FUNCTIONS:
+        varargs_parameters = func.parameters
+        del varargs_parameters[0]
```
bkeryan, line 174: The varargs parameters are described in the metadata: https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/functions.py#L12740
- The C parameters have `repeating_argument`
- The gRPC parameters have `repeated_var_args`
- The metadata doesn't really describe the Python parameters

Suggestion:
- Add an `is_repeating_argument` property to the `Parameter` class. Edit: @Sakthi-SM has already done that in https://github.com/ni/nidaqmx-python/pull/285/
- Change this code to `return [p for p in func.parameters if p.is_repeating_argument]`
- Elsewhere, replace `if func.function_name in CDLL_EXEC_STYLE_VARARGS_FUNCTIONS` with `if any(get_varargs_parameters(func))`

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -122,6 +140,41 @@ def get_instantiation_lines_for_output(func):
...
+        for param in func.output_parameters:
+            instantiation_lines.append(f"{param.parameter_name}_element = {param.ctypes_data_type}")
+            instantiation_lines.append(
+                f"{param.parameter_name}.append({param.parameter_name}_element)"
+            )
+    return instantiation_lines
+
+
+def get_argument_definition_lines_for_varargs(varargs_params):
+    argument_defininion_lines = []
```
bkeryan, line 156: Spelling: "definition"

bkeryan (changes requested):

src/codegen/utilities/interpreter_helpers.py
subash-suresh: updated

src/codegen/utilities/interpreter_helpers.py
subash-suresh: updated as per suggestion

generated/nidaqmx/_library_interpreter.py
subash-suresh: I have provided the information on the approach taken in #298 - https://github.com/ni/nidaqmx-python/pull/298#discussion_r1176130587

generated/nidaqmx/_library_interpreter.py
subash-suresh: I have updated the code to construct lists before returning.

generated/nidaqmx/_library_interpreter.py
zhindes: OK, that makes sense.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2230,11 +2230,49 @@ def get_ai_chan_cal_exp_date(self, task, channel_name):
...
+            argtypes.append(ctypes.c_int32)
+            
+        args.append(None)
+
+        cfunc = lib_importer.cdll.DAQmxGetAnalogPowerUpStates
+        with cfunc.arglock:
+            cfunc.argtypes = argtypes
+            error_code = cfunc(*args)
+        check_for_error(error_code)
+        return state.tolist()
```
zhindes, line 2258: Isn't this returning a list of `ctypes.c_double`? Should we be unwrapping the values in there instead?

zhindes (changes requested):

generated/nidaqmx/_library_interpreter.py
bkeryan: Yes, this is mixing types and values.

`ctypes.c_double` is a type. You should put it in `argtypes`.

`ctypes.c_double()` is an instance of `ctypes.c_double`. You should put it in args using `ctypes.byref()`.

 `ctypes.c_double().value` is a Python float. You should return a list of these from the function.

bkeryan (changes requested):

src/codegen/templates/library_interpreter/exec_cdecl_c_function_call.py.mako
```diff
@@ -0,0 +1,37 @@
...
+            ${instantiation_line}
+        %endfor
+    %endif
+
+    %if len(argument_definition_lines) > 0:
+        %for argument_definition_line in argument_definition_lines:
+            ${argument_definition_line}
+        %endfor
+    %endif
+        args.append(None)
```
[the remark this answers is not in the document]
subash-suresh: The [existing implementation](https://github.com/ni/nidaqmx-python/blob/f64f626aaf349fd44bddd7915c88be8ce1457f14/generated/nidaqmx/system/system.py#L274) of these types of methods does not define an argument for these, and hence I followed the same here. so is argtypes required for this case where we provide the None parameter as a means for varargs termination?.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -194,7 +236,9 @@ def get_return_values(func):
     """Gets the values to add to return statement of the function."""
     return_values = []
     for param in get_output_parameters(func):
-        if param.ctypes_data_type == "ctypes.c_char_p":
+        if any(get_varargs_parameters(func)):
+            return_values.append(f"{param.parameter_name}.tolist()")
```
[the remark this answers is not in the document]
subash-suresh: updated as per suggestions

generated/nidaqmx/_library_interpreter.py
subash-suresh: Yes and here the `state` variable is a list, hence i am trying to return a list in this implementation.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -121,6 +138,32 @@ def get_instantiation_lines_for_output(func):
             instantiation_lines.append(f"{param.parameter_name} = {param.ctypes_data_type}()")
     return instantiation_lines
 
+def get_instantiation_lines_for_varargs(func):
+    instantiation_lines = []
+    if func.function_name in CDLL_EXEC_STYLE_VARARGS_FUNCTIONS:
+        for param in func.output_parameters:
+            instantiation_lines.append(f"{param.parameter_name}_element = {param.ctypes_data_type}")
+            instantiation_lines.append(f"{param.parameter_name}.append({param.parameter_name}_element)")
```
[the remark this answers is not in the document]
subash-suresh: This is implementation has been updated

src/codegen/templates/library_interpreter/exec_cdecl_c_function_call.py.mako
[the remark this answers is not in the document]
bkeryan: I see. We don't **really** need to add varargs to argtypes at all. We only need to add the non-variadic arguments: https://docs.python.org/3/library/ctypes.html#calling-variadic-functions

However, we can't just remove the varargs, because ctypes converts `str` to `wchar_t*` by default: https://docs.python.org/3/library/ctypes.html#calling-functions

> None, integers, bytes objects and (unicode) strings are the only native Python objects that can directly be used as parameters in these function calls. None is passed as a C NULL pointer, bytes objects and strings are passed as pointer to the memory block that contains their data (char* or wchar_t*). Python integers are passed as the platforms default C int type, their value is masked to fit into the C type.

In my opinion, if we are going to add the varargs to argtypes, we should add all of the varargs to argtypes.

generated/nidaqmx/_library_interpreter.py
zhindes: Right, but `state` is currently a list of types, not objects of a type. Even if it were objects of a type, it is the wrong type. You want something like this:

```python
state_list = []
for chans:
   state = ctypes.c_double() # note ctor call
   state_list.append(state)
   args.append(ctypes.byref(state))

return [state.value for state in state_list]
```

zhindes (changes requested):

generated/nidaqmx/_library_interpreter.py
subash-suresh: Thanks for the clarification, i have updated as per suggestions

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2232,11 +2232,50 @@ def get_ai_chan_cal_exp_date(self, task, channel_name):
...
     def get_analog_power_up_states(
             self, device_name, channel_name, channel_type):
-        raise NotImplementedError
+        state = []
+
+        args = [device_name]
+        argtypes = [ctypes_byte_str]
+
+        for index in range(len(channel_name)):
+            state_element = ctypes.c_double
```
zhindes, line 2241: you need to call the ctor here

```suggestion
            state_element = ctypes.c_double()
```

subash-suresh: I have updated this as per suggestions

bkeryan (approved):
This looks fine, except for the potential locking problem I just discovered.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2261,61 +2300,336 @@ def get_auto_configured_cdaq_sync_connections(self):
...
+        value = ctypes.c_int32()
+
+        cfunc = lib_importer.cdll.DAQmxGetChanAttribute
+        with cfunc.arglock:
+            cfunc.argtypes = [
+                lib_importer.task_handle, ctypes_byte_str, ctypes.c_int32,
+                ctypes.POINTER(ctypes.c_int32)]
+
+        error_code = cfunc(
+            task, channel, attribute, ctypes.byref(value))
```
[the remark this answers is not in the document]
subash-suresh: I have created a tech debt item to track this, as mentioned we will circle back to this once we have all the other implementations done. 
[Technical Debt 2377746](https://ni.visualstudio.com/DevCentral/_workitems/edit/2377746): Update the library interpreter c function calls to be thread-safe.

PR #298, opened 2023-04-24
[PR0] [Properties] Update property function calls for get, set, reset to use interpreter for function calls

samchris007:
This PR updates property function calls for get, set, reset to use interpreter for function calls.

### Why should this PR be merged?
Update autogenerated get, set, reset properties to use generic properties from Library Interpreter
- `PhysicalChannel` doesn't have set properties, hence those properties' function calls are not updated to use interpreter.
- In Library interpreter, `Watchdog` related generic property functions have an additional parameter named `lines`.
  - An empty string is passed as value for such functions.  

Some of the tests are currently failing. Will be resolving those test failure either as part of this PR or in this PR https://github.com/ni/nidaqmx-python/pull/297


generated/nidaqmx/_library_interpreter.py
```diff
@@ -2228,11 +2228,52 @@ def get_ai_chan_cal_exp_date(self, task, channel_name):
 
     def get_analog_power_up_states(
             self, device_name, channel_name, channel_type):
-        raise NotImplementedError
+        state = []
+
+        args = [device_name]
+        argtypes = [ctypes_byte_str]
+
+        for index in range(None):
```
zhindes, line 2236: I am guessing you know about this since you mention that fixes will be later, but FYI this isn't right. In mainline, we create a `Device` object and iteration the `ao_physical_chans`

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2228,11 +2228,52 @@ def get_ai_chan_cal_exp_date(self, task, channel_name):
...
+            argtypes.append(ctypes.c_int32)
+            
+        args.append(None)
+
+        cfunc = lib_importer.cdll.DAQmxGetAnalogPowerUpStates
+        with cfunc.arglock:
+            cfunc.argtypes = argtypes
+            error_code = cfunc(*args)
+        check_for_error(error_code)
+        return state.value
```
zhindes, line 2256: You gotta create a list of `AOPowerUpState` objects and return that.

generated/nidaqmx/_library_interpreter.py
zhindes: Lots of issues like this, just FYI

generated/nidaqmx/_task_modules/channels/channel.py
```diff
@@ -198,102 +198,43 @@ def chan_type(self):
...
-                        lib_importer.task_handle, ctypes_byte_str,
-                        ctypes.POINTER(ctypes.c_int)]
 
-        error_code = cfunc(
-            self._handle, self._name, ctypes.byref(val))
-        check_for_error(error_code)
-
-        return ChannelType(val.value)
+        val = self._interpreter.get_chan_attribute_int32(
+                self._handle, self._name, 6271)
```
zhindes, line 204: We most commonly refer to attribute IDs by their hex value, so 0x187F here. Do you mind updating that? Bonus points if we created constants for all of these 😲

zhindes: Constants ala the C header

```c
#define DAQmx_ChanType                                                   0x187F // Indicates the type of the virtual channel.
```

zhindes (changes requested):

bkeryan (changes requested):

generated/nidaqmx/_library_interpreter.py
subash-suresh: I have provided details for this implementation in the [above comment](https://github.com/ni/nidaqmx-python/pull/298#discussion_r1176130587)

generated/nidaqmx/_library_interpreter.py
zhindes: Ahh, I like it. That makes sense!

generated/nidaqmx/_task_modules/channels/channel.py
samchris007: Sure. Updated.

src/codegen/templates/property_template.py.mako
```diff
@@ -3,12 +3,16 @@
...
     %elif attribute.access == "write":
 ${property_empty_getter_template.script_empty_property_getter(attribute)}
 ${property_setter_template.script_property_setter(attribute)}
-    %elif attribute.access == "read-write":
+    %elif attribute.access == "read-write" and attribute.python_class_name != "PhysicalChannel":
+${property_getter_template.script_property_getter(attribute)}
+${property_interpretercall_setter_template.script_property_setter(attribute)}
+    %elif attribute.access == "read-write" and attribute.python_class_name == "PhysicalChannel":
 ${property_getter_template.script_property_getter(attribute)}
 ${property_setter_template.script_property_setter(attribute)}
```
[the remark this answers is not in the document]
keer1111: Resolving fully - update NI-DAQmx & grpc-device is tracked here - [Technical Debt 2375679](https://dev.azure.com/ni/DevCentral/_workitems/edit/2375679): Update NI-DAQmx and grpc-device to support Physical channel Setter & Resetter

src/codegen/templates/property_deleter_template.py.mako
```diff
@@ -1,30 +1,21 @@
...
             function_call_args.append(handle_parameter.accessor)
+        if attribute.python_class_name == "Watchdog":
+            function_call_args.append("\"\"")
+        function_call_args.append(str(attribute.id))
     %>\
-        error_code = cfunc(
-            ${', '.join(function_call_args) | wrap(initial_indent=12)})
-        check_for_error(error_code)
+        self._interpreter.reset_${generic_attribute_func}(
+                ${', '.join(function_call_args) | wrap(initial_indent=12)})
```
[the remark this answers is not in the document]
bkeryan: The wrap function doesn't affect the indentation of the first line. You need to update the whitespace to use 12 spaces.

generated/nidaqmx/system/physical_channel.py
```diff
@@ -150,44 +98,26 @@ def ai_power_control_enable(self, val):
...
-            with cfunc.arglock:
-                if cfunc.argtypes is None:
-                    cfunc.argtypes = [
-                        ctypes_byte_str]
-
-        error_code = cfunc(
-            self._name)
-        check_for_error(error_code)
+        self._interpreter.reset_physical_chan_attribute(
+                self._name, 0x316d)
```
bkeryan, line 108: The interpreter does not have a `reset_physical_chan_attribute` method.

src/codegen/templates/property_template.py.mako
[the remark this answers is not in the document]
bkeryan: @keer1111 Thanks for creating a work item about the NI-DAQmx and grpc-device work.

@samchris007 I thought about it some more. I still prefer **Add attribute-independent functions to the interpreter interface**, but I think **Ignore interpreter and always call the DLL** would be acceptable with some minor changes:
- Rename the old `property_setter_template` to `property_legacy_setter_template` to reflect that this is a special case.
- Rename `property_interpretercall_setter_template` to `property_setter_template` to reflect that this is the common case.
- Fork the old `property_deleter_template` and name it `property_legacy_deleter_template`.
- Add an error check to `property_legacy_setter_template` and `property_legacy_deleter_template`:
```
if not isinstance(self._interpreter, LibraryInterpreter):
    raise NotImplementedError
```

src/codegen/templates/property_deleter_template.py.mako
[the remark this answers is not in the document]
samchris007: Sure. Updated.

src/codegen/templates/property_template.py.mako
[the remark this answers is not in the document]
samchris007: Sure. With **Add attribute-independent functions to the interpreter interface**, it is difficult to figure out which dll calls should be made for the generic attribute functions like`set_physical_chan_attribute_bool`, `reset_physical_chan_attribute` in Library Interpreter.
However, I have created a separate PR with **Ignore interpreter and always call the DLL** approach: https://github.com/ni/nidaqmx-python/pull/302.

generated/nidaqmx/system/physical_channel.py
samchris007: Sure. This has been handled in #302

zhindes (approved):

src/codegen/templates/property_deleter_template.py.mako
```diff
@@ -1,30 +1,21 @@
...
             function_call_args.append(handle_parameter.accessor)
+        if attribute.python_class_name == "Watchdog":
+            function_call_args.append("\"\"")
+        function_call_args.append(hex(attribute.id))
+
     %>\
-        error_code = cfunc(
-            ${', '.join(function_call_args) | wrap(initial_indent=12)})
-        check_for_error(error_code)
+        self._interpreter.reset_${generic_attribute_func}(${', '.join(function_call_args) | wrap(12)})
```
bkeryan, line 20: Using wrap(12) is only correct when the function call args start at column 12, because it will return up to 68 columns on the first line.

You know how a lot of the templates have these `%if (len(func.function_name) + len(parameter_signature)) > 68:` conditionals? This is why. 

If these updated get/set/reset templates always have short arguments, then I'd rather remove the call to wrap().

src/codegen/templates/property_template.py.mako
```diff
@@ -3,12 +3,16 @@
...
     %if attribute.access == "read":
 ${property_getter_template.script_property_getter(attribute)}
     %elif attribute.access == "write":
 ${property_empty_getter_template.script_empty_property_getter(attribute)}
 ${property_setter_template.script_property_setter(attribute)}
-    %elif attribute.access == "read-write":
+    %elif attribute.access == "read-write" and attribute.python_class_name != "PhysicalChannel":
+${property_getter_template.script_property_getter(attribute)}
+${property_interpretercall_setter_template.script_property_setter(attribute)}
+    %elif attribute.access == "read-write" and attribute.python_class_name == "PhysicalChannel":
```
bkeryan, line 15: Moving the special case above the normal case would simplify the conditionals.
```
%elif attribute.access == "read-write" and attribute.python_class_name == "PhysicalChannel":
%elif attribute.access == "read-write":
%endif
```

bkeryan (approved):
Approved with suggestions

src/codegen/templates/property_deleter_template.py.mako
samchris007: Ok. Removed `wrap()` from get, set property templates.

src/codegen/templates/property_template.py.mako
samchris007: Updated.

samchris007:
@bkeryan , @zhindes 
FYI, there are numerous tests that are failing right now. Each test's root cause has been found. Our team is currently planning to address that before merging this PR into master.

PR #300, opened 2023-04-25
gRPC interpreter | PR3 | Handle initialize methods in the interpreters

Sakthi-SM:

As per the suggestion received in https://github.com/ni/nidaqmx-python/pull/285#discussion_r1171620709 the following changes have been made:
* Updated the `grpc_interpreter` to send `ni-api-key` as metadata for these `init` methods.
* Updated the `grpc_interpreter` to send `intialization_behaviour` for these methods.
* Updated the `library_interpreter` to send `is_session_initialized` to `True` in all the scenarios.

NOTE: The comment also suggest updating the `task.py` to store this session initialized boolean and use it while closing the session. But this will be covered in a different PR. [Task 2375614](https://dev.azure.com/ni/DevCentral/_workitems/edit/2375614): Update the task.py to store and use the session initialized boolean
 
This implements [Task 2373857](https://dev.azure.com/ni/DevCentral/_workitems/edit/2373857): Update the interpreters to handle initialization methods


generated/nidaqmx/_library_interpreter.py
```diff
@@ -1774,7 +1774,7 @@ def create_task(self, session_name):
         error_code = cfunc(
             session_name, ctypes.byref(task))
         check_for_error(error_code)
-        return task
+        return True, task
```
zhindes, line 1777: Per your description, I think `True` here is "session initialized boolean", which makes sense. Two things:
1. Is there a good way to indicate that here? Maybe creating a variables set to constant `True` that better clarifies what it means.
2. The gRPC interpreter doesn't do this, yet... that seems like a bad mismatch.

zhindes (changes requested):

generated/nidaqmx/_library_interpreter.py
Sakthi-SM: In `library_interpreter` I have added a private variable called `new_session_initialized` which will be set to `True` and reused where ever necessary. I have also updated the `gRPC_interpreter` to return the session initialized boolean from the response.

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -27,6 +27,7 @@ class LibraryInterpreter(BaseInterpreter):
         self._done_event_callbacks = []
         self._every_n_samples_event_callbacks = []
         self._signal_event_callbacks = []
+        self._new_session_initialized = True
```
bkeryan, line 30: `new_session_initialized` should be stored in the same place as the task handle. 

Because we are not storing the task handle in the interpreter object, one interpreter may create multiple tasks. Some of these tasks may already exist, and others may not.

Example:
```
system = nidaqmx.system.System.local() # creates interpreter
tasks = []
for persisted_task in system.tasks: # reuses interpreter
    tasks.append(persisted_task.load()) # reuses interpreter
```

I realize that `LibraryInterpreter.create_task` and `LibraryInterpreter.load_task` don't actually overwrite this field, but there is no reason for it to be a `LibraryInterpreter` field in the first place. These methods should either define it as a local variable or simply return a `True` literal. (The former is more readable, and I suspect that's what you were going for.)

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -232,9 +236,11 @@ def create_compound_parameter_request(func):
     return f"grpc_types.{compound_parameter_type}(" + ", ".join(parameters) + ")"
 
 
-def get_response_parameters(output_parameters: list):
+def get_response_parameters(output_parameters: list, is_init_method=False):
```
bkeryan, line 239: If this function took `func`, you wouldn't have to update its signature to change its behavior.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -1247,10 +1247,16 @@ def create_table_scale(
...
+        )
         response = self._invoke(
             self._client.CreateTask,
-            grpc_types.CreateTaskRequest(session_name=session_name))
-        return response.task
+            grpc_types.CreateTaskRequest(
+                initialization_behavior=self._grpc_options.initialization_behavior,
+                session_name=session_name),
+            metadata=metadata)
+        return response.new_session_initialized, response.task
```
bkeryan, line 1259: The task handle is the primary result. `new_session_initialized` is secondary, so please add it to the end, not the beginning.

bkeryan (changes requested):

bkeryan:
> NOTE: The comment also suggest updating the task.py to store this session initialized boolean and use it while closing the session. But this will be covered in a different PR. [Task 2375614](https://dev.azure.com/ni/DevCentral/_workitems/edit/2375614): Update the task.py to store and use the session initialized boolean

Until this is resolved, the `Task` object will set `self._handle` to a tuple and nothing will work.

zhindes (approved):

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -53,16 +53,31 @@ class GrpcStubInterpreter(BaseInterpreter):
...
+    %if (func.is_init_method):
+        metadata = (
+            ('ni-api-key', self._grpc_options.api_key),
+        )
     %endif
         response = self._invoke(
             self._client.${snake_to_pascal(func.function_name)},
         %if (len(func.function_name) + len(get_grpc_interpreter_call_params(func, sorted_params))) > 68:
             grpc_types.${snake_to_pascal(func.function_name)}Request(
+            %if (func.is_init_method):
```
bkeryan, line 66: Style: the parentheses on lines 66 and 73 are unnecessary.

bkeryan (approved):

PR #301, opened 2023-04-26
Use interpreter for C Calls | PR2 | Update handwritten functions to use interpreter for c calls

MounikaBattu17:
Updated below handwritten files to use interpreter.
```
stream_readers.py
stream_writers.py
device_collection.py
persisted_channel_collection.py
persisted_scale_collection.py
persisted_task_collection.py
physical_channel_collection.py
persisted_channel.py
persisted_scale.py
persisted_task.py
task.py
```
Implements [Task 2366649](https://dev.azure.com/ni/DevCentral/_workitems/edit/2366649): Update all System related functions with C calls with interpreter usage - Handwritten
[Task 2330459](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330459): Update Handwritten function files to use the interpreter instance for C calls


src/handwritten/stream_readers.py
```diff
@@ -218,9 +208,11 @@ def read_many_sample(
 
         self._verify_array(data, number_of_samples_per_channel, False, True)
 
-        return _read_analog_f_64(
-            self._handle, data, number_of_samples_per_channel,
-            timeout)
+        _, samps_per_chan_read = self._interpreter.read_analog_f64(
+            self._handle, number_of_samples_per_channel,
+            timeout, FillMode.GROUP_BY_CHANNEL.value, data)
```
zhindes, line 213: What PR is updating the interpreter functions to take data arrays? Your branch and the target branch doesn't have a data input after the `FillMode`

zhindes (changes requested):

src/handwritten/stream_readers.py
MounikaBattu17: Yes. This is quite confusing but the function call arguments change is present in `_libary_interpreter.py` of another PR https://github.com/ni/nidaqmx-python/pull/295.

generated/nidaqmx/system/_collections/device_collection.py
```diff
@@ -7,6 +7,7 @@
 from nidaqmx.error_codes import DAQmxErrors
 from nidaqmx.system.device import Device
 from nidaqmx.utils import unflatten_channel_string
+from nidaqmx.system.device import _DeviceAlternateConstructor
```
bkeryan, line 10: Style: combine `from X import A` and `from X import B` into `from X import A, B`

Normally ni-python-styleguide would complain that these imports are out of order, but we are still excluding src/handwritten and removing that exclusion will require a lot of changes.

src/handwritten/task.py
```diff
@@ -674,21 +619,30 @@ def read(self, number_of_samples_per_channel=NUM_SAMPLES_UNSET,
...
+                _, samps_per_chan_read, num_bytes_per_samp = self._interpreter.read_digital_lines(
+                    self._handle, number_of_samples_per_channel, timeout, 
+                    FillMode.GROUP_BY_CHANNEL.value, data)
+
+                ReadDigitalLinesReturnData = (
+                        collections.namedtuple(
+                            'ReadDigitalLinesReturnData',
+                            ['samps_per_chan_read', 'num_bytes_per_samp']))
+
+                samples_read = ReadDigitalLinesReturnData(samps_per_chan_read, num_bytes_per_samp).samps_per_chan_read
```
bkeryan, line 640: IMO, creating the namedtuple type in every function call is an anti-pattern. We should create namedtuple types at file scope. (Also, we should use `typing.NamedTuple`.)

In this case, there is no longer a need for a namedtuple.

```suggestion
                _, samples_read, _ = self._interpreter.read_digital_lines(
                    self._handle, number_of_samples_per_channel, timeout, 
                    FillMode.GROUP_BY_CHANNEL.value, data)
```

src/handwritten/stream_readers.py
```diff
@@ -1833,7 +1867,16 @@ def read_one_sample_multi_line(self, data, timeout=10):
...
+        _, samps_per_chan_read, num_bytes_per_samp = self._interpreter.read_digital_lines(
+            self._handle, data, 1, timeout, FillMode.GROUP_BY_CHANNEL.value)
+        
+        ReadDigitalLinesReturnData = (
+        collections.namedtuple(
+            'ReadDigitalLinesReturnData',
+            ['samps_per_chan_read', 'num_bytes_per_samp']))
+
+        ReadDigitalLinesData = ReadDigitalLinesReturnData(
+        samps_per_chan_read, num_bytes_per_samp)
```
bkeryan, line 1879: None of these namedtuples has any effect. The data is already in the passed-in `data` numpy array and the function doesn't return any values.

src/handwritten/task.py
bkeryan, line 1: Line 243 needs updated to use `_DeviceAlternateConstructor`:
```
        return [Device(v) for v in
                unflatten_channel_string(val.value.decode('ascii'))]
```

bkeryan (changes requested):

src/handwritten/task.py
bkeryan: FYI, I searched for `Device(` etc. but this branch is missing `device.py.mako` changes from the previous PR. I want to audit the code for constructor usage once all these PRs are done.

generated/nidaqmx/stream_readers.py
```diff
@@ -792,7 +792,7 @@ def read_many_sample(
         self._verify_array(voltage_data, number_of_samples_per_channel, False, True)
         self._verify_array(current_data, number_of_samples_per_channel, False, True)
 
-        _, samps_per_chan_read = self._interpreter.read_power_f64(
+        samps_per_chan_read = self._interpreter.read_power_f64(
```
bkeryan, line 795: Sorry, my previous guidance was inconsistent. The hand-written read functions returned the read array, so let's continue to return the read array in the auto-generated read functions, as well as the few that you have hand-written.

It is important for the library interpreter and gRPC interpreter to agree about this.

https://github.com/ni/nidaqmx-python/pull/295#discussion_r1178529912

> [bkeryan](https://github.com/bkeryan) [2 days ago](https://github.com/ni/nidaqmx-python/pull/295#discussion_r1178529912)
> read_array is a mutable in/out parameter, so there is no need to return it.
>
>@[bkeryan](https://github.com/bkeryan) bkeryan [2 days ago](https://github.com/ni/nidaqmx->python/pull/295#discussion_r1178566623)
Nevermind, the hand-written functions did this too.

bkeryan (changes requested):

src/handwritten/task.py
MounikaBattu17: Removed the named tuple and assigned `samples_read` to the return value directly.

src/handwritten/stream_readers.py
MounikaBattu17: Removed the named tuples.

src/handwritten/task.py
MounikaBattu17: > Line 243 needs updated to use `_DeviceAlternateConstructor`:
> 
> ```
>         return [Device(v) for v in
>                 unflatten_channel_string(val.value.decode('ascii'))]
> ```

Since, this series of PR deals with only functions, this updated is available in this https://github.com/ni/nidaqmx-python/pull/303. Please refer.

MounikaBattu17: > FYI, I searched for `Device(` etc. but this branch is missing `device.py.mako` changes from the previous PR. I want to audit the code for constructor usage once all these PRs are done.

I have done `git merge users/mounika/add-interpreter-functions-mako`. Now updated device.py is available in this branch. please check.

generated/nidaqmx/stream_readers.py
MounikaBattu17: I have updated these read functions to return `read_array`. Also, updated the callers too.

bkeryan (approved):

PR #302, opened 2023-04-26
[PR1] [Properties] Fix physical channel set, reset property c calls to adapt interpreter.

samchris007:
This PR fixes physical channel set, reset properties to adapt interpreter.
Implementation is done based on this comment https://github.com/ni/nidaqmx-python/pull/298#discussion_r1176900867

### Why should this PR be merged?
- Rename the old `property_setter_template` to `property_legacy_setter_template` to reflect that this is a special case.
- Rename `property_interpretercall_setter_template` to `property_setter_template` to reflect that this is the common case.
- Fork the old `property_deleter_template` and name it `property_legacy_deleter_template`.
- Add an error check to `property_legacy_setter_template` and `property_legacy_deleter_template`:
```
if not isinstance(self._interpreter, LibraryInterpreter):
    raise NotImplementedError
```

Manually generated the files and tested.


zhindes (approved):

src/codegen/templates/property_setter_template.py.mako
```diff
@@ -22,47 +23,20 @@
...
+        function_call_args.append(hex(attribute.id))
         function_call_args.append('val')
+
         if attribute.has_explicit_write_buffer_size:
             function_call_args.append('len(val)')
     %>\
-        error_code = cfunc(
-            ${', '.join(function_call_args) | wrap(initial_indent=12)})
-        check_for_error(error_code)
+        self._interpreter.set_${generic_attribute_func}(${', '.join(function_call_args) | wrap(12)})
```
bkeryan, line 41: This makes the 1st line extra long because wrap(12) doesn't account for `self._interpreter.set_${generic_attribute_func}(`. Drop the args to the next line.
```suggestion
        self._interpreter.set_${generic_attribute_func}(
            ${', '.join(function_call_args) | wrap(12)})
```

src/codegen/templates/property_template.py.mako
```diff
@@ -3,21 +3,24 @@
...
+${property_legacy_setter_template.script_property_setter(attribute)}
     %endif
 \
-    %if attribute.resettable:
+    %if attribute.resettable and attribute.python_class_name != "PhysicalChannel":
 ${property_deleter_template.script_property_deleter(attribute)}
+    %elif attribute.resettable and attribute.python_class_name == "PhysicalChannel":
+${property_legacy_deleter_template.script_property_deleter(attribute)}
     %endif
 </%def>
```
bkeryan, line 26: Moving the special case above the normal case would simplify the conditionals.

```
%elif attribute.access == "read-write" and attribute.python_class_name == "PhysicalChannel":
%elif attribute.access == "read-write":
%endif

%if attribute.resettable and attribute.python_class_name == "PhysicalChannel":
%elif attribute.resettable:
%endif
```

bkeryan (approved):
Approved with suggestions

src/codegen/templates/property_setter_template.py.mako
samchris007: Ok. As discussed in this PR comment https://github.com/ni/nidaqmx-python/pull/298#discussion_r1178559224, I have removed `wrap()` since the length of the `setter()` properties are always <68 whitespaces.

src/codegen/templates/property_template.py.mako
samchris007: Ok. Updated.

PR #304, opened 2023-04-27
gRPC interpreter | PR4 | Handle RPC exceptions in gRPC interpreter

Sakthi-SM:

The gRPC interpreter has to send the `gRPC` errors as a `DAQ` errors to the caller. The following changes are made to achieve this:

* Updated the `invoke` method in `grpc_interpreter.mako` to convert the gRPC errors to appropriate errors.

* The changes have been tested manually.

NOTE: The read and write methods needs additional parameters to be sent from the gRPC server which is currently not sent. So these methods haven't been tested and will be taken care of in [Task 2366676](https://dev.azure.com/ni/DevCentral/_workitems/edit/2366676): Update the gRPC device server with the necessary changes
This implements [Task 2351331](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351331): All errors thrown via gRPC client should be propagated as DaqError


generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -22,13 +22,42 @@ def __init__(self, grpc_options):
...
+                        error_code = int(value)
+                    except ValueError:
+                        error_message += f'\nError status: {value}'
+                if entry.key == "samps_per_chan_read":
+                    samps_per_chan_read = entry.value
+                if entry.key == "samps_per_chan_written":
+                    samps_per_chan_written = entry.value
+            grpc_error = rpc_error.code()
+            if grpc_error == grpc.StatusCode.UNAVAILABLE:
+                error_message = 'Failed to connect to server'
```
Sakthi-SM, line 43: Error code is used as -1 for the `gRPC` specific errors. Should this be handled differently?

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -22,13 +22,42 @@ def __init__(self, grpc_options):
...
+            elif grpc_error == grpc.StatusCode.UNIMPLEMENTED:
+                error_message = (
+                    'This operation is not supported by the NI gRPC Device Server being used. Upgrade NI gRPC Device Server.'
+                )
+            self.raise_error(error_code, error_message, samps_per_chan_written, samps_per_chan_read)
         return response
 
+    def raise_error(self, error_code, error_message, samps_per_chan_written=None, samps_per_chan_read=None):
+        if error_code < 0:
```
Sakthi-SM, line 52: What is the expectations for the errors whose error code are greater than 0? How are we supposed to handle it?

generated/nidaqmx/_grpc_interpreter.py
zhindes: You can see how nimi-python handles this situation here: https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/_grpc_stub_interpreter.py.

The summary is that they created a new class of error called an [`RpcError`](https://github.com/ni/nimi-python/blob/ca585de99ca191703889eb88841efc4cc6f59b09/generated/nidcpower/nidcpower/errors.py#L46) that is different from a `DriverError` for situations like this. They initialize the code to `None` and raise the `RpcError` in that case. I think that is a pattern we should follow here.

generated/nidaqmx/_grpc_interpreter.py
zhindes: errors > 0 are a warning. We use the `warnings` module for that.

zhindes (changes requested):

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -22,13 +22,42 @@ def __init__(self, grpc_options):
...
+                if entry.key == 'ni-error':
+                    value = entry.value if isinstance(entry.value, str) else entry.value.decode('utf-8')
+                    try:
+                        error_code = int(value)
+                    except ValueError:
+                        error_message += f'\nError status: {value}'
+                if entry.key == "samps_per_chan_read":
+                    samps_per_chan_read = entry.value
+                if entry.key == "samps_per_chan_written":
+                    samps_per_chan_written = entry.value
```
bkeryan, line 40: 1. gRPC metadata is like HTTP headers, so key names should use dashes, not underscores. 
2. We are prefixing NI-specific metadata with `ni-` to ensure that it doesn't conflict with metadata used by HTTP proxies and such.
3. These if-statements are mutually exclusive, so use `elif`.
4. The samps per chan read/written cases should decode `entry.value` if necessary and convert it to `int`.

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
bkeryan: Note about decoding `bytes` -> `str`: only keys that end with `-bin` should contain `bytes`. However, mypy doesn't know that. `assert isinstance(entry.value, str)`  or `int(typing.cast(str, entry.value))` would also work.

generated/nidaqmx/_grpc_interpreter.py
Sakthi-SM: I have created an `RPCError` class in `errors.py` which will handle the errors specific to the gRPC server.

generated/nidaqmx/_grpc_interpreter.py
Sakthi-SM: I have created a `DriverWarning` class to handle all warnings in DAQmx. But there already exists a `DaqWarning` class in `errors.py`, so should we try to reuse it or can we stick with the newly added warning class. I am not sure where this `DaqWarning` is used.

generated/nidaqmx/_grpc_interpreter.py
Sakthi-SM: Updated the method as per the suggestion

keer1111:
@bkeryan , Please review.

bkeryan (changes requested):

bkeryan (approved):

keer1111:
@zhindes , since you requested changes, this PR is blocked until you review again. Please provide review if by chance you are available.

@bkeryan, If there is a possibility of bypassing Zach, Please consider it.

bkeryan:
> @zhindes , since you requested changes, this PR is blocked until you review again. Please provide review if by chance you are available.
> 
> @bkeryan, If there is a possibility of bypassing Zach, Please consider it.

Done.

PR #305, opened 2023-04-27
gRPC interpreter | PR5 | Handle read methods with array parameters

Sakthi-SM:

As per the suggestion made in https://github.com/ni/nidaqmx-python/pull/285#discussion_r1170618836 the following changes have been made:
*  Assigned `array.size` to `array_size_in_samps` in gRPC response.
*  Assigned the `read_array` in response to the input array parameter.
* Updated the `interpreter_helper` with the necessary helper methods.
 

This implements [Task 2374111](https://dev.azure.com/ni/DevCentral/_workitems/edit/2374111): Update read methods in grpc interpreter with read array instead of array sample size


zhindes (approved):

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -56,11 +57,16 @@ class GrpcStubInterpreter(BaseInterpreter):
...
             grpc_types.${snake_to_pascal(func.function_name)}Request(
-                ${get_grpc_interpreter_call_params(func, sorted_params) + ')' | wrap(16, 16)})
+                ${grpc_interpreter_params + ')' | wrap(16, 16)})
         %else:
-            grpc_types.${snake_to_pascal(func.function_name)}Request(${get_grpc_interpreter_call_params(func, sorted_params)+ ')'})
+            grpc_types.${snake_to_pascal(func.function_name)}Request(${grpc_interpreter_params + ')'})
+        %endif
+        %if (check_if_parameters_contain_read_array(sorted_params)):
+        % for param in get_read_array_parameters(func):
+        ${param}[:]=response.${param}
```
bkeryan, line 68: Style: put spaces around `=` in assignments, omit spaces around `=` when declaring optional parameters.

(This seems to match Black's formatting.)

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -204,10 +204,19 @@ def get_interpreter_params(func):
...
         if param.parameter_name not in compound_params:
-            if param.is_enum:
+            if is_read_function and "read_array" in param.parameter_name:
+                if has_read_array_parameter:
+                    continue
+                grpc_params.append(
+                    f"{camel_to_snake_case(param.size.value)}={param.parameter_name}.size"
+                )
+                has_read_array_parameter = True
+            elif param.is_enum and not param.is_list:
```
bkeryan, line 219: I think checking is_list is incorrect. If there was a Set*AttributeInt32Array, it would use `oneof { enum value, int32 value_raw`.

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -2233,9 +2233,10 @@ def read_analog_f64(
...
             self._client.ReadAnalogF64,
             grpc_types.ReadAnalogF64Request(
-                fill_mode_raw=fill_mode,
-                num_samps_per_chan=num_samps_per_chan, read_array=read_array,
-                task=task, timeout=timeout))
+                array_size_in_samps=read_array.size, fill_mode_raw=fill_mode,
+                num_samps_per_chan=num_samps_per_chan, task=task,
+                timeout=timeout))
+        read_array[:]=response.read_array
         return response.read_array, response.samps_per_chan_read
```
bkeryan, line 2240: These are returning the wrong read array. They should return the numpy read array, not the gRPC read array.

bkeryan (changes requested):

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -204,10 +204,19 @@ def get_interpreter_params(func):
...
     for param in params:
         if param.parameter_name not in compound_params:
-            if param.is_enum:
+            if is_read_function and "read_array" in param.parameter_name:
+                if has_read_array_parameter:
+                    continue
+                grpc_params.append(
+                    f"{camel_to_snake_case(param.size.value)}={param.parameter_name}.size"
+                )
+                has_read_array_parameter = True
```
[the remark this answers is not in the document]
Sakthi-SM: This was added because the parameters were getting duplicated earlier. For example, `GetDigitalPowerUpStates` method had `ChannelName` twice in metadata and this was getting reflected in the request. Now that this has been fixed as part of a PR which was merged earlier, I'll remove this and generate the parameters in the existing order.

generated/nidaqmx/_grpc_interpreter.py
bkeryan: By "return the numpy read array", I meant do this:

```
read_array[:]=response.read_array
return read_array, response.samps_per_chan_read
```

You did that, but you also changed the code to allocate a temporary numpy array:
```
read_array[:] = numpy.array(response.read_array)
return read_array, response.samps_per_chan_read
```

The reason why some of the read functions accept a numpy array from the user is to avoid unnecessarily allocating and copying arrays. DAQmx applications may work with large amounts of data, so when working on DAQmx read/write functions, we should assume that allocations and copies are expensive.

The right hand side of `read_array[:] = ...` does not need to be a numpy array. It can be a list:
```
>>> import numpy
>>> read_array = numpy.zeros(3, dtype=numpy.double)
>>> read_array
array([0., 0., 0.])
>>> read_array[:] = [1.1,2.2,3.3]
>>> read_array
array([1.1, 2.2, 3.3])
```

Or a tuple:
```
>>> read_array[:] = (3.3,4.4,5.5)
>>> read_array
array([3.3, 4.4, 5.5])
```

Or a Protobuf repeated field:
```
>>> import nidaqmx._stubs.nidaqmx_pb2
>>> response = nidaqmx._stubs.nidaqmx_pb2.ReadAnalogF64Response()
>>> response.read_array.extend([6.6,7.7,8.8])
>>> response
read_array: 6.6
read_array: 7.7
read_array: 8.8

>>> read_array[:] = response.read_array
>>> read_array
array([6.6, 7.7, 8.8])
```

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -366,12 +365,15 @@ def create_compound_parameter_request(func):
     return f"grpc_types.{compound_parameter_type}(" + ", ".join(parameters) + ")"
 
 
-def get_response_parameters(output_parameters: list):
+def get_response_parameters(output_parameters: list, is_read_method):
```
bkeryan, line 368: As mentioned in other PRs, passing `func` into the function would allow you to change its behavior without changing its signature.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -63,12 +65,12 @@ class GrpcStubInterpreter(BaseInterpreter):
         %else:
             grpc_types.${snake_to_pascal(func.function_name)}Request(${grpc_interpreter_params + ')'})
         %endif
-        %if (check_if_parameters_contain_read_array(sorted_params)):
+        %if is_read_method:
         % for param in get_read_array_parameters(func):
-        ${param}[:]=response.${param}
+        ${param}[:] = numpy.array(response.${param})
```
bkeryan, line 70: Remove the call to `numpy.array`

bkeryan (changes requested):

bkeryan (approved):

PR #303, opened 2023-04-27
[PR2] [Properties] Update handwritten property function calls for get, set, reset to use interpreter.

samchris007:
### What does this PR accomplish?
This PR updates handwritten property function calls for get, set, reset to use interpreter for function calls.

### Why should this PR be merged?
Implements [Task 2330458](https://dev.azure.com/ni/DevCentral/_workitems/edit/2330458): Update Property templates mako to use interpreter for instance for C calls

### What testing has been done?
Manually generated the files and tested.

PR #307, opened 2023-04-27
Library Interpreter | PR1 | Refactor the size parameter addition logic in interpreter classes. 

subash-suresh:
This PR aims on focuses on the following things. 
- refactor the logic for adding size parameters to add parameters based on the metadata.
- update the logic for defining argtypes for handle parameters.

The following updates have been made in this PR,
- The logic to skip certain parameters for interpreters have been removed. 
    - The functions `get_skippable_params_for_interpreter_func` and `is_skippable_param` have been completely removed.
- The `interpreter_parameters` property has been removed from the `functions.py` file and a separate method has been created in the `interpreter_helpers.py` to get the parameters.
    - The new function `get_interpreter_parameters` filters parameters from the `base_parameters`. The parameters are filtered based on the fields `use_python_api` and `proto_only` fields in the metadata. 
    - If a parameter is a size parameter or a handle parameter, then it is considered as an interpreter parameter irrespective of `use_python_api` and `proto_only` fields in the metadata.
    - The method `get_interpreter_params` has been renamed to `get_params_for_function_signature` since it was similarly named to the new function.
-  Updated the logic for getting the function call arguments in `generate_interpreter_function_call_args`,
    - The addition of size parameter is no longer done by assuming that a size parameter always follows a array parameter. The updated logic identifies the size parameter in the parameter list and handles it seperately.
    - If the parameter is reserved, then the argument to be added is considered as `None`
- Updates in function signature,
    - The function signature logic has been updated to not include any size parameter or reserved parameters.
    - To handle some special cases that does not follow the above statement a list called `INCLUDE_SIZE_PARAMETER_IN_SIGNATURE_FUNCTIONS` has been added. Currently only `get_analog_power_up_states_with_output_type"` needs the size parameter in its signature.
- Updated the logic to get argument types
    -  Created a new method in `interpreter_helpers.py` to get argument types similar to the one in `function_helpers`,
    -  All the mako templates have been updated to use this new function.
    - The previous logic first added argument types for handle parameters and then added argument types for the rest of the parameters. But this caused a repeat of argtypes as the parameter filtration did not filter all handle parameters and only filtered the parameters with the name `task`. To fix this the current logic takes in all the parameters, and if the parameter is a handle parameter, then the ctypes data type information is taken accordingly.
    - If the parameter is a size parameter, then `ctypes.c_uint` is added.
    - the logic to map `ctypes.TaskHandle` to `lib_importer.task_handle` has been added to the `to_param_argtype` function to properly define this commonly occurring type. This has been added for handling the cases where there is task parameter without a handle parameter definition. This is case is true for the read/write function metadata.

- All the automated tests will pass if the library interpreter from this PR is run with the updates in https://github.com/ni/nidaqmx-python/pull/303, which has the interpreter implementation for all properties.
-  manually verified the updates in the functions logic.


zhindes (approved):

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -54,67 +57,69 @@ def get_interpreter_functions(metadata):
...
-                if param.parameter_name == size_parameter.name:
-                    function_call_args.append(size_parameter.size)
+    return function_call_args
 
-    # C function call for reserved parameter for read/write function is passed as None.
-    if function_metadata.python_codegen_method == "CustomCode_Read_Write":
-        function_call_args.append("None")
 
-    return function_call_args
+def get_arguments_type(functions_metadata):
```
bkeryan, line 103: Naming: `get_argument_types`

It returns multiple types. The word "argument" modifies "type(s)".

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2268,7 +2268,7 @@ def get_analog_power_up_states_with_output_type(
             cfunc.argtypes = [
                 ctypes_byte_str, wrapped_ndpointer(dtype=numpy.float64,
                 flags=('C','W')), wrapped_ndpointer(dtype=numpy.int32,
-                flags=('C','W')), ctypes.c_uint, ctypes.c_uint32]
+                flags=('C','W')), ctypes.c_uint]
```
bkeryan, line 2271: The size is passed by pointer, so this should use `ctypes.POINTER` and `ctypes.byref` for the size. I think the original code may be wrong; it doesn't add the size to argtypes, so it will pass the size by value.
```
int32 __CFUNC_C DAQmxGetAnalogPowerUpStatesWithOutputType(const char channelNames[], float64 stateArray[], int32 channelTypeArray[], uInt32 *arraySizePtr);
```

https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/system/system.py#L633
```
        if cfunc.argtypes is None:
            with cfunc.arglock:
                if cfunc.argtypes is None:
                    cfunc.argtypes = [
                        ctypes_byte_str,
                        wrapped_ndpointer(dtype=numpy.float64,
                                          flags=('C','W')),
                        wrapped_ndpointer(dtype=numpy.int32,
                                          flags=('C','W'))]

        error_code = cfunc(
            flatten_channel_string(physical_channels), states, channel_types,
            size)
```

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -43,6 +42,10 @@
     "write_raw",
 ]
 
+INCLUDE_SIZE_PARAMETER_IN_SIGNATURE_FUNCTIONS = [
+    "get_analog_power_up_states_with_output_type",
```
bkeryan, line 46: I think this is a special case because the metadata is wrong.

The size is passed by pointer:
```
int32 __CFUNC_C DAQmxGetAnalogPowerUpStatesWithOutputType(const char channelNames[], float64 stateArray[], int32 channelTypeArray[], uInt32 *arraySizePtr);
```

But the metadata says direction:in and doesn't say pointer:True.
https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/functions.py#L12829
```
            {
                'ctypes_data_type': 'ctypes.c_uint32',
                'direction': 'in',
                'name': 'arraySize',
                'python_data_type': 'int',
                'type': 'uInt32'
            }
```

bkeryan (changes requested):

src/codegen/utilities/interpreter_helpers.py
bkeryan: If you want to handle this in a separate PR, that's okay.

keer1111: As you have mentioned, this is a special case and we have manually handled in scrapigen metadata generation in transform, hence updated the direction in this [PR](https://github.com/ni/grpc-device-scrapigen/pull/168). (Only updated the direction since pointer flag is not used and out is used pass as pointer. Let us know if you feel that is also essential in metadata)

src/codegen/metadata/functions.py
```diff
@@ -12837,7 +12837,7 @@
             },
             {
                 'ctypes_data_type': 'ctypes.c_uint32',
-                'direction': 'in',
+                'direction': 'out',
```
bkeryan, line 12840: I think this will remove the `arraySize` field from `GetAnalogPowerUpStatesWithOutputTypeRequest`. Instead, set `pointer=True` for this parameter.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -101,19 +101,25 @@ def generate_interpreter_function_call_args(function_metadata):
...
+                else:
+                    argtypes.append(f"ctypes.POINTER({functions_metadata.handle_parameter.ctypes_data_type})")
             else:
                 argtypes.append("ctypes_byte_str")
         elif param.parameter_name in size_params:
-            argtypes.append("ctypes.c_uint")
+            if(param.direction == "in"):
+                argtypes.append("ctypes.c_uint")
+            else:
+                argtypes.append("ctypes.POINTER(ctypes.c_uint)")
```
bkeryan: Update this to take the `pointer` field into account: use ctypes.POINTER when `param.direction == "out" or` `param.is_pointer`

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -101,19 +101,25 @@ def generate_interpreter_function_call_args(function_metadata):
...
-                argtypes.append(functions_metadata.handle_parameter.ctypes_data_type)
+                if param.direction == "in":
+                    argtypes.append(functions_metadata.handle_parameter.ctypes_data_type)
+                else:
+                    argtypes.append(f"ctypes.POINTER({functions_metadata.handle_parameter.ctypes_data_type})")
             else:
                 argtypes.append("ctypes_byte_str")
         elif param.parameter_name in size_params:
-            argtypes.append("ctypes.c_uint")
+            if(param.direction == "in"):
```
bkeryan: Style: no parens here

bkeryan (changes requested):

bkeryan (approved):

PR #308, opened 2023-04-28
Handle session initialization of task and watchdog task

MounikaBattu17:
- Added new boolean `_is_new_session_initialized` in task and watchdog task to validate `close()` task.
- Passed in `_is_new_session_initialized` to `_TaskAlternateConstructor` to from `persisted_task.load()`.
- Validated `_is_new_session_initialized` in `close()` and `__exit__`.

Implements [Task 2375614](https://dev.azure.com/ni/DevCentral/_workitems/edit/2375614): Update the task.py to store and use the session initialized boolean


generated/nidaqmx/system/watchdog.py
```diff
@@ -538,7 +540,7 @@ def close(self):
         within the loop after you are finished with the task to avoid
         allocating unnecessary memory.
         """
-        if self._handle is None:
+        if self._handle is None or not self._is_new_session_initialized:
```
zhindes, line 543: `_is_new_session_initialized` being `False` isn't a failure case. The user may have specifically requested to attach to an existing task rather than create a new one. They can close that, if they like. The alternative to close is detach, which "leaks" the task on the server.

zhindes: I think you can just remove this check

generated/nidaqmx/system/watchdog.py
```diff
@@ -77,7 +78,8 @@ def __enter__(self):
         return self
 
     def __exit__(self, type, value, traceback):
-        self.close()
+        if self._is_new_session_initialized:
```
zhindes, line 81: This is interesting... The alternative to close is `detach`, a new method, which leaks the session on the server (intentionally), so we wouldn't call close. That being said, I just checked nimi-python, and this is what it does, too. So nevermind :D

zhindes: Actually, nimi-python calls the member variable `_close_on_exit` which is a bit more clear. In the grpc interpreter they do this:

```python
self._close_on_exit = response.new_session_initialized
```

Then the top-level tasks/sessions check the interpreter's value of `close_on_exit` to close or not.

generated/nidaqmx/system/watchdog.py
zhindes: Confirmed, just remove it. If someone explicitly calls `close`, then we should close.

generated/nidaqmx/task.py
```diff
@@ -436,7 +438,7 @@ def close(self):
         within the loop after you are finished with the task to avoid
         allocating unnecessary memory.
         """
-        if self._handle is None:
+        if self._handle is None or not self._is_new_session_initialized:
```
zhindes, line 441: remove this check

generated/nidaqmx/task.py
```diff
@@ -1220,16 +1223,18 @@ class _TaskAlternateConstructor(Task):
...
         Args:
             task_handle: Specifies the task handle from which to create a
                 Task object.
             interpreter: Specifies the interpreter instance.
+            is_new_session_initialized: Specifies if new session is initialized or not.
             
         """
         self._handle = task_handle
         self._interpreter = interpreter
+        self._is_new_session_initialized = is_new_session_initialized
```
zhindes, line 1237: I think `_close_on_exit` is better

zhindes (changes requested):

generated/nidaqmx/system/watchdog.py
MounikaBattu17: Removed the conditional check for close on exit.

generated/nidaqmx/system/watchdog.py
MounikaBattu17: Thanks for the information! Updated the name of `_close_on_exit`.

generated/nidaqmx/task.py
MounikaBattu17: Removed.

generated/nidaqmx/task.py
MounikaBattu17: Thanks for the information! Updated the name to _close_on_exit.

zhindes (approved):

src/handwritten/task.py
```diff
@@ -1220,16 +1223,18 @@ class _TaskAlternateConstructor(Task):
     # Setting __slots__ avoids TypeError: __class__ assignment: 'Base' object layout differs from 'Derived'.
     __slots__ = []
 
-    def __init__(self, task_handle, interpreter):
+    def __init__(self, task_handle, interpreter, close_on_exit=False):
```
bkeryan, line 1226: Please make this parameter required to ensure we don't omit it by accident. All calls to `_TaskAlternateConstructor` correctly specify it.

```
PS C:\dev\nidaqmx-python> rg "_TaskAlternateConstructor"
src\handwritten\task.py
1217:class _TaskAlternateConstructor(Task):

src\handwritten\system\storage\persisted_task.py
147:        return task._TaskAlternateConstructor(task_handle, self._interpreter, close_on_exit)

generated\nidaqmx\task.py
1217:class _TaskAlternateConstructor(Task):

generated\nidaqmx\system\storage\persisted_task.py
147:        return task._TaskAlternateConstructor(task_handle, self._interpreter, close_on_exit)
```

```suggestion
    def __init__(self, task_handle, interpreter, close_on_exit):
```

src/handwritten/task.py
```diff
@@ -1220,16 +1223,18 @@ class _TaskAlternateConstructor(Task):
...
         """
         Args:
             task_handle: Specifies the task handle from which to create a
                 Task object.
             interpreter: Specifies the interpreter instance.
+            close_on_exit: Specifies whether to close task on exit.
```
bkeryan, line 1232: Clarification
```suggestion
            close_on_exit: Specifies whether the task's context manager closes the task.
```

bkeryan (approved):
Approved with suggestions

src/handwritten/task.py
```diff
@@ -71,6 +71,7 @@ def __init__(self, new_task_name='', *, grpc_options=None):
                 results in an error.
         """
         self._handle = lib_importer.task_handle(0)
+        self._close_on_exit = False
```
bkeryan, line 74: These assignments are no longer needed. The code below will initialize these fields or raise an exception.

bkeryan: Oops, `__del__` warns if `self._handle` doesn't exist. Keep the assignments, but initialize `self._handle` to `None`.

src/handwritten/task.py
MounikaBattu17: Changed the docstring.

src/handwritten/task.py
MounikaBattu17: Initialized `self._handle = None`.

MounikaBattu17:
@bkeryan FYI, Respective this https://github.com/ni/nidaqmx-python/pull/313#issuecomment-1532166692, I have initialized `self._handle` to `None` in `watchdog.py`.

generated/nidaqmx/system/watchdog.py
```diff
@@ -56,10 +56,10 @@ def __init__(self, device_name, task_name='', timeout=10, grpc_options=None):
         """
         self._interpreter = utils._select_interpreter(grpc_options)
 
+        self._handle = None
+        self._close_on_exit = False
 
-        self._handle = lib_importer.task_handle(0)
-
-        self._handle, _ = self._interpreter.create_watchdog_timer_task_ex(device_name, task_name, timeout)
+        self._handle, self._close_on_exit = self._interpreter.create_watchdog_timer_task_ex(device_name, task_name, timeout)
```
bkeryan, line 62: Here's the error case we need to be careful about: if `__init__` raises an exception, `__del__` is still called. If `__init__` raises an exception before it initializes `self._handle`, then when `__del__` tries to access `self._handle`, it will raise `AttributeError` and the [Python interpreter will print a warning](https://docs.python.org/3/reference/datamodel.html#object.__del__).

There are a lot of different ways to solve this:
1. Change `__init__` to set `self._handle = None` before calling anything that can raise an exception. This is the path you are on.
2. Change `__init__` to use try/except to initialize `self._handle = None` in error cases.
3. Change `__del__` to use `getattr(self, "_handle", None)`.
4. Change `__del__` to use try/except.

We don't need to initialize `self._close_on_exit` twice. 
- `__del__` doesn't use `self._close_on_exit`.
- `__exit__` uses `self._close_on_exit`, but `__exit__` is only called after `__init__` succeeds. `__exit__` may be called after `__enter__` raises an exception, according to https://docs.python.org/3/reference/compound_stmts.html#the-with-statement

To continue with solution 1, please make the following changes to both `Task` and `WatchdogTask`:
- Delete `self._close_on_exit = False`.
- Put this at the top of `__init__`, right below the docstring. Change `new_task_name` to `task_name` for `WatchdogTask`.
```
# Initialize the fields that __del__ accesses so it doesn't crash when __init__ raises an exception.
self._handle = None
self._saved_task = new_task_name
```

generated/nidaqmx/system/watchdog.py
```diff
@@ -56,10 +56,10 @@ def __init__(self, device_name, task_name='', timeout=10, grpc_options=None):
...
 
         # Saved name is used in self.close() to throw graceful error on
```
bkeryan, line 64: Please change`__del__` to use `self._saved_name` instead of `self.name`, like it does in `Task`.

Reason: `self.name` calls into the interpreter, which is more likely to raise an exception, especially if the process is shutting down.

generated/nidaqmx/system/watchdog.py
```diff
@@ -374,6 +375,7 @@ def close(self):
         self._interpreter.clear_task(self._handle)
 
         self._handle = None
+        self._close_on_exit = False
```
bkeryan, line 378: On further thought, this line changes behavior, so please delete it. This applies to both `Task` and `WatchdogTask`.

Explanation:

With the prior release, calling `close` twice prints a warning:
```
>>> import nidaqmx
>>> task = nidaqmx.Task()
>>> task.close()
>>> task.close()
D:\dev\measurementlink-python\examples\nidaqmx_analog_input\.venv\lib\site-packages\nidaqmx\task.py:456: ResourceWarning: Attempted to close NI-DAQmx task of name "_unnamedTask<0>" but task was already closed.
  warnings.warn(
```

It also prints a warning if the context manager calls `close` the 2nd time:
```
>>> with nidaqmx.Task() as task:
...     task.close()
...
D:\dev\measurementlink-python\examples\nidaqmx_analog_input\.venv\lib\site-packages\nidaqmx\task.py:456: ResourceWarning: Attempted to close NI-DAQmx task of name "_unnamedTask<1>" but task was already closed.
  warnings.warn(
```

If `close` clears `self._close_on_exit`, then the 2nd snippet will no longer print a warning. This allows users to double-close the task when using a context manager.

Hypothetical: If we wanted to allow multiple calls to `close`, the right way to do it would be to remove the warning.

generated/nidaqmx/system/storage/persisted_task.py
```diff
@@ -135,9 +135,10 @@ def load(self):
         Returns:
             nidaqmx.task.Task: Indicates the loaded Task object.
         """
-        task_handle = self._interpreter.load_task(self._name)
+        task_handle = None
```
bkeryan, line 138: This is not an `__init__` function, so we don't have to worry about `__del__` here. Delete this line.

bkeryan (changes requested):

generated/nidaqmx/system/watchdog.py
MounikaBattu17: Thanks for the detailed approach. I have done the following changes.
- Deleted self._close_on_exit = False.
-Added the two initializations . Changed task_name to new_task_name for WatchdogTask.

bkeryan: > Changed task_name to new_task_name for WatchdogTask.

That is the opposite of what I suggested.

bkeryan (changes requested):

generated/nidaqmx/system/watchdog.py
```diff
@@ -28,7 +28,7 @@ class WatchdogTask:
     """
     Represents the watchdog configurations for a DAQmx task.
     """
-    def __init__(self, device_name, task_name='', timeout=10, grpc_options=None):
+    def __init__(self, device_name, new_task_name='', timeout=10, grpc_options=None):
```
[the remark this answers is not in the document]
MounikaBattu17: Oh the code snippet!!!. My bad. Sorry for the misinterpretation. I have reverted back the changes.

generated/nidaqmx/system/watchdog.py
MounikaBattu17: I have reverted back the `new_task_name `to `task_name`.

bkeryan (approved):

PR #310, opened 2023-04-28
Library Interpreter | PR2 | Fix the function arguments in the signal event functions

subash-suresh:
- An issue was identified with the signal even methods where the methods were passing an argument called `callback_function` in place of `callback_method_ptr`. This issue has been fixed in this PR.

- The logic for adding function call arguments has been updated to use `callback_method_ptr` instead of `callback_function` in case of the signal event methods.

This change has been manually verified.


zhindes (approved):

keer1111:
This change passes the "create_ai_voltage_chan" snippet Brad shared in other PRs for Library Interpreter, with a minor fix in argtypes of "read_analog_scalar_f64" method. Will fix this in another PR, along with other similar fixes if any.

PR #313, opened 2023-04-30
tests: Add task init/close unit tests

bkeryan:
## What does this Pull Request accomplish?

- Now that the C DLL details are abstracted away by the interpreter class, we can write unit tests for the public `nidaqmx` APIs.
- Enable `ResourceWarning` checks while running tests.
- Fix bugs found by new tests:
  - `Task.__init__` error cases access `self.name` before `self._interpreter` has been initialized. It's safer to use `new_task_name` or `self._saved_name` instead.
  - `Task.__del__` reports `ResourceWarning` when leaking a task with `_close_on_exit=False`. For now, only warn when `_close_on_exit=True`. We will revisit the explicit gRPC detach case in AB#2383312.

## Why should this Pull Request be merged?

Add some test coverage to the PR pipeline.

- Ran `poetry run pytest tests/unit` and `poetry run pytest`.
- Verified that the PR workflow ran the tests.


bkeryan:
> Running test unit tests still requires installing NI-DAQmx for now: AB#2379526: Importing nidaqmx module tries to load nicaiu.dll

Oops, it was the call to `self._handle = lib_importer.task_handle(0)`, not the import itself. Fixed.

MounikaBattu17:
> 

The assignment `self._handle = lib_importer.task_handle(0)`  is also present in `watchdog.py`, So can we apply the change in `watchdog.py` also?

MounikaBattu17:
> > Running test unit tests still requires installing NI-DAQmx for now: AB#2379526: Importing nidaqmx module tries to load nicaiu.dll
> 
> Oops, it was the call to `self._handle = lib_importer.task_handle(0)`, not the import itself. Fixed.

The assignment self._handle = lib_importer.task_handle(0) is also present in watchdog.py, So can we apply the change in watchdog.py also?

maxxboehme (approved):

bkeryan:
> The assignment `self._handle = lib_importer.task_handle(0)` is also present in `watchdog.py`, So can we apply the change in `watchdog.py` also?

Yes

MounikaBattu17 (approved):

src/handwritten/task.py
```diff
@@ -79,15 +80,15 @@ def __init__(self, new_task_name='', *, grpc_options=None):
...
         self._interpreter = utils._select_interpreter(grpc_options)
         self._handle, self._close_on_exit = self._interpreter.create_task(new_task_name)
 
         self._initialize(self._handle, self._interpreter)
 
     def __del__(self):
-        if self._handle:
+        if self._handle is not None and self._close_on_exit:
             warnings.warn(
                 'Task of name "{}" was not explicitly closed before it was '
```
bkeryan, line 93: I think we need to revisit this for the explicit gRPC detach use case. I filed AB#2383312 about it.

maxxboehme (approved):

PR #315, opened 2023-05-02
Tests Update | PR1 | Add test fixtures to test gRPC and library interpreters

subash-suresh:
What does this PR accomplish?
This PR adds fixtures to create task instance to test the gRPC and library interpreter implementations.

Why should this PR be merged?
- Added `library_session_creation_kwargs` fixture for creating session based on library interpreter.
- Added `grpc_session_creation_kwargs` fixture for creation sessions based on grpc interpreters.
- Added `session_creation_kwargs` fixture for creation of sessions.
- Added `task` fixture for creation of task sessions.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -253,8 +253,8 @@ def test_assets_directory() -> pathlib.Path:
     return pathlib.Path(__file__).parent / "test_assets"
 
 
-@pytest.fixture(scope="package")
-def grpc_session_creation_kwargs():
+@pytest.fixture(scope="session")
+def grpc_init_kwargs(grpc_channel):
     """Gets the keyword arguments required for creating the gRPC interpreter."""
     channel = grpc.insecure_channel(f"localhost:31763")
     grpc_options = GrpcSessionOptions(
@@ -265,22 +265,22 @@ def grpc_session_creation_kwargs():
     return {"grpc_options": grpc_options}
 
 
-@pytest.fixture(scope="package")
-def library_session_creation_kwargs():
+@pytest.fixture(scope="session")
+def library_init_kwargs():
     """Gets the keyword arguments required for creating the library interpreter."""
     return {}
 
 
 @pytest.fixture(
-    params=("library_session_creation_kwargs", "grpc_session_creation_kwargs"), scope="module"
+    params=("library_init_kwargs", "grpc_init_kwargs"), scope="session"
 )
-def session_creation_kwargs(request):
+def init_kwargs(request):
     """Gets the keyword arguments to create a nidaqmx session."""
     return request.getfixturevalue(request.param)
 
 
 @pytest.fixture(scope="function")
-def task(request, session_creation_kwargs):
+def task(request, init_kwargs):
     """Gets a task instance."""
     # set default values used for the initialization of the task.
     init_args = {
@@ -292,4 +292,5 @@ def task(request, session_creation_kwargs):
         if marker.name in init_args:  # only look at markers with valid argument names
             init_args[marker.name] = marker.args[0]  # assume single parameter in marker
 
-    yield nidaqmx.Task(**init_args, **session_creation_kwargs)
+    with nidaqmx.Task(**init_args, **init_kwargs) as task:
+        yield task
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -253,7 +253,7 @@ def test_assets_directory() -> pathlib.Path:
     return pathlib.Path(__file__).parent / "test_assets"
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="session")
 def grpc_channel():
     """Gets the gRPC channel."""
     with grpc.insecure_channel("localhost:31763") as channel:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -253,15 +253,20 @@ def test_assets_directory() -> pathlib.Path:
     return pathlib.Path(__file__).parent / "test_assets"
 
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -249,3 +251,45 @@ def persisted_channel(request):
 def test_assets_directory() -> pathlib.Path:
     """Gets path to test_assets directory."""
     return pathlib.Path(__file__).parent / "test_assets"
+
+
+@pytest.fixture(scope="package")
+def grpc_session_creation_kwargs():
+    """Gets the keyword arguments required for creating the gRPC interpreter."""
+    channel = grpc.insecure_channel(f"localhost:31763")
```
bkeryan, line 259: Please extract this into a separate `grpc_channel` fixture. 

Is scope=package the right scope? I suspect that scope=package would re-create the fixture for each directory. Maybe we want scope=session.

Use a context manager to ensure that it closes the channel properly. The scope makes this less important, but it's still a good practice. 
```
    with grpc.insecure_channel("localhost:31763") as channel:
        yield channel
```

The correct return type for a fixture that fields a `grpc.Channel` is `Generator[grpc.Channel, None, None]`.

tests/conftest.py
bkeryan, line 1: Are we going to use a fixture to launch the grpc-device server? See https://github.com/ni/nimi-python/blob/master/src/shared/system_test_utilities.py for code that does this.

If we use a fixture to launch the server, I think it should have scope=session. I suspect that scope=package would re-create the fixture for each directory.

tests/conftest.py
```diff
@@ -249,3 +251,45 @@ def persisted_channel(request):
...
+    grpc_options = GrpcSessionOptions(
+        grpc_channel=channel,
+        session_name="",
+    )
+    grpc_options = GrpcSessionOptions(channel, "")
+    return {"grpc_options": grpc_options}
+
+
+@pytest.fixture(scope="package")
+def library_session_creation_kwargs():
```
bkeryan, line 269: Naming: nimi-python calls this `session_creation_kwargs` because it's used when creating a `Session` object. In nidaqmx-python, there is no `Session` object, and these will be used to create a lot of different objects, so consider renaming to something more generic like `init_kwargs`.

tests/conftest.py
```diff
@@ -249,3 +251,45 @@ def persisted_channel(request):
...
+    channel = grpc.insecure_channel(f"localhost:31763")
+    grpc_options = GrpcSessionOptions(
+        grpc_channel=channel,
+        session_name="",
+    )
+    grpc_options = GrpcSessionOptions(channel, "")
+    return {"grpc_options": grpc_options}
+
+
+@pytest.fixture(scope="package")
```
bkeryan, line 268: Scope: as long as the `grpc_channel` fixture has a broad scope, it's not critical for all of the kwargs fixtures to have a broad scope. However, pytest does seem to error if a fixture with a broad scope requests a fixture with a narrow scope, so I guess it's okay for the kwargs fixtures to have a broad scope.

tests/conftest.py
```diff
@@ -249,3 +251,45 @@ def persisted_channel(request):
...
+    init_args = {
+        "new_task_name": "",
+    }
+
+    # iterate through markers and update arguments
+    for marker in request.node.iter_markers():
+        if marker.name in init_args:  # only look at markers with valid argument names
+            init_args[marker.name] = marker.args[0]  # assume single parameter in marker
+
+    yield nidaqmx.Task(**init_args, **session_creation_kwargs)
```
bkeryan, line 295: This does not close the task. Please use `with` to enter the task's context manager:
```
with nidaqmx.Task(**init_args, **session_creation_kwargs) as task:
    yield task
```

bkeryan (changes requested):

tests/conftest.py
subash-suresh: Updated as per suggestion, i have updated the scope to session

tests/conftest.py
subash-suresh: I have renamed the fixtures as per suggestions.

tests/conftest.py
subash-suresh: I have updated to use `with`

tests/conftest.py
subash-suresh: I have created a PR to add this logic - #319

tests/conftest.py
```diff
@@ -253,34 +253,37 @@ def test_assets_directory() -> pathlib.Path:
     return pathlib.Path(__file__).parent / "test_assets"
 
 
-@pytest.fixture(scope="package")
-def grpc_session_creation_kwargs():
+@pytest.fixture(scope="module")
```
bkeryan, line 256: If init_kwargs has scope=session, then all fixtures it depends on (grpc_init_kwargs, grpc_channel) must have scope=session.

pytest prints errors when you get it wrong:

```
___________ ERROR at setup of test__channel__get_boolean_property__returns_default_value[grpc_init_kwargs] ____________
ScopeMismatch: You tried to access the module scoped fixture grpc_channel with a session scoped request object, involved factories:
tests\conftest.py:279:  def init_kwargs(request)
tests\conftest.py:263:  def grpc_init_kwargs(grpc_channel)
tests\conftest.py:256:  def grpc_channel()
```

bkeryan (approved):
Approved with suggestions

tests/conftest.py
subash-suresh: Thanks for the explanation, i missed to push this change, now i have updated the fixture to be scoped to `session`.

PR #317, opened 2023-05-03
[Internal] Running tests in pipeline with pytest

MounikaBattu17:
- This PR runs tests in pipeline using `poetry run pytest`.
- Added the command in `PR.yml`.

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -37,3 +37,5 @@ jobs:
           poetry run python src/codegen --dest generated/nidaqmx
       - name: Check for files dirtied by codegen
         run: git diff --exit-code
+      - name: Run tests
+        run: poetry run pytest


MounikaBattu17:
> We can't run acceptance/component/legacy/system tests in the current pipeline because the container used to run the pipeline does not have NI-DAQmx installed.
> 
> nimi-python solves this problem by running system tests on a self-hosted VM that has the appropriate drivers installed: https://github.com/ni/nimi-python/blob/master/.github/workflows/github_actions_aws_windows_python64.yml
> 
> We should be able to run unit tests in the current pipeline, though. I will try adding them to the pipeline in #313

Sure Thank you Brad. So to run acceptance/component/legacy/system tests in the current pipeline, should we also try creating a self-hosted VM?

bkeryan:
> Sure Thank you Brad. So to run acceptance/component/legacy/system tests in the current pipeline, should we also try creating a self-hosted VM?

Let's discuss offline.

PR #320, opened 2023-05-03
Update assert in test_task_double_close test case

MounikaBattu17:
- Since `_close_on_exit` has been introduced in clause of `__exit__()` in `task.py`, there won't be any `DaqResourceWarning` thrown unless `_close_on_exit` remains `True` on double closure of task.
- So, Updated the test case `test_task_double_close` to validate `_close_on_exit` to `False` and `_handle` to `None`.

### DIFF
--- a/tests/legacy/test_resource_warnings.py
+++ b/tests/legacy/test_resource_warnings.py
@@ -13,9 +13,10 @@ class TestResourceWarnings:
 
     def test_task_double_close(self):
         """Test to validate double closure of tasks."""
-        with pytest.warns(DaqResourceWarning):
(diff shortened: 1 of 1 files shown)


bkeryan (changes requested):
Don't do this. See https://github.com/ni/nidaqmx-python/pull/308#discussion_r1183856166

MounikaBattu17:
With respect to the https://github.com/ni/nidaqmx-python/pull/308#discussion_r1183856166, this change is no longer required as this changes the behavior of throwing warnings on multiple calls on close task. Hence closing this Pull request.

PR #330, opened 2023-05-08
Fix | gRPC interpreter| convert the read arrays to the expected array shape.

subash-suresh:
- The read methods in the gRPC implementation did not return the exact dimensions of the expected output arrays, instead the arrays returned where always flatted to a 1D array. The current implementation tries to directly assign this flattened array to the variable that is expecting an array with a particular dimension. This caused errors like shown below,
`ValueError: could not broadcast input array from shape (20,) into shape (2,10)`

- This PR takes the returned arrays from the gRPC response and reshapes them to the expected dimensions.

### Why should this PR merged?
- Updated logic to reshape the array to the expected dimensions before assigning the values from the gRPC response.

- All the related tests ran successfully.


generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -2269,7 +2269,7 @@ def read_analog_f64(
                 task=task, num_samps_per_chan=num_samps_per_chan,
                 timeout=timeout, fill_mode_raw=fill_mode,
                 array_size_in_samps=read_array.size))
-        read_array[:] = response.read_array
+        read_array[:] = numpy.reshape(response.read_array, read_array.shape)
```
bkeryan, line 2272: Using reshape on the right hand side of the assignment makes an unnecessary copy because it must convert the RHS to a numpy array:
```
>>> import numpy
>>> x=[1,2,3,4]
>>> y = numpy.reshape(x, (2,2))
>>> y
array([[1, 2],
       [3, 4]])
>>> x[0]=10
>>> y
array([[1, 2],
       [3, 4]])
```

The left hand side of the assignment is already a numpy array, so if you use reshape on the LHS then it will create a view object for the existing array and it will not make a copy.
```
>>> import numpy
>>> x = numpy.array([[1,2],[3,4]])
>>> x
array([[1, 2],
       [3, 4]])
>>> numpy.reshape(x, (4,))[:] = [5,6,7,8]
>>> x
array([[5, 6],
       [7, 8]])
```

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
bkeryan: Update: there is a simpler solution, `read_array.flat`
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flat.html

```
>>> x.flat = [10,11,12,13]
>>> x
array([[10, 11],
       [12, 13]])
```

bkeryan: Another problem: `read_array` may not be the same size as `response.read_array`. See https://github.com/ni/nidaqmx-python/blob/5657d118f80426ab130a6f83de4cca7be366b085/generated/nidaqmx/task.py#L617

This does not have much test coverage, but I think it has some (see line 623)
image

`numpy.reshape` raises an exception when this happens:
```
>>> numpy.reshape(x, (6,))
  File "<__array_function__ internals>", line 200, in reshape
  File "D:\dev\nidaqmx-python\.venv\lib\site-packages\numpy\core\fromnumeric.py", line 298, in reshape
    return _wrapfunc(a, 'reshape', newshape, order=order)
  File "D:\dev\nidaqmx-python\.venv\lib\site-packages\numpy\core\fromnumeric.py", line 57, in _wrapfunc
    return bound(*args, **kwds)
ValueError: cannot reshape array of size 4 into shape (6,)
>>> numpy.reshape(x, (2,))
  File "<__array_function__ internals>", line 200, in reshape
  File "D:\dev\nidaqmx-python\.venv\lib\site-packages\numpy\core\fromnumeric.py", line 298, in reshape
    return _wrapfunc(a, 'reshape', newshape, order=order)
  File "D:\dev\nidaqmx-python\.venv\lib\site-packages\numpy\core\fromnumeric.py", line 57, in _wrapfunc
    return bound(*args, **kwds)
ValueError: cannot reshape array of size 4 into shape (2,)
```

`read_array.flat = ...` doesn't raise an exception, but it performs unnecessary assignments: 
```
>>> import numpy
>>> x = numpy.array([[1,2],[3,4]])
>>> x.flat = [5,6,7]
>>> x
array([[5, 6],
       [7, 5]])
>>> x.flat = [10,11]
>>> x
array([[10, 11],
       [10, 11]])
```

`read_array.flat[:len(response.read_array)] = response.read_array` leaves the extra array elements unmodified when `len(read_array) > len(response.read_array)`:
```
>>> x.flat[:2] = [20,21]
>>> x
array([[20, 21],
       [10, 11]])
```

and it ignores the extra array elements when `len(read_array) < len(response.read_array)`:
```
>>> x.flat[:6] = [30,31,32,33,34,35]
>>> x
array([[30, 31],
       [32, 33]])
```

bkeryan: Also: `numpy.copyto` assigns the extra array elements when `len(read_array) > len(response.read_array)` but it raises an exception when `len(read_array) < len(response.read_array)`:
```
>>> numpy.copyto(x, [50,51])
>>> x
array([[50, 51],
       [50, 51]])
>>> numpy.copyto(x, [60,61,62,63,64,65])
  File "<__array_function__ internals>", line 200, in copyto
ValueError: could not broadcast input array from shape (6,) into shape (2,2)
```

Out of these possible solutions, I think think `read_array.flat[:len(response.read_array)] = response.read_array` is the best. To ensure that we aren't losing any array elements, we could put it in a helper function with an assertion or runtime check:
```
def assign_numpy_array(numpy_array, grpc_array):
    assert len(grpc_array) >= numpy_array.size:
    numpy_array.flat[:len(grpc_array)] = grpc_array
```

subash-suresh: I have created a function in utils to do a runtime check before assigning the array, and if the condition fails it ignore the assignment. Would it be better if we threw warnings in such cases?

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
bkeryan: I looked at what errors and warnings DAQmx defines:
- There is no appropriate DAQmx warning number for this case.
- `DAQmxErrors.WRITE_BUFFER_TOO_SMALL` is appropriate for this case.

generated/nidaqmx/utils.py
```diff
@@ -207,6 +207,17 @@ def unflatten_channel_string(channel_names):
...
 
+def assign_numpy_array(numpy_array, grpc_array):
+    """ Assigns the grpc array to the numpy array, while still maintaining the 
+    original shape of the numpy array. 
+    
+    The numpy array assignment is skipped when the length of the grpc array 
+    is greater than the numpy array as this might result in data loss.
+    """
+    grpc_array_length = len(grpc_array)
+    numpy_array_length = len(numpy_array)
```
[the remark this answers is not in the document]
subash-suresh: updated as suggested.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -137,10 +139,23 @@ class GrpcStubInterpreter(BaseInterpreter):
...
+    original shape of the numpy array. 
+    
+    The numpy array assignment is skipped when the length of the grpc array 
+    is greater than the numpy array as this might result in data loss.
+    """
+    grpc_array_size = len(grpc_array)
+    if numpy_array.size <= grpc_array_size:
+        numpy_array.flat[:grpc_array_size] = grpc_array
+    else:
+        raise errors.DaqError(error_codes.DAQmxErrors.READ_BUFFER_TOO_SMALL)
```
bkeryan, line 161: DaqError needs 2 parameters. You need to specify an error message.

Further investigation reveals that https://github.com/ni/nidaqmx-python/blob/96ab6c1679acf02a89999a5124fc8e029620a2ff/generated/nidaqmx/stream_readers.py#L83 has some examples of read errors thrown from Python.

```suggestion
        raise errors.DaqError(
            'Read cannot be performed because the NumPy array passed into '
            'this function is too small. You must pass in a NumPy array of the '
            'correct size based on the number of channels in the task and the '
            'number of samples per channel requested.\n\n'
            'NumPy array size: {}\n'
            'Required size: {}'
            .format(numpy_array.size, grpc_array_size),
            error_codes.DAQmxErrors.READ_BUFFER_TOO_SMALL)
```

bkeryan, line 161: I think this deserves a test case, but it can be in a separate task/PR.

bkeryan (changes requested):

src/codegen/templates/_grpc_interpreter.py.mako
subash-suresh: I have created a task to track this.
[Task 2388178](https://ni.visualstudio.com/DevCentral/_workitems/edit/2388178): Add tests to validate the case where read array size is small.

subash-suresh: updated as per suggestion

subash-suresh:
> Good news: I tested it out and this new check is redundant. All of the functions that accept a NumPy array from the user already check its shape. We should use `assert` in the GrpcStubInterpreter after all.
> 
> ```
> import pprint
> 
> import grpc
> import nidaqmx
> import nidaqmx.grpc_session_options
> import nidaqmx.stream_readers
> import numpy
> 
> pp = pprint.PrettyPrinter(indent=4)
> 
> 
> with grpc.insecure_channel("localhost:31763") as grpc_channel:
>     grpc_options = nidaqmx.grpc_session_options.GrpcSessionOptions(grpc_channel, session_name="")
>     with nidaqmx.Task(grpc_options=grpc_options) as task:
>         task.ai_channels.add_ai_voltage_chan("cDAQ1Mod1/ai0")
> 
>         reader = nidaqmx.stream_readers.AnalogSingleChannelReader(task.in_stream)
>         data = numpy.zeros(2, dtype=numpy.double)
>         reader.read_many_sample(data, number_of_samples_per_channel=2)
>         print(data)
> 
>         reader.read_many_sample(data, number_of_samples_per_channel=3)
>         print(data)
> ```
> 
> ```
> PS D:\dev\nidaqmx-python> poetry run python .\examples\ai_voltage_sw_timed.py
> [-0.07476921  0.03997864]
> Traceback (most recent call last):
>   File "D:\dev\nidaqmx-python\examples\ai_voltage_sw_timed.py", line 22, in <module>
>     reader.read_many_sample(data, number_of_samples_per_channel=3)
>   File "D:\dev\nidaqmx-python\generated\nidaqmx\stream_readers.py", line 208, in read_many_sample
>     self._verify_array(data, number_of_samples_per_channel, False, True)
>   File "D:\dev\nidaqmx-python\generated\nidaqmx\stream_readers.py", line 83, in _verify_array
>     raise DaqError(
> nidaqmx.errors.DaqError: Read cannot be performed because the NumPy array passed into this function is not shaped correctly. You must pass in a NumPy array of the correct shape based on the number of channels in task and the number of samples per channel requested.
> 
> Shape of NumPy Array provided: (2,)
> Shape of NumPy Array required: (3,)
> 
> Task Name: _unnamedTask<2>
> ```

I have updated the logic in gRPC interpreters to just do a simple assert logic.
This `assert` most probably won't be hit, since the error is already thrown from the gRPC call. Is this just an additional check just in case?

bkeryan:
> I have updated the logic in gRPC interpreters to just do a simple assert logic. This `assert` most probably won't be hit, since the error is already thrown from the gRPC call. Is this just an additional check just in case?

Yes. Outside of tests, asserts are for conditions that should never happen.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -1,10 +1,12 @@
 # Do not edit this file; it was automatically generated.
 
 import grpc
+import nidaqmx
 import typing
 import warnings
 
 from . import errors as errors
+from . import error_codes
 from nidaqmx._base_interpreter import BaseInterpreter
```
bkeryan, line 10: These new imports are not used, so delete them.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3382,3 +3384,14 @@ def write_to_teds_from_file(
             grpc_types.WriteToTEDSFromFileRequest(
                 physical_channel=physical_channel, file_path=file_path,
                 basic_teds_options_raw=basic_teds_options))
+
+def _assign_numpy_array(numpy_array, grpc_array):
+    """ Assigns the grpc array to the numpy array, while still maintaining the 
+    original shape of the numpy array. 
+    
+    The numpy array assignment is skipped when the length of the grpc array 
+    is greater than the numpy array as this might result in data loss.
```
bkeryan, line 3393: Rerun the codegen script.

https://github.com/ni/nidaqmx-python/actions/runs/4936084221/jobs/8823191714?pr=330
```
Run git diff --exit-code
diff --git a/generated/nidaqmx/_grpc_interpreter.py b/generated/nidaqmx/_grpc_interpreter.py
index 5b9d798..d8253b2 [1](https://github.com/ni/nidaqmx-python/actions/runs/4936084221/jobs/8823191714?pr=330#step:9:1)00644
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -3388,9 +3388,6 @@ class GrpcStubInterpreter(BaseInterpreter):
 def _assign_numpy_array(numpy_array, grpc_array):
     """ Assigns the grpc array to the numpy array, while still maintaining the 
     original shape of the numpy array. 
-    
-    The numpy array assignment is skipped when the length of the grpc array 
-    is greater than the numpy array as this might result in data loss.
     """
     grpc_array_size = len(grpc_array)
     assert numpy_array.size >= grpc_array_size
```

bkeryan (approved):
Approved with suggestions

generated/nidaqmx/_grpc_interpreter.py
subash-suresh: deleted the unused imports

generated/nidaqmx/_grpc_interpreter.py
subash-suresh: regenerated the files properly

PR #333, opened 2023-05-09
[Internal] Make attribute functions c calls thread safe in library interpreter

MounikaBattu17:
The get/set functions generic cfunc call is not thread safe since multiple datatypes function calls the same cfunc with different argtypes which may lead to argtypes swap while working with multiple threads. 
Following the issue with thread safety [Bug 2377746](https://dev.azure.com/ni/DevCentral/_workitems/edit/2377746): Update the library interpreter c function calls to be thread-safe, These changes are made on attribute `get/set` functions to make the c calls thread safe in `_library_interpreter.py`

- Added a enum `AttributeFunctionType` to determine whether the function is a `get/set/reset/none` function.

**GET functions:**

- The `value` parameter is instantiated to its corresponding c datatype.
`value = ctypes.c_uint32()`
- The Variadic arguments are not added to argument list
image
- The numpy array types are type casted to void pointers
`value.ctypes.data_as(ctypes.c_void_p)`

**SET functions:**

- The value parameter type is explicitly converted to the required datatype
     - `ctypes.c_double(value)` -> for int, float, double, bool respective types
     - `value.encode('ascii')` -> for string types
     - `value.ctypes.data_as(ctypes.c_void_p)` -> for numpy array types

**Why should this PR be merged?**
Implements the solution for thread safety issue pointed in [Bug 2377746](https://dev.azure.com/ni/DevCentral/_workitems/edit/2377746): Update the library interpreter c function calls to be thread-safe

**What testing has been done?**
Ran `poetry run pytest`. All non-skippable tests passed.


generated/nidaqmx/_library_interpreter.py
```diff
@@ -2332,8 +2332,7 @@ def get_buffer_attribute_uint32(self, task, attribute):
         cfunc = lib_importer.cdll.DAQmxGetBufferAttribute
         with cfunc.arglock:
             cfunc.argtypes = [
-                lib_importer.task_handle, ctypes.c_int32,
-                ctypes.POINTER(ctypes.c_uint32)]
+                lib_importer.task_handle, ctypes.c_int32]
```
bkeryan, line 2335: You also need to update the template.

Attribute get/set/reset functions should use double-checked locking like normal functions:
```
        if cfunc.argtypes is None:
            with cfunc.arglock:
                if cfunc.argtypes is None:
                    cfunc.argtypes = [...]
```

Non-attribute varargs functions like `get_analog_power_up_states` should still hold the lock when assigning `cfunc.argtypes` and calling `cfunc`, as they do now. I guess those are in a different template, though.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2462,20 +2457,20 @@ def get_chan_attribute_int32(self, task, channel, attribute):
...
+
         cfunc = lib_importer.cdll.DAQmxGetChanAttribute
         with cfunc.arglock:
             cfunc.argtypes = [
-                lib_importer.task_handle, ctypes_byte_str, ctypes.c_int32,
-                ctypes.c_char_p, ctypes.c_uint]
+                lib_importer.task_handle, ctypes_byte_str, ctypes.c_int32]
 
         temp_size = 0
         while True:
```
bkeryan, line 2476: Double-call functions already set `value` to a string buffer. This added assignment is incorrect for two reasons:
- It sets `value` to refer to a type, not a value, because you omitted the parens in `instantiation_lines.append(f"{param.parameter_name} = {param.ctypes_data_type}")`
- It's unnecessary. 

Please remove the added assignment. Maybe the special case needs to exclude strings.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -2410,8 +2408,7 @@ def get_chan_attribute_bool(self, task, channel, attribute):
         cfunc = lib_importer.cdll.DAQmxGetChanAttribute
         with cfunc.arglock:
             cfunc.argtypes = [
-                lib_importer.task_handle, ctypes_byte_str, ctypes.c_int32,
-                ctypes.POINTER(c_bool32)]
+                lib_importer.task_handle, ctypes_byte_str, ctypes.c_int32]
```
bkeryan, line 2411: Duplicating the argtypes setup for one C function between multiple Python methods makes it possible for the Python methods to set the C function's argtypes inconsistently. 

It would be possible to ensure that this never happens by refactoring the codegen to generate shared helper methods. However, I don't think it's critical to do this refactor, as long as we have test coverage for all of the Python get/set/reset methods.

Looking at the test coverage for `_library_interpreter.py`, there are some gaps. (To view the test coverage, run `poetry run pytest -v --cov generated/nidaqmx` and `poetry run coverage html`, then open `htmlcov/index.html`.)

Addressing this is outside the scope of this PR, so please create a task.

bkeryan (changes requested):

generated/nidaqmx/_library_interpreter.py
MounikaBattu17: Sure Thanks for providing the commands for running test coverage. 
I have created a task **[Task 2388233](https://dev.azure.com/ni/DevCentral/_workitems/edit/2388233): Increase test coverage for _library_interpreter.py** to cover the test coverage for _library_interpreter.py. 
Should this task scope for only get/set/reset functions or is it better to scope it for coverage of library interpreter on whole?

generated/nidaqmx/_library_interpreter.py
MounikaBattu17: I have removed the assignment for string types.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -459,3 +489,19 @@ def get_read_array_parameters(func):
...
+def type_cast_attribute_set_function_parameter(ctype: str, parameter_name: str):
+    """Type casting of attribute set parameter during c call."""
+    if ctype == "ctypes.c_char_p":
+        return f"{parameter_name}.encode('ascii')"
+    if is_numpy_array_datatype(ctype):
+        return f"{parameter_name}.ctypes.data_as(ctypes.c_void_p)"
+    return f"{ctype}({parameter_name})"
+
+
+def is_numpy_array_datatype(ctypes_data_type: str):
```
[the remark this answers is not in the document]
MounikaBattu17: Passed in `param` instead of `ctypes_data_type`/`parameter_name`.

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
```diff
@@ -1,12 +1,12 @@
...
 <%
-    from codegen.utilities.interpreter_helpers import generate_interpreter_function_call_args, get_samps_per_chan_read_or_write_param, get_argument_types
+    from codegen.utilities.interpreter_helpers import generate_interpreter_function_call_args, get_samps_per_chan_read_or_write_param, get_argument_types, is_attribute_function
     from codegen.utilities.text_wrappers import wrap, docstring_wrap
 %>\
         cfunc = lib_importer.${'windll' if function.calling_convention == 'StdCall' else 'cdll'}.DAQmx${function.c_function_name}
 \
 ## Create argument ctypes types list.
-%if function.calling_convention == 'StdCall':
+%if function.calling_convention == 'StdCall' or is_attribute_function(function):
```
bkeryan, line 9: Now there are only two functions that use the else case:
- `get_analog_power_up_states_with_output_type`
- `set_analog_power_up_states_with_output_type`

(How do I know? I edited the template and added a comment to the else case.)

These functions should also use double-checked locking. Delete the if-statement and its else case.

src/codegen/templates/library_interpreter/double_c_function_call.py.mako
```diff
@@ -1,13 +1,13 @@
...
-    from codegen.utilities.interpreter_helpers import get_argument_types, generate_interpreter_function_call_args, get_output_param_with_ivi_dance_mechanism, get_output_params 
+    from codegen.utilities.interpreter_helpers import get_argument_types, generate_interpreter_function_call_args, get_output_param_with_ivi_dance_mechanism, get_output_params, is_attribute_function 
     from codegen.utilities.function_helpers import instantiate_explicit_output_param
     from codegen.utilities.text_wrappers import wrap, docstring_wrap
 %>\
         cfunc = lib_importer.${'windll' if function.calling_convention == 'StdCall' else 'cdll'}.DAQmx${function.c_function_name}
 \
 ## Create argument ctypes types list.
-%if function.calling_convention == 'StdCall':
+%if function.calling_convention == 'StdCall' or is_attribute_function(function):
```
bkeryan, line 10: Now there are zero functions that use the else case. Delete the if-statement and its else case.

generated/nidaqmx/_library_interpreter.py
bkeryan: I think we should evaluate the scope for the interpreter as a whole, but that doesn't necessarily mean 100% coverage. We can discuss it more later.

bkeryan (changes requested):
The attribute functions look good. 

The else-case in the template is no longer necessary, so please delete it.

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
MounikaBattu17: I have removed the if/else blocks.

src/codegen/templates/library_interpreter/double_c_function_call.py.mako
MounikaBattu17: I have removed the if/else blocks.

bkeryan (approved):

generated/nidaqmx/_library_interpreter.py
keer1111: @bkeryan , adding coverage for interpreter would require adding test cases for almost all of the nidaqmx which we did not plan when putting down the HLD. We are getting closer to the dev exit, do you think the new test cases addition are in the critical path of the DAQmx epic? 

Could we have this work item as tech debt out of the Epic scope. What do you think? We could also take this discussion over teams.

PR #334, opened 2023-05-09
Update init file to import grpc_session_options

samchris007:
Updates **init.py** file to import `grpc_session_options`.

**Why should this PR merged?**
Fixes [Bug 2387046](https://dev.azure.com/ni/DevCentral/_workitems/edit/2387046): nidaqmx-python should alias gRPC types into nidaqmx module

**What testing has been done?**
Manually generated and verified out files.

### DIFF
--- a/generated/nidaqmx/__init__.py
+++ b/generated/nidaqmx/__init__.py
@@ -1,4 +1,3 @@
-from nidaqmx.constants import *
 from nidaqmx.errors import DaqError, DaqReadError, DaqWriteError, DaqWarning, DaqResourceWarning
 from nidaqmx.grpc_session_options import *
 from nidaqmx.scale import Scale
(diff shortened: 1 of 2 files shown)


generated/nidaqmx/__init__.py
```diff
@@ -1,4 +1,6 @@
+from nidaqmx.constants import *
```
bkeryan, line 1: Please revert this. AB#2387046 is only about gRPC types, not nidaqmx.constants.

bkeryan (changes requested):

generated/nidaqmx/__init__.py
samchris007: Sure. Reverted.

bkeryan (approved):
Don't forget to update the PR title and PR description

PR #336, opened 2023-05-09
Add markers for running tests with only library or grpc interpreters

subash-suresh:

The existing implementation of `init_kwargs` fixture always runs for both the `library_init_kwargs` and `grpc_init_kwargs` implementations.  This PR adds 2 new markers namely `library_only` and `grpc_only`, which can be used to restrict the test to only run for any one of the required implementations.

- Added the [`pytest_generate_tests` function](https://docs.pytest.org/en/6.2.x/parametrize.html#basic-pytest-generate-tests-example), which is a hook in pytest, which is called during test collection phase. Here based on the `library_only` and `grpc_only` markers the parameters for the `init_kwargs` fixture is passed.
- The `test_hash_operations` test only can be tested with a task instance created through the c functions calls (libaray interpreter), since the hash value cannot be obtained for a grpc task instance. Therefore, this test has been marked to work with only library interpreter implementation with the use of `library_only` marker.
- Added the known markers to `pyproject.toml` file to avoid any pytest warning related to unknown markers.

- The `test_hash_operations` marked with `library_only` marker successfully skipped the grpc test implementation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,6 @@ extend_exclude = "docs,generated,src/codegen/metadata,src/codegen/templates,src/
 addopts = "--doctest-modules --strict-markers"
 filterwarnings = ["always::ImportWarning", "always::ResourceWarning"]
 testpaths = ["tests"]
-filterwarnings = ["always::ResourceWarning"]
 markers = [
   # Defines custom markers used by nidcpower system tests. Prevents PytestUnknownMarkWarning.
   "library_only: tests only the libaray interpreter implementation.",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,8 +82,8 @@ addopts = "--doctest-modules --strict-markers"
 filterwarnings = ["always::ImportWarning", "always::ResourceWarning"]
 testpaths = ["tests"]
 markers = [
-  # Defines custom markers used by nidcpower system tests. Prevents PytestUnknownMarkWarning.
-  "library_only: tests only the libaray interpreter implementation.",
+  # Defines custom markers used by nidaqmx tests. Prevents PytestUnknownMarkWarning.
+  "library_only: tests only the library interpreter implementation.",
   "grpc_only: tests only the grpc interpreter implementation.",
   "new_task_name: name of the new task to be created."
 ]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -81,6 +81,12 @@ extend_exclude = "docs,generated,src/codegen/metadata,src/codegen/templates,src/
 addopts = "--doctest-modules --strict-markers"
 filterwarnings = ["always::ImportWarning", "always::ResourceWarning"]
 testpaths = ["tests"]
+markers = [
+  # Defines custom markers used by nidcpower system tests. Prevents PytestUnknownMarkWarning.
```
bkeryan, line 85:
```suggestion
  # Defines custom markers used by nidaqmx system tests. Prevents PytestUnknownMarkWarning.
```

pyproject.toml
```diff
@@ -81,6 +81,12 @@ extend_exclude = "docs,generated,src/codegen/metadata,src/codegen/templates,src/
...
+  "library_only: tests only the libaray interpreter implementation.",
```
bkeryan, line 86: spelling
```suggestion
  "library_only: tests only the library interpreter implementation.",
```

tests/legacy/test_container_operations.py
```diff
@@ -81,6 +81,7 @@ def test_equality_operations(self, task, any_x_series_device, seed):
         assert ai_channel_1 != ai_channel_2
 
     @pytest.mark.parametrize("seed", [generate_random_seed()])
+    @pytest.mark.library_only
     def test_hash_operations(self, generate_task, any_x_series_device, seed):
```
bkeryan, line 85: > The test_hash_operations test only can be tested with a task instance created through the c functions calls (libaray interpreter), since the hash value cannot be obtained for a grpc task instance. Therefore, this test has been marked to work with only library interpreter implementation with the use of library_only marker.

gRPC tasks should support the same operations as library tasks. Please file a bug and add a comment with a link/reference to the bug.

bkeryan (approved):
Approved with suggestions

pyproject.toml
subash-suresh: fixed the typo

pyproject.toml
subash-suresh: fixed the typo

tests/legacy/test_container_operations.py
subash-suresh: I have added a comment to the test case related to the bug https://ni.visualstudio.com/DevCentral/_workitems/edit/2388532

PR #342, opened 2023-05-11
Handle event functions | PR2 | filter out the parameters that are not used in proto for grpc interpreters

subash-suresh:

The existing implementations allow parameters not used in the proto definition to be used in the gRPC function calls. This leads to the addition of unwanted parameters while making the gRPC calls. This PR updates the logic to filter out parameters not used in gRPC in the gRPC interpreter.

Additionally, this PR also handles a special case, where there are two parameters with similar name in the metadata. The assumption here is that, one parameter data would contain information on the c function calls and other parameter data would contain the information on the proto definition and would likely be marked as a `proto-only` parameter. The current logic filters out `proto-only` parameters by default, but the filtering of parameters with `include_in_proto` tag, would also filter out the other parameter with the same name as well. Hence avoiding both parameters when defining args for the grpc request.

To avoid this, in cases where there are multiple parameters with the same name, the proto-only parameter is also considered as an interpreter parameter when using grpc interpreters.
This scenario can only be currently observed in [GetDigitalPowerUpStates](https://github.com/ni/nidaqmx-python/blob/a844b74473ced34ad8b23de4ef608e87effc3a1e/src/codegen/metadata/functions.py#LL13730C6-L13730C29) and [GetDigitalPullUpPullDownStates](https://github.com/ni/nidaqmx-python/blob/a844b74473ced34ad8b23de4ef608e87effc3a1e/src/codegen/metadata/functions.py#LL13780C6-L13780C36)

- This PR fixes the invalid RPC calls with unwanted parameters, by filtering out the parameters using the `include_in_proto` metadata.
- The logic has been updated to allow functions with `reapeated_varargs` as True, in the interpreter parameters list if the interpreter used is gRPC interpreter.
- The logic for getting function signature in the grpc interpreter mako has to updated to not meht to be similar to the base and library interpreter mako files.

The code generated removes the unnecessary parameters from the gRPC requests in the gRPC interpreter file.


zhindes (approved):

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -469,11 +472,27 @@ def get_interpreter_parameters(func, is_grpc_interpreter=False):
...
 
 
+def _get_repeated_params(params):
+    """Gets the names of the parameters which has another parameter with a similar name in the same function
+    
+    In these scenarios, usually one parameter contains the data for the c function calls 
+    and the other parameter contains the data required for the grpc calls and proto definitions.
+    """
+    parameter_names = [param.parameter_name for param in params]
+    return [param_name for param_name in parameter_names if parameter_names.count(param_name) > 1]
```
bkeryan, line 493: This is described by the metadata. The C parameters have `repeating_argument: True` and the gRPC parameters have `repeated_var_args: True`.
https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/functions.py#L13730

The gRPC parameters identified by this function are filtered out by the code on line 466:
```
        if (
            (
                parameter.is_used_in_python_api
                and not parameter.is_proto_only
                and not parameter.repeated_var_args
            )
```

Instead of second-guessing the metadata, use it to avoid filtering these parameters out in the first place.
```
        if (
            (
                parameter.is_used_in_python_api
                and not parameter.is_proto_only
                and (not parameter.repeated_var_args or is_grpc_interpreter)
            )
```

bkeryan (changes requested):

src/codegen/utilities/interpreter_helpers.py
bkeryan: Something else is bothering me about this.

The interpreter method signature should not be different for gRPC. LibraryInterpreter and GrpcStubInterpreter should use the same signature as BaseInterpreter.

BaseInterpreter and LibraryInterpreter agree about how to generate the interpreter method signature:
image

BaseInterpreter and GrpcStubInterpreter disagree about whether to pass is_grpc_interpreter=True to `get_params_for_function_signature`:
image

`_grpc_interpreter.py.mako` should call `params = get_params_for_function_signature(func)` without is_grpc_interpreter=True. This will ensure that any gRPC-specific parameters don't affect the interpreter method signature.

subash-suresh: Updated as per comments

bkeryan (approved):

PR #347, opened 2023-05-12
[RFC] Event lifetime prototype

bkeryan:

Prototype a solution to AB#2390580: Unregistering events crashes when sharing interpreters

When `interpreter.register_x_event()` registers the event, it returns a context manager that controls the lifetime of the event registration resources (ctypes pointer, gRPC response stream & thread).

The task uses an `ExitStack` to track these context managers and attempts to clean up at the right time.

`task.close()` aborts the task if there are any registered event handlers.
- This is prevents `GrpcStubInterpreter`'s thread from calling the user's event callback after the task is cleared. 
- For `LibraryInterpreter`, I originally did not abort the task; instead, I called `clear_task` before `stack.close()`. However, the docs for [DAQmxRegisterEveryNSamplesEvent](https://www.ni.com/docs/en-US/bundle/ni-daqmx-c-api-ref/page/daqmxcfunc/daqmxregistereverynsamplesevent.html) explicitly say that the callback may be called after the task is cleared. I think I saw this happen once?

This prototype also updates the event callbacks to pass the user a `nidaqmx.Task` and `nidaqmx.EveryNSamplesEventType`. We need to discuss the compatibility implications of this.

This prototype also updates the `LibraryInterpreter` to use double-checked locking for the event functions. We shouldn't need to change `cfunc.argtypes` to pass a null pointer.

Prototype, do not merge


generated/nidaqmx/task.py
```diff
@@ -343,7 +341,22 @@ def close(self):
...
+            # event thread, which ensures that the thread doesn't call event callbacks.
+            if self._every_n_acquired_event_stack is not None:
+                self._every_n_acquired_event_stack.close()
+                self._every_n_acquired_event_stack = None
+
+            if self._done_event_stack is not None:
+                self._done_event_stack.close()
+                self._done_event_stack = None
+
         self._interpreter.clear_task(self._handle)
```
bkeryan, line 361: It's probably better to split cancel vs. close so that we can cancel events, clear task, and close events.

generated/nidaqmx/task.py
```diff
@@ -649,7 +662,22 @@ def register_done_event(self, callback_method):
                 Passing None for this parameter unregisters the event callback
                 function.
         """
-        self._interpreter.register_done_event(self._handle, 0, callback_method, None)
+        if callback_method is not None:
+            # Pass the user a Task object instead of a task handle
+            def call_done_event_callback(task_handle, status, callback_data):
+                return callback_method(self, status, callback_data)
```
bkeryan, line 668: It would also be good to remove the callback_data parameter and the need to return an int.

bkeryan: I prototyped more signature changes, with a `use_new_signature=False` parameter to opt in

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -2579,24 +2594,84 @@ def read_raw(self, task, num_samps_per_chan, timeout, read_array):
...
+                        response.n_samples,
+                        callback_data)
+            except grpc.RpcError as e:
+                if e.code() == grpc.StatusCode.CANCELLED:
+                    return
+                raise
+            except errors.RpcError as e:
+                if e.rpc_code == grpc.StatusCode.CANCELLED:
+                    return
+                raise
```
zhindes, line 2670: This is identical to `done_event_thread` except for the args to the callback_function. Making it generic is pretty easy except that you pull different things out of the `response`. Once you generalize that, then its probably not all that useful to share anymore... ah well

zhindes (commented):
this looks really nice, definitely a huge improvement

generated/nidaqmx/_grpc_interpreter.py
bkeryan: This duplication is also in https://github.com/ni/nidaqmx-python/pull/343#discussion_r1194248172

I posted some suggestions for refactoring the exception handling:

> Consider adding a private helper function to `_grpc_interpreter.py.mako` like `_is_canceled(ex: Exception) -> bool` or `_get_rpc_code(ex: Exception) -> Optional[grpc.StatusCode]`. I included the type annotations because they communicate what the function does.

It's also possible to write a helper function that takes a callable with no arguments, or a helper function that takes a callable that takes a response.

generated/nidaqmx/task.py
```diff
@@ -685,8 +726,32 @@ def register_every_n_samples_acquired_into_buffer_event(
...
                 function.
         """
-        self._interpreter.register_every_n_samples_event(self._handle, EveryNSamplesEventType.ACQUIRED_INTO_BUFFER.value,
-                sample_interval, 0, callback_method, None)
+        if callback_method is not None:
+            if use_new_signature:
+                # New signature: callback(task, every_n_samples_event_type, number_of_samples) -> None
+                def call_every_n_samples_event_callback(task_handle, every_n_samples_event_type, number_of_samples, callback_data):
+                    callback_method(self, EveryNSamplesEventType(every_n_samples_event_type), number_of_samples)
+                    return 0
```
bkeryan, line 734: Idea: use [`inspect.signature`](https://docs.python.org/3/library/inspect.html#inspect.signature) to detect whether the callback has the old or new signature, based on the number of `POSITIONAL_OR_KEYWORD` arguments.

PR #350, opened 2023-05-15
Test Failures | TypeError: Value must be iterable fix

MounikaBattu17:

When single sample is passed to a single channel and is converted to `numpy.ndarray` as `data = numpy.asarray(data, dtype=bool)`, `TypeError: Value must be iterable fix` is thrown from grpc tests which are using this conversion. So to fix this

- Added a if block in `task.py` to check `number_of_samples_per_channel/number_of_channels == 1`, then change the type of data from `int/float/bool..` to `list`.

This PR fixes the error **TypeError: Value must be iterable fix**

Ran `poetry run pytest`

### DIFF
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -3345,7 +3345,8 @@ class GrpcStubInterpreter(BaseInterpreter):
             grpc_types.WriteDigitalLinesRequest(
                 task=task, num_samps_per_chan=num_samps_per_chan,
                 auto_start=auto_start, timeout=timeout,
-                data_layout_raw=data_layout, write_array=bytes(write_array)))
+                data_layout_raw=data_layout,
+                write_array=write_array.tobytes()))
         return response.samps_per_chan_written
 
     def write_digital_scalar_u32(self, task, auto_start, timeout, value):
@@ -3388,7 +3389,8 @@ class GrpcStubInterpreter(BaseInterpreter):
             grpc_types.WriteDigitalU8Request(
                 task=task, num_samps_per_chan=num_samps_per_chan,
                 auto_start=auto_start, timeout=timeout,
-                data_layout_raw=data_layout, write_array=bytes(write_array)))
+                data_layout_raw=data_layout,
+                write_array=write_array.tobytes()))
         return response.samps_per_chan_written
 
     def write_raw(self, task, num_samps, auto_start, timeout, write_array):
@@ -3397,7 +3399,7 @@ class GrpcStubInterpreter(BaseInterpreter):
             self._client.WriteRaw,
             grpc_types.WriteRawRequest(
                 task=task, num_samps=num_samps, auto_start=auto_start,
-                timeout=timeout, write_array=bytes(write_array)))
+                timeout=timeout, write_array=write_array.tobytes()))
         return response.samps_per_chan_written
 
     def write_to_teds_from_array(
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -3212,7 +3212,7 @@ class GrpcStubInterpreter(BaseInterpreter):
             grpc_types.WriteAnalogF64Request(
                 task=task, num_samps_per_chan=num_samps_per_chan,
                 auto_start=auto_start, timeout=timeout,
-                data_layout_raw=data_layout, write_array=write_array))
+                data_layout_raw=data_layout, write_array=write_array.flat))
         return response.samps_per_chan_written
 
     def write_analog_scalar_f64(self, task, auto_start, timeout, value):
@@ -3231,7 +3231,7 @@ class GrpcStubInterpreter(BaseInterpreter):
             grpc_types.WriteBinaryI16Request(
                 task=task, num_samps_per_chan=num_samps_per_chan,
                 auto_start=auto_start, timeout=timeout,
-                data_layout_raw=data_layout, write_array=write_array))
+                data_layout_raw=data_layout, write_array=write_array.flat))
         return response.samps_per_chan_written
 
     def write_binary_i32(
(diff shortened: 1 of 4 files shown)


src/handwritten/task.py
```diff
@@ -1007,6 +1007,8 @@ def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
 
         # Analog Input
```
zhindes, line 1008: This comment is wrong, and also not useful. The if statement is self-documenting. Just delete it and any other similar ones in this function.

src/handwritten/task.py
```diff
@@ -1007,6 +1007,8 @@ def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
...
         if write_chan_type == ChannelType.ANALOG_OUTPUT:
+            if number_of_samples_per_channel == 1  and number_of_channels == 1:
+                data = [data]
```
zhindes, line 1011: I don't think this is right... if someone passes in a 1-element array, this would make it an array of array. Looking at the code block above this, I think you might mean `data = [element]`?

zhindes (changes requested):

bkeryan:
> This PR fixes the error TypeError: Value must be iterable fix

@MounikaBattu17 Does the error contain the word "fix" or is that a typo?

Which test is failing and what is the error output? Has a bug been filed about this?

bkeryan (changes requested):

MounikaBattu17:
> > This PR fixes the error TypeError: Value must be iterable fix
> 
> @MounikaBattu17 Does the error contain the word "fix" or is that a typo?

No, the error is just **TypeError: Value must be iterable**

> Which test is failing and what is the error output? Has a bug been filed about this?

These are the 3 tests failing with this error
```
FAILED tests/legacy/test_read_write.py::TestAnalogReadWrite::test_1_chan_1_samp[grpc_init_kwargs-5147014139623261137] - TypeError: Value must be iterable
FAILED tests/legacy/test_read_write.py::TestDigitalReadWrite::test_uint_1_chan_1_samp[grpc_init_kwargs-1232673557569935366] - TypeError: Value must be iterable
FAILED tests/legacy/test_read_write.py::TestDigitalReadWrite::test_uint_multi_port[grpc_init_kwargs-5221339145220964105] - TypeError: Value must be iterable
```
My understanding is that for a single channel & single sample the data is passed as `int/float..` and not as `list`, since this is single value the error is thrown. I observed in `stream_writers.py` that in this scenario, the list of data is converted to numpy array rather the direct int/float value.
image

So, I took reference from here. But now I see the `write_array.flat` (which is basically a 1D iteration which I'm also trying to achieve). So I'll revert these changes & Update with `write_array.flat` changes.


Bug is not created but it is tracked in this [Task 2390707](https://dev.azure.com/ni/DevCentral/_workitems/edit/2390707): Fix test failure: "Value is not iterable"

src/handwritten/task.py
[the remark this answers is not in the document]
MounikaBattu17: Yes! Thank you @bkeryan. I was looking into the  AB#2393319 & observed the fix for AB#2393319 also fixes the issue in this PR.

MounikaBattu17: As Brad pointed out, the change has to happen in grpc interpreter and not in `task.py`. I'll revert these changes and update with the correct solution.

src/handwritten/task.py
MounikaBattu17: Removed unnecessary comments.

zhindes (approved):

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3376,7 +3376,7 @@ def write_digital_u32(
             grpc_types.WriteDigitalU32Request(
                 task=task, num_samps_per_chan=num_samps_per_chan,
                 auto_start=auto_start, timeout=timeout,
-                data_layout_raw=data_layout, write_array=write_array))
+                data_layout_raw=data_layout, write_array=write_array.flat))
         return response.samps_per_chan_written
 
     def write_digital_u8(
```
bkeryan, line 3382: bytes() doesn't handle this case correctly for uint8.
```
>>> import numpy
>>> x = numpy.asarray(True, dtype=bool)
>>> bytes(x)
b'\x01'
>>> y = numpy.asarray(False, dtype=bool)
>>> bytes(y)
b'\x00'
>>> z = numpy.asarray(0xa5, dtype=numpy.uint8)
>>> bytes(z)
b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

It's interpreting the value 0xa5 (165) as a size:
```
>>> len(bytes(z))
165
```

`x.flat` works, but it reports DeprecationWarning for bool.
```
>>> bytes(x.flat)
<stdin>:1: DeprecationWarning: In future, it will be an error for 'np.bool_' scalars to be interpreted as an index
b'\x01'
>>> bytes(y.flat)
b'\x00'
>>> bytes(z.flat)
b'\xa5'
```

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
MounikaBattu17: Thank you! I have changed to `tobytes()` and verified by running tests for any discrepancies. All expected tests passed :)

bkeryan (approved):

PR #354, opened 2023-05-16
tests: Add component tests for events

bkeryan:

Add component tests for the event callback functions.

The current test coverage is inadequate.

Ran new tests on my test machine.
Ran an earlier version of tests on a test machine with a physical X Series installed.

### DIFF
--- a/tests/component/test_task_events.py
+++ b/tests/component/test_task_events.py
@@ -1,0 +1,310 @@
+import pytest
+
+import nidaqmx
+from nidaqmx.constants import EveryNSamplesEventType, Signal
+from nidaqmx.error_codes import DAQmxErrors
+from tests._event_utils import (
+    DoneEventObserver,
+    EveryNSamplesEventObserver,
+    SignalEventObserver,
+)
+
+
+@pytest.fixture
+def ai_task(task: nidaqmx.Task, any_x_series_device: nidaqmx.system.Device) -> nidaqmx.Task:
+    task.ai_channels.add_ai_voltage_chan(any_x_series_device.ai_physical_chans[0].name)
+    return task
+
+
+@pytest.fixture
+def ai_task_with_real_device(
+    task: nidaqmx.Task, real_x_series_device: nidaqmx.system.Device
+) -> nidaqmx.Task:
+    task.ai_channels.add_ai_voltage_chan(real_x_series_device.ai_physical_chans[0].name)
+    return task
+
+
+@pytest.fixture
+def ao_task(task: nidaqmx.Task, any_x_series_device: nidaqmx.system.Device) -> nidaqmx.Task:
+    task.ao_channels.add_ao_voltage_chan(any_x_series_device.ao_physical_chans[0].name)
+    return task
+
+
+@pytest.mark.library_only
+def test___done_event_registered___run_finite_acquisition___callback_invoked_once_with_success_status(
+    ai_task: nidaqmx.Task,
+) -> None:
+    event_observer = DoneEventObserver()
+    ai_task.register_done_event(event_observer.handle_done_event)
+    ai_task.timing.cfg_samp_clk_timing(rate=10000.0, samps_per_chan=1000)
+
+    ai_task.start()
+
+    event_observer.wait_for_events()
+    with pytest.raises(TimeoutError):
+        event_observer.wait_for_events(timeout=100e-3)
+    assert len(event_observer.events) == 1
+    assert all(e.status == 0 for e in event_observer.events)
+
+
+@pytest.mark.library_only
+def test___every_n_samples_event_registered___run_finite_acquisition___callback_invoked_n_times_with_type_and_num_samples(
+    ai_task: nidaqmx.Task,
+) -> None:
+    event_observer = EveryNSamplesEventObserver()
+    ai_task.register_every_n_samples_acquired_into_buffer_event(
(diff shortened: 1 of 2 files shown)


tests/component/test_task_events.py
```diff
@@ -0,0 +1,310 @@
...
+    return task
+
+
+@pytest.mark.library_only
+def test___done_event_registered___run_finite_acquisition___callback_invoked_once_with_success_status(
+    ai_task: nidaqmx.Task,
+) -> None:
+    event_observer = DoneEventObserver()
+    ai_task.register_done_event(event_observer.handle_done_event)
+    ai_task.timing.cfg_samp_clk_timing(rate=10000.0, samps_per_chan=1000)
```
zhindes, line 39: I guess finite is the default, but being specific would be nice.

zhindes (approved):

tests/component/test_task_events.py
```diff
@@ -0,0 +1,414 @@
...
+        ai_task.start()
+        done_event_observer.wait_for_events()
+        every_n_samples_event_observers[i].wait_for_events(every_n_samples_event_counts[i])
+        ai_task.stop()
+        ai_task.register_every_n_samples_acquired_into_buffer_event(100, None)
+
+    assert len(done_event_observer.events) == num_acquisitions
+    assert [
+        len(observer.events) for observer in every_n_samples_event_observers
+    ] == every_n_samples_event_counts
```
bkeryan, line 228: The first batch of register/unregister tests don't verify that the event works when you reregister it, so I added this test.

tests/component/test_task_events.py
```diff
@@ -0,0 +1,414 @@
...
+
+    with pytest.raises(nidaqmx.DaqError) as exc_info:
+        ao_task.register_every_n_samples_acquired_into_buffer_event(
+            100, event_observer.handle_every_n_samples_event
+        )
+
+    assert (
+        exc_info.value.error_code
+        == DAQmxErrors.EVERY_N_SAMPLES_ACQ_INTO_BUFFER_EVENT_NOT_SUPPORTED_BY_DEVICE
+    )
```
bkeryan, line 358: I added these tests because we could reasonably catch double registration in Python and return the correct error, but we can't catch all of the other registration errors that DAQmx might possibly return.

tests/component/test_task_events.py
```diff
@@ -0,0 +1,414 @@
...
+def test___task___register_unregister_signal_event___callback_not_invoked(
+    ai_task: nidaqmx.Task,
+) -> None:
+    event_observer = SignalEventObserver()
+
+    for _ in range(10):
+        ai_task.register_signal_event(Signal.SAMPLE_COMPLETE, event_observer.handle_signal_event)
+        ai_task.register_signal_event(Signal.SAMPLE_COMPLETE, None)
+
+    assert len(event_observer.events) == 0
```
bkeryan, line 414: I was pleasantly surprised to see that these tests fail with gRPC: grpc-device reports the registration error as a normal response, so we call the event callback. For every N and signal events, we ignore the response's `status` field, though...

tests/component/test_task_events.py
```diff
@@ -162,51 +169,96 @@ def test___done_and_every_n_samples_events_registered___run_finite_acquisition__
...
 
-    assert len(done_event_observer.events) == 3
-    assert len(every_n_samples_event_observer.events) == 30
+    assert len(done_event_observer.events) == num_acquisitions
+    assert len(every_n_samples_event_observer.events) == num_acquisitions * every_n_event_count
 
 
-@pytest.mark.library_only
+@pytest.mark.xfail(reason="AB#2395984: GrpcStubInterpreter does not unregister events with DAQmx")
+@pytest.mark.library_only  # Test hangs with gRPC
```
zhindes, line 198: It doesn't run on gRPC, so why is it expected to fail?

tests/component/test_task_events.py
```diff
@@ -240,21 +296,69 @@ def test___every_n_samples_transferred_from_buffer_event_registered___register_e
...
+    )
 
     with pytest.raises(nidaqmx.DaqError) as exc_info:
         ai_task.register_signal_event(Signal.SAMPLE_COMPLETE, event_observer.handle_signal_event)
 
     assert exc_info.value.error_code == DAQmxErrors.SIGNAL_EVENT_ALREADY_REGISTERED
 
 
-@pytest.mark.library_only
+@pytest.mark.xfail(
```
zhindes, line 317: Doesn't this test pass for lib? So would show up as "unexpectedly passing"?

zhindes (approved):
minor issues that are probably just my misunderstanding. I trust that this all works, or the modifications are minor.

tests/component/test_task_events.py
bkeryan: It's expected to fail with gRPC, but it fails by hanging, so I also added library_only. 

I don't think these marks are expressive or accurate enough. We should add `grpc_skip(reason)` and `grpc_xfail(reason)`.

tests/component/test_task_events.py
bkeryan: Yes, for library it will have result XPASS. A grpc_xfail(reason) mark would be appropriate for most of these.

tests/component/test_task_events.py
bkeryan: I'll address this in a separate PR.

PR #371, opened 2023-05-22
tests: Fix namespacing by adding __init__.py files

bkeryan:

Add `__init__.py` files for new test directories.

Running all the tests (not just tests/component) was returning errors like this one:
```
import file mismatch:
imported module 'test_task' has this __file__ attribute:
  D:\dev\nidaqmx-python\tests\component\test_task.py
which is not the same as the test file we want to collect:
  D:\dev\nidaqmx-python\tests\unit\test_task.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

This is because in https://github.com/ni/nidaqmx-python/pull/369 I added new test folders but I forgot the `__init__.py` files. I didn't catch this at the time because I was running a subset of the tests.

Ran all the tests with `poetry run pytest -v`. One legacy test (`test_mixed_chans_with_power`) fails with gRPC but I think that's because I'm testing with a grpc-device build that doesn't return `ni-samps-per-chan-read` yet.

### DIFF
--- a/tests/acceptance/__init__.py
+++ b/tests/acceptance/__init__.py
@@ -1,0 +1,1 @@
+"""Acceptance tests for the nidaqmx package."""
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,1 +1,1 @@
-"""NI-DAQmx tests."""
+"""Tests for the nidaqmx package."""
--- a/tests/component/__init__.py
+++ b/tests/component/__init__.py
@@ -1,0 +1,1 @@
+"""Component tests for the nidaqmx package."""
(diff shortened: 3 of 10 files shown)


zhindes (approved):

tests/acceptance/__init__.py
bkeryan, line 1:
```
Run poetry run ni-python-styleguide lint
./tests/component/__init__.py:1:1: D104 Missing docstring in public package
./tests/component/system/__init__.py:1:1: D104 Missing docstring in public package
./tests/component/system/storage/__init__.py:1:1: D104 Missing docstring in public package
./tests/acceptance/__init__.py:1:1: D104 Missing docstring in public package
Error: Process completed with exit code 255.
```

PR #377, opened 2023-05-24
codegen: Add support for gRPC event unregistration

bkeryan:

Add support for unregistering event when using gRPC.

Update xfail reason for corresponding tests.

Fixes AB#2395984: GrpcStubInterpreter does not unregister events with DAQmx

Tested with an NI gRPC Device Server build from my event unregistration dev branch:
- Ran `poetry run pytest -v` with no failures.
- Ran `poetry run pytest -v --runxfail .\tests\component\test_task_events.py` with no failures.


tests/component/test_task_events.py
```diff
@@ -91,6 +92,9 @@ def test___signal_event_registered___run_finite_acquisition___callback_invoked_n
     assert all(e.signal_type == Signal.SAMPLE_COMPLETE.value for e in event_observer.events)
 
 
+@pytest.mark.grpc_xfail(
+    reason="Requires NI gRPC Device Server version 2.2 or later", raises=RpcError
```
zhindes, line 96: It'd be kind of nice if we could ask the gRPC server what version it is, but that's a lot of work just for making stuff like this a little bit more resilient.

bkeryan: I don't think the NI gRPC Device Server supports querying its version number over RPC.

For tests, we could read the version number from the EXE:
```
>>> import win32api
>>> server_exe = "C:\\Program Files\\National Instruments\\Shared\\NI gRPC Device Server\\ni_grpc_device_server.exe"
>>> lang, codepage = win32api.GetFileVersionInfo(server_exe, '\\VarFileInfo\\Translation')[0]
>>> product_version = win32api.GetFileVersionInfo(server_exe, f"\\StringFileInfo\\{lang:04X}{codepage:04X}\\ProductVersion")
>>> tuple(int(x) for x in product_version.split('.'))
(2, 1)
```

I think it's beneficial to test that nidaqmx-python has reasonable behavior with the old server, so I don't think we should automatically skip these tests with the old server. It would be reasonable to do a conditional xfail, but it's probably not worth it.

PR #382, opened 2023-05-26
fix: update tests to also run with grpc options

keer1111:

Update the test cases using to use fixtures that passes grpc options as well to cover grpc calls and fixes in those calls. 

To fix the following
- [Bug 2400882](https://dev.azure.com/ni/DevCentral/_workitems/edit/2400882): tests/legacy/test_watchdog.py and tests/legacy/test_power_up_states.py do not test gRPC
- [Bug 2400884](https://dev.azure.com/ni/DevCentral/_workitems/edit/2400884): tests/legacy/test_multi_threading.py does not test gRPC

Test case updates are tested in remote PC.

### DIFF
--- a/tests/acceptance/test_examples.py
+++ b/tests/acceptance/test_examples.py
@@ -15,7 +15,9 @@ EXAMPLE_PATHS = [p for p in EXAMPLES_DIRECTORY.glob("**/*.py") if p.name != "__i
 
 
 @pytest.mark.parametrize("example_path", EXAMPLE_PATHS)
+@pytest.mark.library_only
 def test___shipping_example___run___no_errors(example_path: Path, system):
+    # This test runs the example programs, which do not support gRPC. Hence this test is marked library_only
     example_source = example_path.read_text()
     for device_name in _find_device_names(example_source):
         if device_name not in system.devices:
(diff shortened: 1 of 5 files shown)


tests/acceptance/test_examples.py
```diff
@@ -15,13 +15,12 @@
 
 
 @pytest.mark.parametrize("example_path", EXAMPLE_PATHS)
-def test___shipping_example___run___no_errors(example_path: Path):
-    local_system = nidaqmx.system.System.local()
+def test___shipping_example___run___no_errors(example_path: Path, system):
```
bkeryan, line 18: This test runs the example programs, which do not support gRPC, so mark it library_only and add a comment stating the reason.

bkeryan (approved):
Approved with suggestions

zhindes (approved):

tests/acceptance/test_examples.py
```diff
@@ -15,13 +15,15 @@
 
 
 @pytest.mark.parametrize("example_path", EXAMPLE_PATHS)
-def test___shipping_example___run___no_errors(example_path: Path):
-    local_system = nidaqmx.system.System.local()
+@pytest.mark.library_only
+def test___shipping_example___run___no_errors(example_path: Path, system):
+    # This test runs the example programs, which do not support gRPC.
+    # Hence this test is marked library_only.
```
bkeryan, line 21: I was expecting something more like:
```
@pytest.mark.library_only  # The example programs do not support gRPC.
```

But now that I think about it, it would make sense for the `grpc_only` and `library_only` marks to have a `reason=` keyword argument like the skip/xfail marks. https://github.com/ni/nidaqmx-python/pull/387

PR #385, opened 2023-05-26
codegen: Make BaseInterpreter.get_error_string() behavior consistent across library/grpc

bkeryan:

Make `BaseInterpreter.get_error_string()` behavior consistent across `LibraryInterpreter` and `GrpcStubInterpreter`.

LibraryInterpreter:
- Extract `get_error_string` and `get_extended_error_info` methods from the body of `check_for_error`.
- Add an error check to `get_error_string` and `get_extended_error_info`.
- Log error codes returned from `DAQmxGetErrorString` and `DAQmxGetExtendedErrorInfo`.
- Move `check_for_error` into the `LibraryInterpreter` class and update all callers.

GrpcStubInterpreter:
- Add a hand-written implementation of `get_error_string` that does not recursively call itself.
- Add an error check to `get_error_string`.
- Log exceptions returned from the `GetErrorString` RPC.
- Add a special case to distinguish between RPC errors and "error not found".

tests:
- Add a `grpc_channel_with_errors` fixture that returns errors for all RPCs.
- Add `interpreter` and `grpc_interpreter_with_errors` fixtures.
- Add component tests for `get_error_string`.

Make `get_error_string` behavior more consistent.

Allow more internal code to call `get_error_string`.

`poetry run pytest -v`, only `test_mixed_chans_with_power` failing.
`poetry run tox -e py38-base`


src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -157,6 +162,20 @@ class GrpcStubInterpreter(BaseInterpreter):
...
         return hash(task_handle.name)
 
+    ## get_error_string has special error handling.
+    def get_error_string(self, error_code):
+        try:
+            # Do not use self._invoke() because it may call back into self.get_error_string().
+            response = self._client.GetErrorString(
+                grpc_types.GetErrorStringRequest(error_code=error_code))
+            if not response.error_string:
+                return _UNABLE_TO_LOCATE_ERROR_RESOURCES_ERROR_MESSAGE
```
bkeryan, line 172: FYI, this is another IVI dance inconsistency: if the buffer pointer is not NULL, DAQmxGetErrorString(-12345) returns "Error code could not be found...", but if the buffer pointer is NULL, it returns buffer size of 0. If grpc-device specified a non-NULL buffer pointer, it would have different behavior.

image

PR #386, opened 2023-05-29
fix: update write teds array grpc call to pass bytes

keer1111:

Fix the write_to_teds_array grpc call to pass bytes as expected

To fix - [Bug 2400963](https://dev.azure.com/ni/DevCentral/_workitems/edit/2400963): write_to_teds_with_array() errors with GrpcStubInterpreter

The testing code snippet in the above bug item is run and grpc call throws meaningful error if TEDS is not configured.

### DIFF
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -3335,7 +3335,8 @@ class GrpcStubInterpreter(BaseInterpreter):
         response = self._invoke(
             self._client.WriteToTEDSFromArray,
             grpc_types.WriteToTEDSFromArrayRequest(
-                physical_channel=physical_channel, bit_stream=bit_stream,
+                physical_channel=physical_channel,
+                bit_stream=bit_stream.tobytes(),
                 basic_teds_options_raw=basic_teds_options))
 
     def write_to_teds_from_file(
--- a/tests/component/system/test_physical_channel.py
+++ b/tests/component/system/test_physical_channel.py
@@ -1,4 +1,9 @@
+import nidaqmx
+from nidaqmx.error_codes import DAQmxErrors
 from nidaqmx.system import PhysicalChannel
+import pytest
+
+from tests.component.system.test_physical_channel_properties import VALUES_IN_TED
 
 
 def test___physical_channels_with_same_name___compare___equal(init_kwargs):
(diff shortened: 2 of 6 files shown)


zhindes (approved):

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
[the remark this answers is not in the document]
keer1111: I tried using the 'VALUES_IN_TED' array (Please point out if this is not the right one) from the `test_physical_channel_properties` test file. But still getting `TEDS_SENSOR_NOT_DETECTED` since there is not configured TEDS sensor. (Is there a way to simulate this?)

In order to verify this test case, will it be possible to configure a TEDS sensor?

tests/component/system/test_physical_channel.py
```diff
@@ -37,3 +42,53 @@ def test___physical_channels_with_different_names___hash___not_equal(
 
     assert hash(phys_chan1) != hash(phys_chan2)
     assert hash(phys_chan1) != hash(phys_chan3)
+
+
+def test__physical_channel__write_to_teds_from_array_with_invalid_stream__throws_data_error(
```
bkeryan, line 47: Move "invalid stream" to the given clause.

Correct terminology: "bitstream" not "stream")

This also applies to the other test cases.

`test___invalid_bitstream___write_to_teds_from_array___throws_data_error`
`test___invalid_file_path___write_to_teds_from_file___throws_data_error`

tests/component/system/test_physical_channel.py
```diff
@@ -37,3 +42,53 @@ def test___physical_channels_with_different_names___hash___not_equal(
...
+    ]
+
+
+def test__physical_channel__write_to_teds_from_file_with_invalid_file_path__throws_data_error(
+    any_x_series_device,
+):
+    phys_chan = any_x_series_device.ai_physical_chans["ai0"]
+
+    with pytest.raises(nidaqmx.DaqError) as exception:
+        phys_chan.write_to_teds_from_file()
```
bkeryan, line 78: FYI: I wonder why the file path is optional

Cc: @zhindes

bkeryan (changes requested):

bkeryan (approved):

PR #389, opened 2023-05-31
Handle timeout behavior of read functions

Sakthi-SM:
Whenever a timeout exception occurs in read functions of NI-DAQmx, the error code will be sent through the response rather than through gRPC exception. This PR adds an error check in all the read functions. 

This implements [Bug 2399915](https://dev.azure.com/ni/DevCentral/_workitems/edit/2399915): Partial reads report samps-per-chan-read via DaqReadError but don't update read_array.

The changes have been tested manually.


generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3362,14 +3394,15 @@ def get_error_string(self, error_code):
...
-def _assign_numpy_array(numpy_array, grpc_array):
+def _assign_numpy_array(numpy_array, grpc_array, samples_read):
     """
     Assigns grpc array to numpy array maintaining the original shape.
 
     Checks for the instance of grpc_array with bytes, if validated to True,
     the numpy array is assigned to a 1D array of the grpc arrray.
     """
     grpc_array_size = len(grpc_array)
+    grpc_array[samples_read:] = numpy_array[samples_read:]
```
bkeryan, line 3405: It looks like you're trying to preserve the unread part of the numpy array by copying it to the grpc array and then copying it back to the numpy array. This approach might happen to fix the one failing test, but it's wrong.
- It modifies the wrong array. A function named `_assign_numpy_array` should assign to the numpy array. It should modify the source array.
- I doubt that it works with `bytes` or 2D arrays.

The non-destructive way to do this is: `grpc_array_size = min(len(grpc_array), samples_read)`

However, this is still wrong. The array contains `(num_channels * samps_per_chan)` elements, not `samps_per_chan` elements.

Can you get the info you need from the numpy array's shape? I don't know. Looking at https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/stream_readers.py#L72 there are 3 supported array shapes:

```
        if is_many_chan:
            if is_many_samp:
                array_shape = (number_of_channels,
                               number_of_samples_per_channel)
            else:
                array_shape = (number_of_channels,)
        else:
            if is_many_samp:
                array_shape = (number_of_samples_per_channel,)
```

Both `(number_of_channels,)` and `(number_of_samples_per_channel,)` are 1D arrays, so I don't think you can distinguish between them based on the shape of the array.

generated/nidaqmx/_grpc_interpreter.py
bkeryan, line 1: The failing test only covers single-channel tasks. We need a multi-channel version of this test case.

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
bkeryan: I think all of the affected read functions have a `num_samps_per_chan` parameter. It should be possible to calculate `num_channels` based on the array size and `num_samps_per_chan`, but be careful about the corner cases (`num_samps_per_chan == 0`, `read_array.size % num_samps_per_chan != 0`). Also make sure to use integer division (`//`) not floating point division (`/`).

generated/nidaqmx/_grpc_interpreter.py
bkeryan: Some unit tests for `_assign_numpy_array` could also be very helpful. Import `nidaqmx._stubs.nidaqmx_pb2` and construct the gRPC response objects directly in the test cases (in order to use the real gRPC repeated type), then call `_assign_numpy_array` and validate the results.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -84,6 +84,14 @@ def _handle_rpc_error(self, rpc_error):
         else:
             self._raise_error(error_code, error_message, samps_per_chan_written, samps_per_chan_read)
 
+    def _check_for_error_from_response(self, error_code, samps_per_chan_read):
+        if error_code != 0:
```
bkeryan, line 88: How much overhead does calling GetErrorString() add? (Check NI IO Trace.)

Timeout errors are normal program flow, not exceptional cases, so it may make sense to have a special case that returns a hardcoded copy of the timeout error string. (The DAQmx C API does not currently support localization.)

generated/nidaqmx/_grpc_interpreter.py
bkeryan, line 1: For non-timeout read errors, I'm starting to wonder if we should discard samps_per_chan_read and not report it as DaqReadError. 
- If the user passed in an array, a DaqReadError will tell them that the array contains samples that it does not contain.
- If the user didn't pass in an array, DaqReadError is benign, but I don't think the samps_per_chan_read is very useful in that case.

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -84,6 +84,14 @@ def _handle_rpc_error(self, rpc_error):
         else:
             self._raise_error(error_code, error_message, samps_per_chan_written, samps_per_chan_read)
 
+    def _check_for_error_from_response(self, error_code, samps_per_chan_read):
```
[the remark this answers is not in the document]
Sakthi-SM: As per the suggestion, added the optional parameters into the method definition.

src/codegen/templates/grpc_interpreter/default_grpc_function_call.py.mako
```diff
@@ -59,7 +59,7 @@
 %if is_read_method:
     %for param in get_read_array_parameters(function):
         _assign_numpy_array(${param}, response.${param}, response.samps_per_chan_read)
-        self._check_for_error_from_response(response.status, response.samps_per_chan_read)
+        self._check_for_error_from_response(response.status, samps_per_chan_read=response.samps_per_chan_read)
```
bkeryan, line 62: Move this outside the loop to avoid calling it multiple times

generated/nidaqmx/_grpc_interpreter.py
bkeryan: The current version of this code still modifies the gRPC array

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
Sakthi-SM: I have added unit test to cover the various cases for `_assign_numpy_array`

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3392,10 +3438,26 @@ def _assign_numpy_array(numpy_array, grpc_array):
...
-        numpy_array.flat[:grpc_array_size] = numpy.frombuffer(grpc_array, dtype=numpy_array.dtype)
+        if numpy_array.ndim>1:
+            (number_of_channels, number_of_samples_per_channel) = numpy_array.shape
+            read_array = numpy.frombuffer(grpc_array, dtype=numpy_array.dtype).reshape((number_of_channels, samples_read))
+            for row in range(0, number_of_channels):
+                for column in range(0, number_of_samples_per_channel):
+                    if column < samples_read:
+                        numpy_array[row, column] = read_array[row, column]
+        else:
+            numpy_array.flat[:grpc_array_size] = numpy.frombuffer(grpc_array, dtype=numpy_array.dtype)
```
zhindes, line 3449: why does this case not care about `samples_read`?

zhindes (changes requested):

generated/nidaqmx/_grpc_interpreter.py
Sakthi-SM: I have updated the logic to include `sample_read` while converting bytes into numpy array

zhindes (approved):

generated/nidaqmx/_grpc_interpreter.py
bkeryan: FYI, here are my timing numbers from when we discussed this over Teams chat:

```
>>> import nidaqmx._library_interpreter
>>> interpreter = nidaqmx._library_interpreter.LibraryInterpreter()
>>> interpreter.get_error_string(-200284)
'Some or all of the samples requested have not yet been acquired.\nTo wait for the samples to become available use a longer read timeout or read later in your program. To make the samples available sooner, increase the sample rate. If your task uses a start trigger,  make sure that your start trigger is configured correctly. It is also possible that you configured the task for external timing, and no clock was supplied. If this is the case, supply an external clock.'
>>> import timeit
>>> timeit.timeit(lambda: interpreter.get_error_string(-200284), number=1000) / 1000
4.499499999980117e-06
```
```
>>> import nidaqmx._grpc_interpreter
>>> import grpc
>>> grpc_channel = grpc.insecure_channel("localhost:31763")
>>> grpc_options = nidaqmx.GrpcSessionOptions(grpc_channel, "")
>>> grpc_interpreter = nidaqmx._grpc_interpreter.GrpcStubInterpreter(grpc_options)
>>> timeit.timeit(lambda: grpc_interpreter.get_error_string(-200284), number=1000) / 1000
0.0007822051999999644
```

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3392,10 +3438,26 @@ def _assign_numpy_array(numpy_array, grpc_array):
...
     else:
         assert numpy_array.size >= grpc_array_size
-        numpy_array.flat[:grpc_array_size] = grpc_array
+        if numpy_array.ndim > 1:
+            (number_of_channels, number_of_samples_per_channel) = numpy_array.shape
+            read_array = numpy.array(grpc_array).reshape((number_of_channels, samples_read))
+            for row in range(0, number_of_channels):
+                for column in range(0, number_of_samples_per_channel):
+                    if column < samples_read:
+                        numpy_array[row, column] = read_array[row, column]
```
bkeryan, line 3458: Performance problems:
1. Looping over numpy arrays in Python is slow. Let numpy do the looping in C.
2. This creates a temporary numpy array, which I think we can avoid.

For 1, use slicing to create a view of a subset of the array (https://numpy.org/doc/stable/user/basics.indexing.html)
```
numpy_array[:, :samps_per_chan_read] = read_array
```

For 2, combine slicing with a flat iterator:
```
numpy_array[:, :samps_per_chan_read].flat[:grpc_array_size] = grpc_array
```

Note that `flat[:grpc_array_size]` handles the case where grpc_array is too short, preventing the flat iterator assignment from repeating the grpc_array elements. If samps_per_chan_read is consistent with grpc_array, then maybe this is not needed, but I'm not sure.

tests/unit/test_grpc_interpreter.py
```diff
@@ -0,0 +1,145 @@
+import numpy
+
+from nidaqmx._grpc_interpreter import _assign_numpy_array
+from nidaqmx._stubs import nidaqmx_pb2
+
+
+def test__grpc_array_with_all_samples__update_inuput_array__all_samples_updated():
```
bkeryan, line 7: 3 underscores

bkeryan, line 7: spelling: inuput -> input (whole file)

tests/unit/test_grpc_interpreter.py
```diff
@@ -0,0 +1,145 @@
...
+    number_of_samples_read = 100
+    initial_value = -1
+    read_value = 100
+    input_array = numpy.full(number_of_samples_in_input, initial_value, dtype=numpy.float64)
+    output_array = numpy.full(number_of_samples_in_input, read_value, dtype=numpy.float64)
+    response = nidaqmx_pb2.ReadAnalogF64Response(read_array=output_array)
+
+    _assign_numpy_array(input_array, response.read_array, number_of_samples_read)
+
+    assert all(element == read_value for element in input_array)
```
bkeryan, line 18: I think this is more numpy-esque:
```
assert (read_value == input_array).all()
assert (read_value == input_array[:number_of_samples_read]).all()
```
etc.

== returns an ndarray of bool and all() checks that all of the bools in the array are true. I see you are also using this technique with count_nonzero, which is good.

I would hope that this allows pytest -vvv to print the ndarray of bools when the test fails.

tests/unit/test_grpc_interpreter.py
bkeryan, line 7: grpc is in the filename so you don't have to put it in each test name.

tests/unit/test_grpc_interpreter.py
bkeryan, line 1: Please add 1d array tests for N channels

bkeryan (changes requested):

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -143,6 +143,14 @@ class GrpcStubInterpreter(BaseInterpreter):
         else:
             self._raise_error(error_code, error_message, samps_per_chan_written, samps_per_chan_read)
 
+    def _check_for_error_from_response(self, error_code, samps_per_chan_written=None, samps_per_chan_read=None):
+        if error_code != 0:
+            if error_code in ERROR_MESSAGES:
+                error_message = ERROR_MESSAGES.get(error_code)
+            else:
+                error_message = self.get_error_string(error_code)
```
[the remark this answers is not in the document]
bkeryan: Also add a comment that this is an optimization for partial reads, so that someone doesn't refactor it into `error_message = ERROR_MESSAGES.get(error_code, self.get_error_string(error_code))`

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3392,10 +3438,26 @@ def _assign_numpy_array(numpy_array, grpc_array):
...
+        else:
+            numpy_array[0:samples_read] = grpc_array[0:samples_read]
```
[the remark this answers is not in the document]
Sakthi-SM: Yes Brad, as you pointed out the above case will not work. But I don't think this a common use case/ behavior. Can we take this as part of a separate task?

generated/nidaqmx/_grpc_interpreter.py
bkeryan: "We need a multi-channel version of this test case." is not resolved.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -143,6 +144,14 @@ class GrpcStubInterpreter(BaseInterpreter):
         else:
             self._raise_error(error_code, error_message, samps_per_chan_written, samps_per_chan_read)
 
+    def _check_for_error_from_response(self, error_code, samps_per_chan_written=None, samps_per_chan_read=None):
+        if error_code != 0:
+            # This is an optimization for the partial read operation.
+            error_message=_ERROR_MESSAGES.get(error_code, None)
+            if not error_message:
+                error_message=self.get_error_string(error_code)
+            self._raise_error(error_code, error_message, samps_per_chan_written = samps_per_chan_written, samps_per_chan_read = samps_per_chan_read)
```
bkeryan, line 153: > [bkeryan](https://github.com/bkeryan) [2 weeks ago](https://github.com/ni/nidaqmx-python/pull/389#discussion_r1212397145)
>[style] Don't leave a space around the = in keyword arguments.

```suggestion
            error_message = _ERROR_MESSAGES.get(error_code, None)
            if not error_message:
                error_message = self.get_error_string(error_code)
            self._raise_error(error_code, error_message, samps_per_chan_written=samps_per_chan_written, samps_per_chan_read=samps_per_chan_read)
```

Sakthi-SM: Updated as per the suggestion

generated/nidaqmx/_grpc_interpreter.py
Sakthi-SM: I have added a component test to validate the multichannel case

tests/acceptance/test_multi_threading.py
```diff
@@ -200,6 +200,7 @@ def test___shared_interpreter___run_multiple_acquisitions_with_events___callback
         assert all(status == 0 for status in done_statuses)
```
Sakthi-SM, line 202: As per the fix made in https://github.com/ni/grpc-device/pull/940, this test requires the latest gRPC device server.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -221,10 +230,24 @@ def _assign_numpy_array(numpy_array, grpc_array):
...
+            numpy_array.flat[0:samps_per_chan_read] = numpy.frombuffer(grpc_array, dtype=numpy_array.dtype)
     else:
         assert numpy_array.size >= grpc_array_size
-        numpy_array.flat[:grpc_array_size] = grpc_array
+        if numpy_array.ndim > 1:
+            (number_of_channels, number_of_samples_per_channel) = numpy_array.shape
+            read_array = numpy.array(grpc_array).reshape((number_of_channels, samps_per_chan_read))
+            numpy_array[:, :samps_per_chan_read].flat[:grpc_array_size] = grpc_array
+        elif num_samps_per_chan==1 and numpy_array.size<=samps_per_chan_read:
+            numpy_array.flat[0:numpy_array.size] = grpc_array[0:numpy_array.size]
```
bkeryan, line 248: What is this middle case intended to do?

It looks like it's an attempt to handle N-channel 1-sample, but it won't work because samps_per_chan_read will be 1 in that case, not N.

It won't handle flat arrays (raw reads) either because those typically have num_samps_per_chan > 1.

generated/nidaqmx/_grpc_interpreter.py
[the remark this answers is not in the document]
bkeryan: I think there are two cases you aren't handling correctly:
- Flat arrays
- N channels 1 sample

This PR introduces a data truncation bug: it copies fewer samples than it should in some cases. I think that's worse than the timeout bug you are trying to fix, which overwrites unread samples.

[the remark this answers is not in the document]
bkeryan: FYI, I just noticed that read_raw and some other read functions return num_bytes_per_samp, which can be used to solve this without passing additional parameters.

tests/unit/test_grpc_interpreter.py
```diff
@@ -0,0 +1,193 @@
...
+from nidaqmx._stubs import nidaqmx_pb2
+
+
+def test___1_channel_n_sample_per_chan___update_input_array___all_samples_updated():
+    number_of_samples_per_chan = 100
+    number_of_samples_read = 100
+    initial_value = -1
+    read_value = 100
+    input_array = numpy.full(number_of_samples_per_chan, initial_value, dtype=numpy.float64)
+    output_array = numpy.full(number_of_samples_per_chan, read_value, dtype=numpy.float64)
```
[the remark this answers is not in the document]
Sakthi-SM: The tests related to `asssign_numpy_array` are moved and these comments will be addressed in the PR that fixes  [Bug 2420742](https://dev.azure.com/ni/DevCentral/_workitems/edit/2420742): gRPC interpreter doesn't retain the initial array values when converting flattened array into numpy array.

tests/unit/test_grpc_interpreter.py
```diff
@@ -0,0 +1,193 @@
...
+    response = nidaqmx_pb2.ReadAnalogF64Response(read_array=output_array)
+
+    _assign_numpy_array(
+        input_array, response.read_array, number_of_samples_read, number_of_samples_per_chan
+    )
+
+    assert (read_value == input_array[:number_of_samples_read]).all()
+    assert (
+        initial_value == input_array[number_of_samples_per_chan - number_of_samples_read :]
+    ).all()
```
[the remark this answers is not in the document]
Sakthi-SM: This will be addressed in  [Bug 2420742](https://dev.azure.com/ni/DevCentral/_workitems/edit/2420742): gRPC interpreter doesn't retain the initial array values when converting flattened array into numpy array.

tests/unit/test_grpc_interpreter.py
```diff
@@ -0,0 +1,193 @@
...
+    number_of_channels = 10
+    number_of_samples_read = 100
+    initial_value = -1
+    read_value = 100
+    input_array = numpy.full(
+        (number_of_channels, number_of_samples_per_chan), initial_value, dtype=numpy.float64
+    )
+    output_array = numpy.full(
+        (number_of_channels * number_of_samples_read), read_value, dtype=numpy.float64
+    ).tobytes()
```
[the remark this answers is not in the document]
Sakthi-SM: This will be addressed in  [Bug 2420742](https://dev.azure.com/ni/DevCentral/_workitems/edit/2420742): gRPC interpreter doesn't retain the initial array values when converting flattened array into numpy array.

tests/legacy/test_read_exceptions.py
```diff
@@ -203,3 +206,80 @@ def test_timeout_stream(self, generate_task, real_x_series_device):
...
+        # Do a partial read which should succeed.
+        data = numpy.full(
+            (number_of_channels, samples_to_read), sentinel_value, dtype=numpy.float64
+        )
+        num_values_read = reader.read_many_sample(
+            data, number_of_samples_per_channel=samples_to_read, timeout=2.0
+        )
+        assert num_values_read == samples_to_read
+        # All the data should have been overwritten.
+        assert (data != sentinel_value).any()
```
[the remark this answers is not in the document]
Sakthi-SM: This will be addressed in  [Bug 2420742](https://dev.azure.com/ni/DevCentral/_workitems/edit/2420742): gRPC interpreter doesn't retain the initial array values when converting flattened array into numpy array.

tests/legacy/test_read_exceptions.py
```diff
@@ -203,3 +206,80 @@ def test_timeout_stream(self, generate_task, real_x_series_device):
...
+            num_values_read = reader.read_many_sample(
+                data, number_of_samples_per_channel=samples_to_read, timeout=2.0
+            )
+
+        assert timeout_exception.value.error_code == DAQmxErrors.SAMPLES_NOT_YET_AVAILABLE
+        # We should have read a partial dataset.
+        number_of_samples_expected = clocks_to_give - samples_to_read
+        assert timeout_exception.value.samps_per_chan_read == number_of_samples_expected
+        # The data that wasn't read should have been unmodified.
+        assert sum(numpy.count_nonzero(element == sentinel_value) for element in data) == 75
```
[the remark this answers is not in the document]
Sakthi-SM: The number of samples per channel is 75 and based on  the timeout that we have given it should read 50 samples and 25 samples should be unmodified in each channel. But what I actually observed is in one channel 50 samples was read and in another channel only 25 samples was read. I wasn't able to figure out the exact reason for this discrepancy. Do you have any suggestions on the above behavior @bkeryan?

src/codegen/templates/_grpc_interpreter.py.mako
Sakthi-SM: It was supposed to handle N-channel 1-sample case but as you pointed out it needs rework. This will be addressed in [Bug 2420742](https://dev.azure.com/ni/DevCentral/_workitems/edit/2420742): gRPC interpreter doesn't retain the initial array values when converting flattened array into numpy array.

tests/legacy/test_read_exceptions.py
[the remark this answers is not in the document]
bkeryan: It's using the wrong array as the source of the assignment.
```
        if numpy_array.ndim > 1:
            (number_of_channels, number_of_samples_per_channel) = numpy_array.shape	
            read_array = numpy.array(grpc_array).reshape((number_of_channels, samps_per_chan_read))	
            numpy_array[:, :samps_per_chan_read].flat[:grpc_array_size] = grpc_array
```

Once you fix that, it will fail for a different reason.

The server should return something like this (but with 10x the data)
```
A0 A1 A2 X X X X B0 B1 B2 X X X X
```
```
>>> import numpy
>>> a = numpy.array([1,2,3,0,0,0,0,4,5,6,0,0,0,0])
>>> a
array([1, 2, 3, 0, 0, 0, 0, 4, 5, 6, 0, 0, 0, 0])
```

samps_per_chan_read should be 3, but we can't reshape a 1x14 array into 2x3:
```
>>> b = a.reshape((2,3))
ValueError: cannot reshape array of size 14 into shape (2,3)
```

We can reshape it into 2x7:
```
>>> b = a.reshape((2,7))
>>> b
array([[1, 2, 3, 0, 0, 0, 0],
       [4, 5, 6, 0, 0, 0, 0]])
```

We can't assign a 2x3 slice based on an 2x7 array:
```
>>> c = numpy.full((2,7),-1)
>>> c
array([[-1, -1, -1, -1, -1, -1, -1],
       [-1, -1, -1, -1, -1, -1, -1]])
>>> c[:,:3] = b
ValueError: could not broadcast input array from shape (2,7) into shape (2,3)
```

But we can assign a 2x3 slice based on the 2x3 slice of a 2x7 array:
```
>>> c[:,:3] = b[:,:3]
>>> c
array([[ 1,  2,  3, -1, -1, -1, -1],
       [ 4,  5,  6, -1, -1, -1, -1]])
```

I think you need something like 
```
        if numpy_array.ndim > 1:
            (number_of_channels, number_of_samples_per_channel) = numpy_array.shape	
            read_array = numpy.array(grpc_array).reshape((number_of_channels, grpc_array_size // number_of_channels))	
            numpy_array[:, :samps_per_chan_read] = read_array[:, :samps_per_chan_read]
```

I also think that reducing the number of samples and comparing against an expected array like `numpy.array([[1, 2, 3, -1, -1, -1, -1], [4, 5, 6, -1, -1, -1, -1]])` will make the assertions simpler and more accurate.

bkeryan (approved):

PR #390, opened 2023-05-31
Fix - powerup states with output type

keer1111:

Fix get_analog_power_up_states_with_output_type call

- Add ctype cast before C call
- Replace`AOPowerUpOutputBehavior` with`PowerUpChannelType` to cast to the ChannelType enum to frame the PowerUpStates

To Fix: [Bug 2400896](https://dev.azure.com/ni/DevCentral/_workitems/edit/2400896): get_analog_power_up_states_with_output_type() returns errors with both LibraryInterpreter and GrpcStubInterpreter

Added the testing code shared as part of the above bug as separate test case

### DIFF
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -2274,6 +2274,7 @@ class LibraryInterpreter(BaseInterpreter):
             self, channel_names, array_size):
         state_array = numpy.zeros(array_size, dtype=numpy.float64)
         channel_type_array = numpy.zeros(array_size, dtype=numpy.int32)
+        array_size = ctypes.c_uint32(array_size)
 
         cfunc = lib_importer.cdll.DAQmxGetAnalogPowerUpStatesWithOutputType
         if cfunc.argtypes is None:
--- a/tests/component/system/test_system.py
+++ b/tests/component/system/test_system.py
@@ -1,0 +1,12 @@
+from nidaqmx.constants import PowerUpChannelType
+
+
+def test___get_analog_power_up_states___returns_power_up_states(system):
+    channel_names = ["aoTester/ao0", "aoTester/ao1"]
+
+    power_up_states = system.get_analog_power_up_states_with_output_type(channel_names)
+
+    for i in range(len(channel_names)):
+        assert channel_names[i] ==  power_up_states[i].physical_channel
+        assert 0.0 == power_up_states[i].power_up_state
+        assert PowerUpChannelType.CHANNEL_HIGH_IMPEDANCE == power_up_states[i].channel_type
--- a/tests/component/system/test_system.py
+++ b/tests/component/system/test_system.py
@@ -1,10 +1,12 @@
 from nidaqmx.constants import PowerUpChannelType
 
 
-def test___get_analog_power_up_states___returns_power_up_states(system):
+def test___get_analog_power_up_states_with_output_type___returns_power_up_states(system):
     channel_names = ["aoTester/ao0", "aoTester/ao1"]
 
     power_up_states = system.get_analog_power_up_states_with_output_type(channel_names)
+
+    assert len(power_up_states) == len(channel_names)
 
     for i in range(len(channel_names)):
         assert channel_names[i] == power_up_states[i].physical_channel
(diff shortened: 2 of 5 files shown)


zhindes (approved):

tests/component/system/test_system.py
bkeryan, line 1: Sorry, I didn't file a separate bug about the set function. Please test and fix it too.

```
def test___set_analog_power_up_states_with_output_type___no_errors(system):
    power_up_states = [
        AOPowerUpState("aoTester/ao0", 1.0, PowerUpChannelType.CHANNEL_VOLTAGE),
        AOPowerUpState("aoTester/ao1", 2.0, PowerUpChannelType.CHANNEL_VOLTAGE),
    ]

    power_up_states = system.set_analog_power_up_states_with_output_type(power_up_states)
```

```
FAILED tests/component/system/test_system.py::test____set_analog_power_up_states_with_output_type____no_errors[library_init_kwargs] - TypeError: set_analog_power_up_states_with_output_type() takes 4 positional arguments but 5 were given
FAILED tests/component/system/test_system.py::test____set_analog_power_up_states_with_output_type____no_errors[grpc_init_kwargs] - TypeError: set_analog_power_up_states_with_output_type() takes 4 positional arguments but 5 were given
```

tests/component/system/test_system.py
```diff
@@ -0,0 +1,12 @@
+from nidaqmx.constants import PowerUpChannelType
+
+
+def test___get_analog_power_up_states___returns_power_up_states(system):
```
bkeryan, line 4: Add `_with_output_type` to distinguish this from tests for `get_analog_power_up_states` (if we write any).

bkeryan (changes requested):

tests/component/system/test_system.py
keer1111: Created a separate bug for set - [Bug 2409371](https://dev.azure.com/ni/DevCentral/_workitems/edit/2409371): set_analog_power_up_states_with_output_type() returns errors with both LibraryInterpreter and GrpcStubInterpreter
Will cover this in a separate PR . Hope this is okay.

generated/nidaqmx/_library_interpreter.py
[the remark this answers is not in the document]
keer1111: If we assert them to be equal, then we are expecting the array_size to be same from before and after? So shrinking is not allowed from daqmx-python?

Is it okay to assert in dev code?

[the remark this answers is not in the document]
bkeryan: It's okay to use assertions for conditions that should truly never happen.

When this function is used correctly, `flatten_channel_string(physical_channels)` should return a channel string with the same number of channels as `len(physical_channels)`, so `DAQmxGetAnalogPowerUpStatesWithOutputType` should not have a reason to shrink the array.

If this function is used incorrectly, like `get_analog_power_up_states_with_output_type(["Dev1/ao0:3"])`, then the channel string will have more channels than the array, so `DAQmxGetAnalogPowerUpStatesWithOutputType` should return an error. It shouldn't hit the assertion in this case.

tests/component/system/test_system.py
```diff
@@ -0,0 +1,14 @@
...
+    channel_names = ["aoTester/ao0", "aoTester/ao1"]
+
+    power_up_states = system.get_analog_power_up_states_with_output_type(channel_names)
+
+    assert len(power_up_states) == len(channel_names)
+
+    for i in range(len(channel_names)):
+        assert channel_names[i] == power_up_states[i].physical_channel
+        assert 0.0 == power_up_states[i].power_up_state
+        assert PowerUpChannelType.CHANNEL_HIGH_IMPEDANCE == power_up_states[i].channel_type
```
bkeryan, line 14: The test body should have 3 sections: arrange, act, assert

```suggestion
    assert len(power_up_states) == len(channel_names)
    for i in range(len(channel_names)):
        assert channel_names[i] == power_up_states[i].physical_channel
        assert 0.0 == power_up_states[i].power_up_state
        assert PowerUpChannelType.CHANNEL_HIGH_IMPEDANCE == power_up_states[i].channel_type
```

bkeryan (approved):

tests/component/system/test_system.py
bkeryan: Thanks

PR #393, opened 2023-06-02
handwritten: Fix interpreter sharing and add missing `name` properties

bkeryan:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).

### What does this Pull Request accomplish?

Fix interpreter sharing in `PhysicalChannelCollection`.

Add missing `name` property to `PersistedChannel`, `PersistedScale`, and `PersistedTask`.

Add tests for system storage collection classes.

### Why should this Pull Request be merged?

Tests interpreter sharing in system storage collection classes.

Fixes some bugs that I ran into while writing the tests.

### What testing has been done?

`poetry run pytest -v`
There was one failure related to the bitfield enum property change.

PR #396, opened 2023-06-02
fix: physical-channel-collection: use grpc interpreter instance in all & __getitem

keer1111:

Replaces PhysicalChannel init with AlternatePhysicalChannel constructor in appropriate places

To fix: [Bug 2400971](https://dev.azure.com/ni/DevCentral/_workitems/edit/2400971): PhysicalChannelCollection.all and PhysicalChannelCollection.__getitem__(str) always use LibraryInterpreter

Automated tests passing


bkeryan:
> I think @bkeryan might have a PR up doing this? Or maybe it was another instance of the same issue. Just FYI.
> 
> Also, are there tests that were failing that are now passing, or should we add a test for this?

Oops, I forgot I had filed a bug about this and fixed it in #393 along with adding tests.

PR #398, opened 2023-06-02
codegen: Move per-task event handling state to Task class and revise gRPC event error handling

bkeryan:

Move the per-task event handling state from the interpreter class to the `Task` class. Previously, tasks with a shared interpreter would interfere with each other when registering and unregistering events, which led to errors or crashes.

Revise gRPC event error handling:
- Distinguish between gRPC event stream exceptions and exceptions raised by the user-provided event callback function.
- gRPC event stream exceptions: Log the exception, save it, and reraise it when unregistering the event or closing the task.
- User-provided event callback exceptions: Log the exception and the `repr` of the callback function, then ignore it. 
  - This is closer to the behavior of `LibraryInterpreter`, which writes a message and the `repr` of the callback function to stderr and continues to deliver events.
  - It's not identical to `LibraryInterpreter`: `ctypes` reports callback errors using `sys.unraisablehook`, which application authors can replace with a custom hook. However, `sys.unraisablehook` seems to be designed to be called from C, not Python, so I am hesitant to try to do the same thing in `GrpcStubInterpreter`.

Remove null logging handler from `nidaqmx/__init__.py`. If the application doesn't configure logging, this hides all log messages generated by the `nidaqmx` moduel. The default behavior of logging warnings and errors to stderr is better because it displays errors logged from `GrpcStubInterpreter` even when the application doesn't configure logging.

Fixes AB#2390580: Unregistering events crashes when sharing interpreters

Ran `poetry run pytest -v` and `poetry run tox` at various points. Will run again.
Manually tested error reporting by adding exceptions to the `every_n_samples_event.py` example and to the event `_thread_main` function.


zhindes (approved):
nice!

src/handwritten/task.py
```diff
@@ -349,9 +365,36 @@ def close(self):
             return
 
         self._interpreter.clear_task(self._handle)
```
bkeryan, line 367: If you close the gRPC channel or shut down the server, `clear_task` raises an error and this code doesn't clean up the event handlers, so `__del__` reports a warning. I'll fix it and add tests in a follow-up PR.

src/handwritten/task.py
```diff
@@ -76,6 +77,11 @@ def __init__(self, new_task_name='', *, grpc_options=None):
         self._saved_name = new_task_name  # _initialize sets this to the name assigned by DAQmx.
         self._grpc_options = grpc_options
 
+        self._done_event_handler = None
+        self._every_n_acquired_event_handler = None
+        self._every_n_transferred_event_handler = None
+        self._signal_event_handler = None
```
bkeryan, line 83: Using 4 separate member variables makes this more complicated than it has to be. We should use a dict. I'll address this in a follow-up PR.

src/handwritten/task.py
bkeryan: #400

src/handwritten/task.py
bkeryan: #401

PR #402, opened 2023-06-05
tests: Add tests for event error handling

bkeryan:

Add component tests for event error handling:
- Given a closed gRPC channel, when you close the task, an exception is raised from clear_task but everything is still cleaned up.
- Given event callbacks that raise exceptions, when you run a finite acquisition, the exceptions are ignored (and in the case of gRPC, they are logged).

Add unit tests for event error handling:
- Given that clear_task or closing the event handler raises an exception, when you close the task, the exception is raised but everything is still cleaned up.

Add a `temporary_grpc_channel` marker:
- Specifying this marker causes the `grpc_channel` fixture to return a temporary gRPC channel instead of the per-session shared gRPC channel.
- This allows tests to close the gRPC channel or specify custom gRPC channel options.
- Implementing it this way allows reusing fixtures that depend on the `grpc_channel` fixture.
- Retrofit the interpreter tests to use this marker.

Add test coverage for event error handling.

`poetry run pytest -v`

### DIFF
--- a/tests/unit/test_task_events.py
+++ b/tests/unit/test_task_events.py
@@ -1,0 +1,89 @@
+from typing import Dict
+from unittest.mock import Mock
+
+import pytest
+from pytest_mock import MockerFixture
+
+from nidaqmx import Task
+from nidaqmx.task import _TaskEventType
+from tests.unit._grpc_utils import create_grpc_options
+from tests.unit._task_utils import (
+    expect_create_task,
+    expect_get_task_name,
+    register_event_handler,
+    register_event_handlers,
+)
+
+
+@pytest.mark.parametrize("event_type", _TaskEventType)
+def test___grpc_event_registered___leak_task___raises_resource_warning(
+    event_type: _TaskEventType, interpreter: Mock, mocker: MockerFixture
+):
+    expect_create_task(interpreter, "GrpcTaskHandle")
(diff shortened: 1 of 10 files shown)


tests/unit/test_task_events.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def test___events_registered_and_clear_task_error_raised___close___task_resources_cleaned_up_and_clear_task_error_raised(
+    interpreter: Mock, mocker: MockerFixture
+):
+    expect_create_task(interpreter, "MyTaskHandle")
+    expect_get_task_name(interpreter, "MyTask")
+    task = Task("MyTask")
+    event_handlers = register_event_handlers(mocker, task, _TaskEventType)
+    interpreter.clear_task.side_effect = RuntimeError("clear_task")
```
zhindes, line 41: Ahh, I was wondering what the side effect was going to be for. I like it!

zhindes (approved):

tests/unit/test_task_events.py
bkeryan: Are you referring to the side effect field in `BaseEventObserver`? This side effect field is in the `Mock` class: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock

Now that I think about it, passing an exception for the side effect is really what I wanted in `BaseEventObserver`.

PR #404, opened 2023-06-06
Fix -  set power up states with output type

keer1111:

Fixes the `get_analog_power_up_states_with_output_type` call

To fix: [Bug 2409371](https://dev.azure.com/ni/DevCentral/_workitems/edit/2409371): set_analog_power_up_states_with_output_type() returns errors with both LibraryInterpreter and GrpcStubInterpreter

New test cases added

### DIFF
--- a/tests/component/system/test_system.py
+++ b/tests/component/system/test_system.py
@@ -31,22 +31,23 @@ def test_valid_power_up_states___set_analog_power_up_states_with_output_type___s
         for channel_name in channel_names
     ]
 
-    result = system.set_analog_power_up_states_with_output_type(power_up_states)
-
-    assert result == None
+    system.set_analog_power_up_states_with_output_type(power_up_states)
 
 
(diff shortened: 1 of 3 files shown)


bkeryan (approved):
Looks good.

FYI, I updated the PR title to say "with" instead of "without"

tests/component/system/test_system.py
```diff
@@ -11,3 +16,40 @@ def test___get_analog_power_up_states_with_output_type___returns_power_up_states
...
+):
+    channel_names = ["aoTester/ao0", "aoTester/ao1"]
+    power_up_states = [
+        AOPowerUpState(
+            physical_channel=channel_name,
+            power_up_state=20,
+            channel_type=PowerUpChannelType.CHANNEL_VOLTAGE,
+        )
+        for channel_name in channel_names
+    ]
```
zhindes, line 50: Can you modify or add a new test case where only the second power up state is bad? That way we know that its actually iterating through all of them.

zhindes (approved):
one minor request

PR #408, opened 2023-06-06
Tests - set powerup states without output type

keer1111:

Add test case for set powerup states API

To complete - [Bug 2409381](https://dev.azure.com/ni/DevCentral/_workitems/edit/2409381): set_analog_power_up_states() (without output type) returns errors with both LibraryInterpreter and GrpcStubInterpreter

Tests added.


keer1111:
@bkeryan , The `set_analog_power_up_states`  only supports setting- PowerUpChannelTypes - Voltage - 0, Current -1 and doesn't support High_Impedance -2 which is supported by `set_analog_power_up_states_with_output_type`.

Should we let this API be? or should we have separate enum to denote the PowerUpChannelTypes supported by this API?

keer1111:
Will create a separate PR for deprecation warnings

PR #415, opened 2023-06-19
Testing Pipeline | Added self hosted GH runner nidaqmxbot in pipeline

MounikaBattu17:

Runs the system tests in pipeline.

To validate system tests in GitHub pipeline

Testing in progress


tests/component/test_interpreter.py
bkeryan, line 1: What problem does this fix? I get the basic idea but I don't know exactly why it failed.

.github/workflows/github_actions_aws_win_py32.yml
bkeryan, line 1: Rename to `run_system_tests.yml` and delete the other file.

.github/workflows/github_actions_aws_win_py32.yml
```diff
@@ -0,0 +1,20 @@
+name: nidaqmxbot_win_32
+
+on:
+  pull_request:
+    branches:
+      - master
```
bkeryan, line 6: Call this from `PR.yml`. 

Change the trigger:
```suggestion
on:
  workflow_call:
  workflow_dispatch:
```

and add a job to `PR.yml`:
```
  run_system_tests:
    uses: ./.github/workflows/run_system_tests.yml
    needs: [build, run-unit-tests]
```

.github/workflows/github_actions_aws_win_py32.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  run-system-tests-python-32bit:
```
bkeryan:
```suggestion
  run_system_tests:
```

.github/workflows/github_actions_aws_win_py32.yml
```diff
@@ -0,0 +1,20 @@
+name: nidaqmxbot_win_32
```
bkeryan, line 1:
```suggestion
name: run_system_tests
```

.github/workflows/github_actions_aws_win_py32.yml
```diff
@@ -0,0 +1,20 @@
...
+  pull_request:
+    branches:
+      - master
+jobs:
+  run-system-tests-python-32bit:
+    runs-on:
+      - self-hosted
+      - windows
+      - x64
+      - rdss-nidaqmxbot-win-10-py32
```
bkeryan: This line and the name are the only differences between these files, so we should combine them and use a matrix.

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

Something like this:
```
runs-on:
  - self-hosted
  - rdss-nidaqmxbot-${{ matrix.configuration }}
strategy:
  matrix:
    configuration: [win-10-py32, win-10-py64]
```

I don't think it's necessary to specify runs-on: windows and runs-on: x64. Restricting to a specific runner should constrain the OS and CPU.

bkeryan (changes requested):
FYI @MounikaBattu17 @zhindes I am refactoring measurementlink-python's GitHub workflows: https://github.com/ni/measurementlink-python/pull/289/

My suggestions in this PR review are in line with those changes. I am interested in making the same refactors here.

.github/workflows/github_actions_aws_win_py32.yml
```diff
@@ -5,16 +5,18 @@ on:
...
-    timeout-minutes: 80
-    steps:
-      - uses: actions/checkout@v3
-      - name: Install dependencies
-        run: poetry install
-      - name: Run system tests
-        run: poetry run tox
\ No newline at end of file
+  do-the-job1:    
+    run-system-tests-python-32bit:
```
bkeryan, line 9: You can only define one level of job id.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id

tests/component/test_interpreter.py
MounikaBattu17: Some of the imports in `_grpc_interpreter.py` are not installed by `poetry `for `py-base` in `tox.ini`. So, I have made the `_grpc_interpreter` imports specific to only test methods.
image

bkeryan:
> FYI @MounikaBattu17 @zhindes I am refactoring measurementlink-python's GitHub workflows: [ni/measurementlink-python#289](https://github.com/ni/measurementlink-python/pull/289)
> 
> My suggestions in this PR review are in line with those changes. I am interested in making the same refactors here.

https://github.com/ni/nidaqmx-python/pull/418

MounikaBattu17:
@bkeryan The [workflow](https://github.com/ni/nidaqmx-python/actions/runs/5325188356/jobs/9645616754?pr=415) has been running for more than 30 minutes and I believe it doesn't run at all. The previous workflows for this PR also had the same issue of running either py32 or py64 or not running at all. Any thoughts on why this behavior?

.github/workflows/github_actions_aws_win_py32.yml
bkeryan: on second thought, let's use camel_case for ids and "Sentence case" for names

bkeryan:
> @bkeryan The [workflow](https://github.com/ni/nidaqmx-python/actions/runs/5325188356/jobs/9645616754?pr=415) has been running for more than 30 minutes and I believe it doesn't run at all. The previous workflows for this PR also had the same issue of running either py32 or py64 or not running at all. Any thoughts on why this behavior? or can we try running py64 only after py32 is ran successfully?

This is the point where we need additional assistance/access to debug the problem. I posted on the compute farm support channel in Teams.

.github/workflows/run_system_tests.yml
```diff
@@ -0,0 +1,21 @@
...
+      - self-hosted
+      - rdss-nidaqmxbot-${{ matrix.configuration }}
+    timeout-minutes: 80
+    strategy:
+      matrix:
+        configuration: [win-10-py32, win-10-py64]
+    steps:
+      - name: checkout repository
+        uses: actions/checkout@v3
+      - name: Install dependencies
```
bkeryan, line 18: This should also run `nidaqmxconfig` to import the MAX config. That way we don't need to rebuild the AMI when the MAX config changes.

You will need to update the AMI setup to add `nidaqmxconfig` to the system path, then come back to this in a separate PR.

tests/component/test_interpreter.py
MounikaBattu17: This tox failure was not captured while building nidaqmxbot since the change merged to master only after successful deployment of VM, I have created a separate PR to track [https://github.com/ni/nidaqmx-python/pull/419](url) and to get merged without any conflicts.

.github/workflows/github_actions_aws_win_py32.yml
MounikaBattu17: Thanks for the information! Updated.

.github/workflows/github_actions_aws_win_py32.yml
MounikaBattu17: Removed the second level job id.

.github/workflows/run_system_tests.yml
```diff
@@ -0,0 +1,21 @@
+name: Run system tests
+
+on:
+  workflow_call:
+  workflow_dispatch:
+jobs:
+  run_system_tests:
+    runs-on:
```
bkeryan, line 8: [cosmetic] Give this a name so GH doesn't display the snake_case id

```suggestion
  run_system_tests:
    name: Run system tests
    runs-on:
```

.github/workflows/run_system_tests.yml
```diff
@@ -0,0 +1,21 @@
...
+      - self-hosted
+      - rdss-nidaqmxbot-${{ matrix.configuration }}
+    strategy:
+      matrix:
+        configuration: ["win-10-py32", "win-10-py64"]
+    timeout-minutes: 80
+    steps:
+      - name: checkout repository
```
bkeryan, line 16: consistency with other workflows
```suggestion
      - name: Check out repo
```

bkeryan (approved):
Looks good. Now we need to debug why the self-hosted runner is not connecting.

zhindes (approved):

tests/component/test_interpreter.py
MounikaBattu17: Resolving this here.

.github/workflows/run_system_tests.yml
MounikaBattu17: @bkeryan I have created a **[Technical Debt 2436601](https://dev.azure.com/ni/DevCentral/_workitems/edit/2436601): Import nidaqmxconfig in GitHub workflow** to track this item.

PR #419, opened 2023-06-20
Moved grpc_interpreter specific imports inside the required test methods

MounikaBattu17:

The `poetry run tox` fails for `py{37, 38, 39 ,310, 311}-base` due to some of the imports of `_grpc_interpreter` in `test_interpreter.py`, So moved these imports inside the required test methods.

To successfully run `tox`.

Ran `poetry run tox`, all python versions pased.

### DIFF
--- a/tests/component/test_interpreter.py
+++ b/tests/component/test_interpreter.py
@@ -3,15 +3,10 @@ import pytest
 from nidaqmx._base_interpreter import BaseInterpreter
 from nidaqmx.error_codes import DAQmxErrors
 
-
-def get_error_messages():
-    try:
-        from nidaqmx._grpc_interpreter import _ERROR_MESSAGES
-
-        error_messages = _ERROR_MESSAGES
-    except ImportError:
-        error_messages = {}
-    return error_messages
+try:
+    from nidaqmx._grpc_interpreter import _ERROR_MESSAGES
+except ImportError:
+    _ERROR_MESSAGES = {}
 
 
 def test___known_error_code___get_error_string___returns_known_error_message(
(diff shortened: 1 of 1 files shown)


tests/component/test_interpreter.py
```diff
@@ -1,13 +1,17 @@
...
 from nidaqmx._base_interpreter import BaseInterpreter
-from nidaqmx._grpc_interpreter import GrpcStubInterpreter
 from nidaqmx.error_codes import DAQmxErrors
 
-try:
-    from nidaqmx._grpc_interpreter import _ERROR_MESSAGES
-except ImportError:
-    _ERROR_MESSAGES = {}
+
+def get_error_messages():
```
bkeryan, line 7: I don't think this function is necessary. I think it's sufficient to move the import of GrpcStubInterpreter into the try block or into the test function.

bkeryan (changes requested):

tests/component/test_interpreter.py
```diff
@@ -36,10 +40,13 @@ def test___grpc_channel_with_errors___get_error_string___returns_failed_to_retri
     assert error_message.startswith("Failed to retrieve error description.")
 
 
-@pytest.mark.parametrize("error_code", list(_ERROR_MESSAGES))
+@pytest.mark.parametrize("error_code", list(get_error_messages()))
```
MounikaBattu17, line 43: `from nidaqmx._grpc_interpreter import _ERROR_MESSAGES`

`_ERROR_MESSAGES` is used here as well as inside the test function, so importing inside the function is making it not accessible for `parametrize` to use.

MounikaBattu17: Removed the `get_error_messages()` method.

tests/component/test_interpreter.py
MounikaBattu17: Removed the get_error_messages() method.

tests/component/test_interpreter.py
```diff
@@ -40,6 +39,8 @@ def test___grpc_channel_with_errors___get_error_string___returns_failed_to_retri
 def test___known_error_codes___get_error_string_with_error_code___returns_matching_error_message(
     grpc_init_kwargs, error_code: int
 ):
+    from nidaqmx._grpc_interpreter import GrpcStubInterpreter
+
     interpreter = GrpcStubInterpreter(**grpc_init_kwargs)
```
bkeryan, line 44: FYI, I'm not a fan of how `black` adds a newline here, because it splits the test function into 4 sections (import, arrange, act, assert). Perhaps moving the import inside the try block is slightly better. However, it's not a dealbreaker.

bkeryan (approved):

tests/component/test_interpreter.py
MounikaBattu17: Moved to `try` block.

PR #429, opened 2023-06-30
test: update test to verify data returned after partial read

keer1111:

Updates the test_timeout_stream test case to verify the partial read data on timeout exception
- Removed sentinal_value since grpc call is not passed by reference and hence cannot retain the sentinal_value
- Updated the test case to setup and verify multiple channels, so as to showcase the way to extract multiple channel data and verify (we have to extract properly from the partial read data, since DAQmx overwrites first channel's data with other channel's data in a partial read)

To verify the read partial data by properly extracting from the returned data of various channels and validate - as discussed here [Bug 2420742](https://dev.azure.com/ni/DevCentral/_workitems/edit/2420742): nidaqmx-python read timeout test interprets partial read data incorrectly

Test passed in the remote PC with hardware

### DIFF
--- a/tests/legacy/test_read_exceptions.py
+++ b/tests/legacy/test_read_exceptions.py
@@ -137,9 +137,6 @@ class TestReadExceptions:
         number_of_samples_expected = clocks_to_give - samples_to_read
         assert timeout_exception.value.samps_per_chan_read == number_of_samples_expected
 
-    @pytest.mark.grpc_xfail(
-        reason="AB#2420742: gRPC interpreter doesn't retain the initial array values when converting flattened array into numpy array."
-    )
     def test_timeout_stream(self, generate_task, real_x_series_device):
         """Test for validating read timeout."""
         # USB streaming is very tricky.
@@ -149,14 +146,20 @@ class TestReadExceptions:
         ):
             pytest.skip("Requires a plugin device.")
 
+        number_of_channels = 4
         samples_to_read = 75
         clocks_to_give = 100
         sample_rate = 1000
-        # Choose a sentinel value that can't be returned by the read.
-        sentinel_value = -100.0
 
+        write_task = generate_task()
         read_task = generate_task()
         sample_clk_task = generate_task()
+
+        write_task.ao_channels.add_ao_voltage_chan(
+            f"{real_x_series_device.name}/ao0:{number_of_channels - 1}"
+        )
+        write_task.write([float(i + 1) for i in range(number_of_channels)], auto_start=True)
+
         # Use a counter output pulse train task as the sample clock source
         # for both the AI and AO tasks.
         sample_clk_task.co_channels.add_co_pulse_chan_freq(
@@ -170,13 +173,17 @@ class TestReadExceptions:
         samp_clk_terminal = f"/{real_x_series_device.name}/Ctr0InternalOutput"
 
         read_task.ai_channels.add_ai_voltage_chan(
-            real_x_series_device.ai_physical_chans[0].name, max_val=10, min_val=-10
+            ",".join(
+                f"{real_x_series_device.name}/_ao{i}_vs_aognd" for i in range(number_of_channels)
+            ),
+            max_val=10,
+            min_val=-10,
         )
         read_task.timing.cfg_samp_clk_timing(
-            sample_rate, source=samp_clk_terminal, sample_mode=AcquisitionType.CONTINUOUS
+            sample_rate, source=samp_clk_terminal, sample_mode=AcquisitionType.FINITE
         )
 
-        reader = nidaqmx.stream_readers.AnalogSingleChannelReader(read_task.in_stream)
+        reader = nidaqmx.stream_readers.AnalogMultiChannelReader(read_task.in_stream)
 
         # Start the read task before the clock.
         read_task.start()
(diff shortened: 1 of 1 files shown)


tests/legacy/test_read_exceptions.py
```diff
@@ -208,8 +216,11 @@ def test_timeout_stream(self, generate_task, real_x_series_device):
...
-        # The data that wasn't read should have been unmodified.
-        assert all(element == sentinel_value for element in data[number_of_samples_expected:])
+        number_of_samples_read = timeout_exception.value.samps_per_chan_read
+        assert number_of_samples_read == number_of_samples_expected
+
+        # DAQmx overwrites first channel array slices with other channel data as in AB#2420742
+        # Hence resize of the data is needed to extract each data correctly
+        resized_data = numpy.resize(data, (number_of_channels, number_of_samples_read))
+        for i in range(number_of_channels):
+            assert all(i + 1 == round(element) for element in resized_data[i])
```
bkeryan, line 226: Rounding to integer is too tolerant. Use `element == pytest.approx(i + 1, abs=1e-3)` to compare with 1 mV tolerance.

tests/legacy/test_read_exceptions.py
```diff
@@ -149,14 +146,20 @@ def test_timeout_stream(self, generate_task, real_x_series_device):
         ):
             pytest.skip("Requires a plugin device.")
 
+        number_of_channels = 4
```
zhindes, line 149: not all `real_x_series` devices have 4 AO channels. Some only have 2, and a few have 0. If possible, I'd reduce this to 2 and skip if its 0. If that doesn't work, then skip if not 4 is fine, too.

tests/legacy/test_read_exceptions.py
```diff
@@ -190,16 +197,17 @@ def test_timeout_stream(self, generate_task, real_x_series_device):
...
-        data = numpy.full(samples_to_read, sentinel_value, dtype=numpy.float64)
+        data = numpy.zeros((number_of_channels, samples_to_read), dtype=numpy.float64)
         num_values_read = reader.read_many_sample(
             data, number_of_samples_per_channel=samples_to_read, timeout=2.0
         )
+
         assert num_values_read == samples_to_read
         # All the data should have been overwritten.
-        assert not any(element == sentinel_value for element in data)
+        assert not any(element == 0 for element in data.reshape(data.size))
```
bkeryan, line 207: This might as well validate that each channel has the right voltage.

tests/legacy/test_read_exceptions.py
zhindes, line 226: `round` is clever, but I think `pytest.approx` with a pretty large rel value (like 0.2V or something) is probably more appropriate

zhindes (approved):
minor issues

tests/legacy/test_read_exceptions.py
bkeryan, line 226: Use NumPy slicing: `assert (resized_data[i, :] == pytest.approx(...)).all()`

tests/legacy/test_read_exceptions.py
```diff
@@ -149,14 +146,20 @@ def test_timeout_stream(self, generate_task, real_x_series_device):
...
-        sentinel_value = -100.0
 
+        write_task = generate_task()
         read_task = generate_task()
         sample_clk_task = generate_task()
+
+        write_task.ao_channels.add_ao_voltage_chan(
+            f"{real_x_series_device.name}/ao0:{number_of_channels - 1}"
+        )
+        write_task.write([float(i + 1) for i in range(number_of_channels)], auto_start=True)
```
bkeryan, line 161: Save the list of floats in a variable so you don't have to duplicate the +1

tests/legacy/test_read_exceptions.py
```diff
@@ -137,9 +137,6 @@ def test_timeout_raw(self, generate_task, real_x_series_device):
         number_of_samples_expected = clocks_to_give - samples_to_read
         assert timeout_exception.value.samps_per_chan_read == number_of_samples_expected
 
-    @pytest.mark.grpc_xfail(
-        reason="AB#2420742: gRPC interpreter doesn't retain the initial array values when converting flattened array into numpy array."
-    )
     def test_timeout_stream(self, generate_task, real_x_series_device):
```
bkeryan, line 140: FYI, I can't think of a strong reason to keep a single-channel version or parameterize on # channels

bkeryan (changes requested):

tests/legacy/test_read_exceptions.py
bkeryan: Use abs to specify a tolerance in V / mV. 

rel is relative to the value being compared.

tests/legacy/test_read_exceptions.py
bkeryan: 4 is a much better demonstration of how DAQmx packs the samples together.

How about using the `Device` object to query how many AO channels are supported? call `pytest.skip(reason)` if 0.

Also skip if `device.ai_simultaneous_sampling_supported` because those devices have a different way to program AO loopback (acquire from Dev1/ai0:3 with `ai_input_src="_ao0_vs_aognd"` etc.)

bkeryan:
> This pull request removes 3250 and adds 3250 tests. Note that renamed tests count towards both.

Oops, the random seed randomizes the test names ☹️ 

Cc: @zhindes

bkeryan:
> > This pull request removes 3250 and adds 3250 tests. Note that renamed tests count towards both.
> 
> Oops, the random seed randomizes the test names ☹️

https://github.com/ni/nidaqmx-python/pull/431

tests/legacy/test_read_exceptions.py
keer1111: The pytest.approx returns a ApproxScaler type and is not working in all(), instead the whole expression is evaluated to bool and tries to call all() on top of it. whereas values types like int, float work in the statement.

keer1111: approximated to 2 decimal places, since there are discrepancies above that.

bkeryan: ok, iterating is fine then

bkeryan (approved):

PR #433, opened 2023-07-10
Consistently return DaqNotFoundError when NI-DAQmx is not installed on all platforms

zhindes:

Ensures `nidaqmx` consistently uses `DaqNotFoundError` on Windows. Previously, it would leak ctypes' `OSError` or `WindowsError`. Note: `WindowsError` is now an alias for `OSError`, but they were separate before Python 3.3. IronPython 3.4 seems to indicate its ctypes implementation still uses `WindowsError`, so I am catching both. To be honest, I'm not sure if we care about IronPython, but whatever.

Other small improvement:
* Single source common error strings

Improves UX when using `nidaqmx` without the `NI-DAQmx` driver installed.

Validated error reporting using `system_properties.py` example on a system without NI-DAQmx installed:

```
(.venv) C:\Users\User>python system_properties.py
  File "C:\Users\User\.venv\lib\site-packages\nidaqmx\_lib.py", line 150, in _import_lib
    windll = ctypes.windll.LoadLibrary('nicaiu')
  File "C:\Users\User\.pyenv\pyenv-win\versions\3.7.9\lib\ctypes\__init__.py", line 442, in LoadLibrary
    return self._dlltype(name)
  File "C:\Users\User\.pyenv\pyenv-win\versions\3.7.9\lib\ctypes\__init__.py", line 364, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: [WinError 126] The specified module could not be found

The above exception was the direct cause of the following exception:

  File "system_properties.py", line 5, in <module>
    driver_version = local_system.driver_version
  File "C:\Users\User\.venv\lib\site-packages\nidaqmx\system\system.py", line 89, in driver_version
    return System.DriverVersion(self._major_version, self._minor_version,
  File "C:\Users\User\.venv\lib\site-packages\nidaqmx\system\system.py", line 122, in _major_version
    val = self._interpreter.get_system_info_attribute_uint32(0x1272)
  File "C:\Users\User\.venv\lib\site-packages\nidaqmx\_library_interpreter.py", line 3272, in get_system_info_attribute_uint32
    cfunc = lib_importer.cdll.DAQmxGetSystemInfoAttribute
  File "C:\Users\User\.venv\lib\site-packages\nidaqmx\_lib.py", line 119, in cdll
    self._import_lib()
  File "C:\Users\User\.venv\lib\site-packages\nidaqmx\_lib.py", line 153, in _import_lib
    raise DaqNotFoundError(DAQ_NOT_FOUND_MESSAGE) from e
nidaqmx.errors.DaqNotFoundError: Could not find an installation of NI-DAQmx. Please ensure that NI-DAQmx is installed on this machine or contact National Instruments for support.
```

Ran `pytest` on a system that has NI-DAQmx installed:

```
...
=============== 973 passed, 114 skipped, 54 xfailed in 100.67s (0:01:40) =============== 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ All notable changes to this project will be documented in this file.
 * ### Merged Pull Requests
     * ...
 * ### Major Changes
-    * ...
+    * Promoted `DaqNotFoundError`, `DaqNotSupportedError`, and `DaqFunctionNotSupportedError` to `nidaqmx.errors`.
+    * Consistently return `DaqNotFoundError` on all platforms when the NI-DAQmx driver is not installed.
 * ### Known Issues
     * ...
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,10 @@ All notable changes to this project will be documented in this file.
 * ### Merged Pull Requests
     * ...
 * ### Major Changes
-    * Promoted `DaqNotFoundError`, `DaqNotSupportedError`, and `DaqFunctionNotSupportedError` to `nidaqmx.errors`.
-    * Consistently return `DaqNotFoundError` on all platforms when the NI-DAQmx driver is not installed.
+    * `nidaqmx.errors.DaqNotFoundError`, `nidaqmx.errors.DaqNotSupportedError`, and
+    `nidaqmx.errors.DaqFunctionNotSupportedError` are now public exceptions.
+    * Consistently return `nidaqmx.errors.DaqNotFoundError` on all platforms when the NI-DAQmx
+    driver is not installed.
 * ### Known Issues
     * ...
 
--- a/generated/nidaqmx/_lib.py
+++ b/generated/nidaqmx/_lib.py
@@ -9,14 +9,14 @@ from typing import cast
 from nidaqmx.errors import DaqNotFoundError, DaqNotSupportedError, DaqFunctionNotSupportedError
 
 
-DAQ_NOT_FOUND_MESSAGE = "Could not find an installation of NI-DAQmx. Please ensure that NI-DAQmx " \
-                        "is installed on this machine or contact National Instruments for support."
+_DAQ_NOT_FOUND_MESSAGE = "Could not find an installation of NI-DAQmx. Please ensure that NI-DAQmx " \
+                         "is installed on this machine or contact National Instruments for support."
 
-DAQ_NOT_SUPPORTED_MESSAGE = "NI-DAQmx Python is not supported on this platform: {0}. Please " \
-                            "direct any questions or feedback to National Instruments."
+_DAQ_NOT_SUPPORTED_MESSAGE = "NI-DAQmx Python is not supported on this platform: {0}. Please " \
+                             "direct any questions or feedback to National Instruments."
 
-FUNCTION_NOT_SUPPORTED_MESSAGE = "The NI-DAQmx function \"{0}\" is not supported in this version " \
-                                 "of NI-DAQmx. Visit ni.com/downloads to upgrade."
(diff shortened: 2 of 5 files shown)


generated/nidaqmx/_lib.py
```diff
@@ -146,29 +162,25 @@ def _import_lib(self):
...
 
-        if sys.platform.startswith('win') or sys.platform.startswith('cli'):
-            if 'iron' in platform.python_implementation().lower():
-                windll = ctypes.windll.nicaiu
-                cdll = ctypes.cdll.nicaiu
-            else:
-                windll = ctypes.windll.LoadLibrary('nicaiu')
-                cdll = ctypes.cdll.LoadLibrary('nicaiu')
-
+        if sys.platform.startswith('win'):
```
bkeryan, line 165: FYI: the "cli" check was for IronPython. 
https://www.github.com/IronLanguages/ironpython3/issues/395

zhindes: Ah, good to know, thanks!

zhindes:
@bkeryan -Just now getting a chance to look at this again. Any other concerns?

generated/nidaqmx/_lib.py
```diff
@@ -146,29 +162,25 @@ def _import_lib(self):
...
+            try:
+                if 'iron' in platform.python_implementation().lower():
```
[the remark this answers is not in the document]
bkeryan: If you're not removing IronPython, should you keep the "cli" check?

bkeryan:
> @bkeryan -Just now getting a chance to look at this again. Any other concerns?

Should these exception classes be part of the public API (i.e. `nidaqmx.errors`)? Currently they are part of the private `_lib` sub-module, so clients can't catch specific exceptions.

I don't know if hardcoding the message in the `__init__` method without allowing the caller to override the message is a best practice or not. nimi-python does it, though. Cc: @mshafer-NI

mshafer-NI:
> > @bkeryan -Just now getting a chance to look at this again. Any other concerns?
> 
> Should these exception classes be part of the public API (i.e. `nidaqmx.errors`)? Currently they are part of the private `_lib` sub-module, so clients can't catch specific exceptions.
> 
> I don't know if hardcoding the message in the `__init__` method without allowing the caller to override the message is a best practice or not. [nimi-python does it](https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/errors.py), though. Cc: @mshafer-NI

It's common to put exception classes in public API so users can catch them.

Don't hard code the message, make it a default.

```python
class ValidationError(Exception):
    def __init__(self, message="default"):            
        # Call the base class constructor with the parameters it needs
        super().__init__(message)
```

CHANGELOG.md
```diff
@@ -23,7 +23,8 @@ All notable changes to this project will be documented in this file.
 * ### Merged Pull Requests
     * ...
 * ### Major Changes
-    * ...
+    * Promoted `DaqNotFoundError`, `DaqNotSupportedError`, and `DaqFunctionNotSupportedError` to `nidaqmx.errors`.
```
bkeryan, line 26: It may not be clear that "promoted" means "made public"

generated/nidaqmx/_lib.py
```diff
@@ -6,19 +6,17 @@
 import threading
 from typing import cast
 
-from nidaqmx.errors import Error
+from nidaqmx.errors import DaqNotFoundError, DaqNotSupportedError, DaqFunctionNotSupportedError
 
 
-class DaqNotFoundError(Error):
-    pass
+DAQ_NOT_FOUND_MESSAGE = "Could not find an installation of NI-DAQmx. Please ensure that NI-DAQmx " \
```
bkeryan, line 12: These constants are only used in this module so I recommend making them private.

bkeryan (approved):

PR #435, opened 2023-07-11
tests: Update a gRPC specific test case so that `pytest -k "not grpc"` skips it

bkeryan:

Update a gRPC specific test case to use `@pytest.mark.grpc_only` and `init_kwargs` so `pytest -k "not grpc"` skips it. 

This works because the `init_kwargs` parameterization appends `[grpc_init_kwargs]` to the test case name.

Remove a skipped test from pyXX-base runs.

Ran `pytest -vvv tests/component/test_interpreter.py` with and without `-k "not grpc"`.

### DIFF
--- a/tests/component/test_interpreter.py
+++ b/tests/component/test_interpreter.py
@@ -37,12 +37,12 @@ def test___grpc_channel_with_errors___get_error_string___returns_failed_to_retri
 
 
 @pytest.mark.parametrize("error_code", list(_ERROR_MESSAGES))
-def test___known_error_codes___get_error_string_with_error_code___returns_matching_error_message(
+def test___grpc_interpreter_with_known_error_message___get_error_string___returns_matching_error_message(
     grpc_init_kwargs, error_code: int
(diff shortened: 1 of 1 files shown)


tests/component/test_interpreter.py
```diff
@@ -37,12 +37,12 @@ def test___grpc_channel_with_errors___get_error_string___returns_failed_to_retri
 
 
 @pytest.mark.parametrize("error_code", list(_ERROR_MESSAGES))
-def test___known_error_codes___get_error_string_with_error_code___returns_matching_error_message(
+def test___grpc_interpreter_with_known_error_message___get_error_string___returns_matching_error_message(
     grpc_init_kwargs, error_code: int
-):
+) -> None:
     interpreter = GrpcStubInterpreter(**grpc_init_kwargs)
```
bkeryan, line 43: On second thought, we should mark this test grpc_only

zhindes (approved):

PR #449, opened 2023-08-21
Add lib and grpc time data types and conversions to enable future NI-DAQmx time support

zhindes:

It proposes one new data type:
* `nidaqmx._lib_time.AbsoluteTime` implements NI-BTF and conversion to and from Python's `datetime` and NI's `hightime` (see [hightime](https://github.com/ni/hightime)) types

It also provides conversions to and from gRPCs `Timestamp` type:
* `nidaqmx._grpc_time.convert_time_to_timestamp` and `nidaqmx._grpc_time.convert_timestamp_to_time` provide conversion to and from `hightime` type (in addition to maintaining `datetime` support)

The proposal is that our user-facing API would all use `datetime` or `hightime` interchangeably, and we would use these types under-the-hood to call into the C API or gRPC. One minor issue I see:

* Non-UTC timezones as out-params is something to think about. I think only returning UTC is good enough, but we could also add in-params that tell us what timezone to use. That doesn't work well with attributes, though, so my hunch is to stick with UTC-only as out-params and the user can convert.
* `hightime` as out-params. I think its probably fine to only return `hightime` since they inherit from `datetime`, so it all works out.

Trying to get some base functionality in place so future PRs can start adding time support.

New component tests.


src/handwritten/time.py
```diff
@@ -0,0 +1,68 @@
+import ctypes
+import functools
+from datetime import datetime, timezone
+
+@functools.total_ordering
+class AbsoluteTime(ctypes.Structure):
```
zhindes, line 6: On further reflection, I don't think users should actually interact with this type for now. Assuming that passes muster, where should it go? @bkeryan - thoughts?

tests/component/test_time.py
```diff
@@ -0,0 +1,85 @@
+from copy import copy
+import pytest
+import random
+
+from datetime import datetime, timedelta, timezone
+from nidaqmx.time import AbsoluteTime
+
+# Jan 1, 2002 = 98 years + 25 leapdays = 35,795 days = 3,092,688,000 seconds
+JAN_01_2022_BTF = AbsoluteTime(lsb=0, msb=0xb856ac80)
```
zhindes, line 9: This is the one test case I found internally, in nitypest component. Fun!

src/handwritten/time.py
bkeryan: This is ctypes-specific, so it belongs in `_lib.py`.

I think the public API and interpreter interfaces should use the standard Python time types, NI's `hightime` package, or both:
- https://pypi.org/project/hightime/
- https://github.com/ni/hightime/tree/master

Note that `hightime` doesn't have the NI-BTF binary layout, so it will require conversion.

`nidaqmx.proto` uses `google.protobuf.Timestamp`, which already has a Python class with conversion methods.
https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/well_known_types.py#L98

zhindes: Ooh, nice, thanks for the info! `hightime` is great, and since we have precedence with nimi-python, I'll shift to that. I'll still need this class for the lib version, and then something similar for gRPC to convert to `google.protobuf.Timestamp` to maintain at least nanosecond precision which that class supports. I'll add that, too.

tests/component/test_time.py
```diff
@@ -0,0 +1,277 @@
...
+    # now convert back
+    to_dt = ts.to_datetime(tzinfo=timezone.utc)
+
+    to_dt_microsecond = to_dt.microsecond
+    # comparison is tricky since imprecision in the conversion to NI-BTF are
+    # caught by the higher precision values in hightime, so we round here.
+    if to_dt.femtosecond > LibTimestamp.MAX_FS / 2:
+        to_dt_microsecond += 1
+
+    assert to_dt_microsecond == microsecond
```
bkeryan, line 112: The test body should have 3 blocks, separated by blank lines: arrange, act, assert

If the middle "when" clause contains "convert_to_and_from", then that code belongs in the middle "act" block.

Please look over the regression test conventions and unit test conventions again.

If you feel the need to write comments with a running commentary, consider writing helper functions instead.

bkeryan (changes requested):

tests/component/test_time.py
bkeryan, line 1: Organize component/unit tests like the code under tests. Consider splitting this into `test_grpc_time.py` and `test_lib_time.py` to match the submodules.

tests/component/test_time.py
```diff
@@ -0,0 +1,277 @@
...
+    assert to_dt.yoctosecond == yoctosecond_round_trip
+
+
+# Note: I can't actually test overflow because Python's datetime object is
+# limited to years 1-9999, but the NI-BTF format can represent time before the
+# Big Bang and until about year 292 billion. Oh well.
+
+
+@pytest.mark.parametrize("from_dt", [(JAN_01_2022_DATETIME), (JAN_01_2022_HIGHTIME)])
+def test___grpctimestamp___convert_from_utc___succeeds(from_dt):
```
bkeryan, line 172: The gRPC tests need to be skipped when gRPC isn't available

poetry.lock
```diff
@@ -972,763 +1667,6 @@ docs = ["Sphinx", "sphinx_rtd_theme"]
 grpc = ["grpcio", "grpcio", "protobuf"]
 
 [metadata]
-lock-version = "1.1"
+lock-version = "2.0"
```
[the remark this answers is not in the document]
zhindes: Bah, my bad. Somehow my poetry is up to 1.5 or something. I'll redo with 1.2.2

tests/component/test_time.py
zhindes: > Please look over the regression test conventions and unit test conventions again.

Good recommendation! I think I have it better now, but I have low confidence it is correct. Thanks for the help!

tests/component/test_time.py
zhindes: They're skipped in the CI/PR since it runs with `not grpc`. Is there something else I should do explicitly?

tests/component/test_time.py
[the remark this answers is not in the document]
zhindes: I ran out of time, but I'll get to this. I am pushing all the other fixes up since they're significant.

generated/nidaqmx/_grpc_time.py
```diff
@@ -0,0 +1,40 @@
+from datetime import timezone
+from hightime import datetime as ht_datetime
+
+from google.protobuf.internal.well_known_types import Timestamp as grpc_Timestanp
+
+
+class Timestamp(grpc_Timestanp):
```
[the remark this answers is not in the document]
bkeryan: Yeah, protobuf returns a `timestamp_pb2.Timestamp`, not a `google.protobuf.internal.well_known_types.Timestamp`, so we can't use these conversion functions directly (without also creating a well_known_types.Timestamp).

```
PS D:\dev\measurementlink-python\examples\nidaqmx_analog_input> poetry run python
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
>>> import nidaqmx._stubs.nidaqmx_pb2 as nidaqmx_pb2
>>> req = nidaqmx_pb2.CfgTimeStartTrigRequest()
>>> type(req.when)
<class 'google.protobuf.timestamp_pb2.Timestamp'>
```

tests/unit/_time_utils.py
```diff
@@ -0,0 +1,12 @@
...
+
+from hightime import datetime as ht_datetime
+
+# Jan 1, 2002 = 32 years + 8 leapdays = 11688 days = 1009843200 seconds
+JAN_01_2022_TIMESTAMP_1970_EPOCH = 0x3C30FC00
+# Jan 1, 2002 = 98 years + 25 leapdays = 35795 days = 3092688000 seconds
+JAN_01_2022_TIMESTAMP_1904_EPOCH = 0xB856AC80
+
+JAN_01_2022_DATETIME = std_datetime(2002, 1, 1, tzinfo=timezone.utc)
+JAN_01_2022_HIGHTIME = ht_datetime(2002, 1, 1, tzinfo=timezone.utc)
```
bkeryan, line 12: 2022 or 2002?

src/handwritten/_grpc_time.py
```diff
@@ -0,0 +1,48 @@
...
+
+_NS_PER_S = 10**9
+_NS_PER_US = 10**3
+
+_YS_PER_US = 10**18
+_YS_PER_NS = 10**15
+_YS_PER_FS = 10**9
+
+
+def convert_time_to_timestamp(dt, ts):
```
bkeryan, line 15: Sorry to bring this up late, but consider adding type hints in order to document the signature. I expected this to return a timestamp object, but it mutates the one you pass in, probably because protobuf in Python doesn't allow reassigning child messages after construction.

```suggestion
def convert_time_to_timestamp(dt: ht_datetime, ts: GrpcTimestamp) -> None:
```

tests/component/test_time.py
bkeryan: Oh, I see. I forgot the "base" tox configuration filters out grpc tests to avoid unnecessary skipped tests.

`@pytest.mark.grpc_only(...)` or `@pytest.mark.grpc_skip(...)` won't work because those markers are only for init_kwargs.

It sounds like `-k not grpc` matches against both filenames and test names, then I guess it's fine to leave it as-is.

If it only matched test names, you would still need the try/except around the gRPC imports so that pytest could import the file and find out the test names. BTW, the try/except could probably be used to skip the whole module:
```
try:
    import grpcio
except ImportError:
    pytest.skip(allow_module_level=True, reason="Requires gRPC")
```

bkeryan (changes requested):
The important remaining issue is whether we should be converting to/from `google.protobuf.timestamp_pb2.Timestamp` or `google.protobuf.internal.well_known_types.Timestamp`. 

These two new issues are not that important.

src/handwritten/_lib_time.py
```diff
@@ -54,7 +56,7 @@ def from_datetime(cls, dt):
 
         return AbsoluteTime(lsb=lsb, msb=timestamp_1904_epoch)
 
-    def to_datetime(self, tzinfo=None):
+    def to_datetime(self, tzinfo: timezone = None):
```
bkeryan, line 59:
```suggestion
    def to_datetime(self, tzinfo: timezone = None) -> hd_datetime:
```

src/handwritten/_lib_time.py
```diff
@@ -0,0 +1,97 @@
+import ctypes
```
bkeryan, line 1:
```suggestion
from __future__ import annotations
import ctypes
```

src/handwritten/_lib_time.py
```diff
@@ -0,0 +1,97 @@
...
+    _US_PER_S = 10**6
+    _YS_PER_S = 10**24
+    _YS_PER_US = 10**18
+    _YS_PER_FS = 10**9
+
+    MAX_FS = 10**9
+    MAX_YS = 10**9
+
+    @classmethod
+    def from_datetime(cls, dt: Union[std_datetime, ht_datetime]) -> "AbsoluteTime":
```
bkeryan, line 31: If you use `from __future__ import annotations` then you don't have to quote forward references.

```suggestion
    def from_datetime(cls, dt: Union[std_datetime, ht_datetime]) -> AbsoluteTime:
```

bkeryan (approved):

src/handwritten/_grpc_time.py
```diff
@@ -0,0 +1,50 @@
...
+        # round up, if necessary
+        if remainder_yoctoseconds >= _YS_PER_NS / 2:
+            nanos += 1
+    else:
+        nanos = utc_dt.microsecond * _NS_PER_US
+
+    ts.FromNanoseconds(seconds * _NS_PER_S + nanos)
+
+
+def convert_timestamp_to_time(ts: GrpcTimestamp, tzinfo: timezone = None) -> ht_datetime:
```
bkeryan, line 35: Oops, `None` is not a `timezone`. This should be `tzinfo: Optional[timezone] = None` in both files.

PR #450, opened 2023-09-01
tests: Update GrpcServerProcess to ignore additional output

bkeryan:

Update `GrpcServerProcess` to ignore additional output.

This is required to handle the additional log output added in https://github.com/ni/grpc-device/pull/989

Also see https://github.com/ni/nimi-python/pull/2006 

Enable tests to pass with the next version of NI gRPC Device Server.

Manually tested with a local build of NI gRPC Device Server.
Added a bogus argument "-asdf" to test the case where the server returns an error. 
At one point I ran the local build without the VS2017 debug CRT installed, and this also caused a hang, but I can't reproduce it because I now have the debug CRT installed.

### DIFF
--- a/tests/_grpc_utils.py
+++ b/tests/_grpc_utils.py
@@ -21,7 +21,7 @@ class GrpcServerProcess:
             self.server_port = None
             while self.server_port is None and self._proc.poll() is None:
                 line = self._proc.stdout.readline()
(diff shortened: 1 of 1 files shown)


zhindes (approved):

bkeryan:
The issue with the self-hosted GitHub runner is not related to this change.

PR #458, opened 2023-11-17
Create channels.py

StSav012:
Create the `nidaqmx/channels.py` file to import the channel types gracefully.

The file improves typing annotations. Without the file, there are warnings about accessing `_task_modules`, which is a private module, when getting the `Channel` class or its subclasses.

Expose channel classes from `nidaqmx._task_modules.channels`, primarily for typing annotations, but not limiting to it.

As Python has adopted optional typing annotations, various IDEs help writing code by providing the methods available for a variable. This significantly increases the speed of the code writing, without damaging the quality. Unfortunately, without the file I propose, I can't annotate, for instance, what the `Task.ai_channels.add_ai_voltage_chan()` function returns.

No regression testing has been done by me, for it requires a piece of hardware I'm not allowed to use for the purpose of the testing. However, I do use the code in my copy of `nidaqmx` for more than a year, and it has not caused any troubles.

### DIFF
--- a/src/handwritten/channels.py
+++ b/src/handwritten/channels.py
@@ -1,0 +1,9 @@
+from nidaqmx._task_modules.channels.channel import Channel
+from nidaqmx._task_modules.channels.ai_channel import AIChannel
+from nidaqmx._task_modules.channels.ao_channel import AOChannel
+from nidaqmx._task_modules.channels.ci_channel import CIChannel
+from nidaqmx._task_modules.channels.co_channel import COChannel
+from nidaqmx._task_modules.channels.di_channel import DIChannel
+from nidaqmx._task_modules.channels.do_channel import DOChannel
+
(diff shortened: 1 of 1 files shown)


bkeryan:
Hi @StSav012, thank you for your contribution. Making types public to support type hints is a prerequisite for #209 and I recognize that it has value even before type hints are added for the `nidaqmx` module itself.

I have been using type hints for nidaqmx-python and nimi-python types in another project. Mypy treats all of the types from these packages as `Any` due to `--ignore-missing-imports`, but type hints still improve the accuracy of autocomplete in Visual Studio Code and other LSP-aware editors.

I'm currently out of office for the Thanksgiving holiday. I would like to sync up with the other nidaqmx-python code owners regarding where the other types (channel collections, timing/triggering, etc.) belong and whether we should create aliases as you have proposed or reorganize the source tree to make the internal submodules public.

Issue #23, opened 2017-08-18
Support querying device calibration information and Self-Cal


epage:
I don't think this stuff is supported in the API.

See http://zone.ni.com/reference/en-XX/help/370471AA-01/mxcprop/attributeclasscalibrationinfo/

spalatofhi:
Has self calibration (`DAQmxSelfCal`) been added to the API?

spalatofhi:
This seems to work, for self-calibration at least. It doesn't raise an error on a simulated device...

```
def self_cal(device: str):
    cfunc = lib_importer.windll.DAQmxSelfCal
    if cfunc.argtypes is None:
        with cfunc.arglock:
            cfunc.argtypes = [ctypes_byte_str]
    retcode = cfunc(device)
    check_for_error(retcode)
    return True # success!
```

flef:
Hey, 

Can you propose a Pull Request with this change ? I'm also looking for DAQmxSelfCal function in the API.

Regards

zhindes:
@flef - this is a 6-year old Feature request. Nobody here will create a PR with that change. You can if you'd like!

bkeryan:
Updated link to calibration info properties: https://www.ni.com/docs/en-US/bundle/ni-daqmx-c-api-ref/page/mxcprop/attributeclasscalibrationinfo.html

PR #460, opened 2023-12-11
enabled support for DAQmxSelfCal

### LINKED ISSUE #22: Support calibrating DAQmx devices
The calibration portion of the DAQmx API is not supported.

See http://zone.ni.com/reference/en-XX/help/370471AA-01/TOC32.htm

Sov-trotter:

This PR enables the support for `DAQmxSelfCal` hence enabling self calibration of the supported DAQ devices. 

Self Calibration is critical functionality for many devices that is encouraged to be run regularly. This also a simple and straight-forward addition that doesn't require tackling the complexity of #22 and #23  

I have manually tested this on Ni-6133 and Ni-4309 DAQ devices.
cc: @neutronstriker

This is my first time contributing to this project. Please let me know if I have missed something. :)


bkeryan:
> Will close https://github.com/ni/nidaqmx-python/issues/22 and https://github.com/ni/nidaqmx-python/issues/23

This is a step in the right direction but we shouldn't close either issue yet.

#22 includes the external calibration session class and external calibration adjustment functions, which is a sizeable effort.

#23 includes the [calibration info properties](https://www.ni.com/docs/en-US/bundle/ni-daqmx-c-api-ref/page/mxcprop/attributeclasscalibrationinfo.html), which is a smaller effort.

zhindes:
> Hi @Sov-trotter, thank you for contributing.
> 
> `source/codegen/metadata/functions.py` is auto-generated by a build process located in an NI internal repo. Someone at NI will need to update that build process to produce the updated SelfCal function definition, or else this change will get reverted the next time we update the API metadata.

I am willing to take on that effort for this contribution!

zhindes:
> > Will close #22 and #23
> 
> This is a step in the right direction but we shouldn't close either issue yet.
> 
> #22 includes the external calibration session class and external calibration adjustment functions, which is a sizeable effort.
> 
> #23 includes the [calibration info properties](https://www.ni.com/docs/en-US/bundle/ni-daqmx-c-api-ref/page/mxcprop/attributeclasscalibrationinfo.html), which is a smaller effort.

I agree. I'll clean up this PR description a bit.

bkeryan:
> > Hi @Sov-trotter, thank you for contributing.
> > `source/codegen/metadata/functions.py` is auto-generated by a build process located in an NI internal repo. Someone at NI will need to update that build process to produce the updated SelfCal function definition, or else this change will get reverted the next time we update the API metadata.
> 
> I am willing to take on that effort for this contribution!

I forgot: we can put customizations in https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/functions_addon.py so that they are not overwritten.

However, updating the build process that generates `functions.py` is better in the long term.

Sov-trotter:
Thanks for the feedback @bkeryan. I moved the SelfCal customization to `functions_addon.py` . This doesn't quite seem to work though.

flef:
Can you please add support for LibraryInterpreter ? selfCal has the same C prototype than Selftest. 

The body of the Selftest function can be almost  copy paste.

Regards

bkeryan:
> Can you please add support for LibraryInterpreter ? selfCal has the same C prototype than Selftest.
> 
> The body of the Selftest function can be almost copy paste.

The interpreter classes are auto-generated and LibraryInterpreter was supported in the first commit, before @Sov-trotter tried my suggestion of using `functions_addon.py` (which didn't work and I need to look into).

Sov-trotter:
> > Hi @Sov-trotter, thank you for contributing.
> > `source/codegen/metadata/functions.py` is auto-generated by a build process located in an NI internal repo. Someone at NI will need to update that build process to produce the updated SelfCal function definition, or else this change will get reverted the next time we update the API metadata.
> 
> I am willing to take on that effort for this contribution!

Hey @zhindes @bkeryan  let me know if guys are able to update the said build process. It would be really great if you can tag a release as soon as possible. :)

zhindes:
> > > Hi @Sov-trotter, thank you for contributing.
> > > `source/codegen/metadata/functions.py` is auto-generated by a build process located in an NI internal repo. Someone at NI will need to update that build process to produce the updated SelfCal function definition, or else this change will get reverted the next time we update the API metadata.
> > 
> > 
> > I am willing to take on that effort for this contribution!
> 
> Hey @zhindes @bkeryan let me know if guys are able to update the said build process.
> 
> Also it would be really great if you can tag a release as soon as possible. :)

This week has been hell, but next week will be much more calm. I'll dig in then, and I also want to make a release soon as well.

bkeryan:
> Thanks for the feedback @bkeryan. I moved the SelfCal customization to `functions_addon.py` . This doesn't quite seem to work though.

I figured out what's going on with `functions_addon.py`. In your first commit, you removed `'python_codegen_method': 'no',`. However, [`src/codegen/metadata/__init__.py`](https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/__init__.py) merges the original function metadata with the addon metadata, so the override in your second commit leaves `'python_codegen_method': 'no',` in place.

Adding `'python_codegen_method': 'yes',` or `'python_codegen_method': 'None',` to the override doesn't entirely work, because some of the codegen assumes that the presence of `python_codegen_method` means that the function is either hand-written or unsupported.

It's possible to work around this by adding `del all_functions["SelfCal"]["python_codegen_method"]` to `src/codegen/metadata/__init__.py`, but it would be better to change  [`src/codegen/utilities/function_helpers.py`](https://github.com/ni/nidaqmx-python/blob/master/src/codegen/utilities/function_helpers.py#L43) to do `if function_data.get("python_codegen_method") or function_name in EXCLUDED_FUNCTIONS:` instead of `if "python_codegen_method" in function_data or function_name in EXCLUDED_FUNCTIONS:` and set `'python_codegen_method': 'None',` in the override.

Since @zhindes is planning to update the metadata next week, this is kind of a moot point, but I wanted to understand why the override didn't work.

bkeryan:
> it would be better to change [`src/codegen/utilities/function_helpers.py`](https://github.com/ni/nidaqmx-python/blob/master/src/codegen/utilities/function_helpers.py#L43) to do `if function_data.get("python_codegen_method") or function_name in EXCLUDED_FUNCTIONS:` instead of `if "python_codegen_method" in function_data or function_name in EXCLUDED_FUNCTIONS:` 

BTW, I made this change in #462

generated/nidaqmx/_base_interpreter.py
```diff
@@ -1345,6 +1345,10 @@ def save_scale(self, scale_name, save_as, author, options):
     def save_task(self, task, save_as, author, options):
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def self_cal(self, device_name):
+        raise NotImplementedError
```
bkeryan, line 1350: Please add a test case to catch errors like this. It can be almost identical to the `self_test_device` test case in #463.

zhindes:
> > Hi @Sov-trotter, thank you for contributing.
> > `source/codegen/metadata/functions.py` is auto-generated by a build process located in an NI internal repo. Someone at NI will need to update that build process to produce the updated SelfCal function definition, or else this change will get reverted the next time we update the API metadata.
> 
> I am willing to take on that effort for this contribution!

@Sov-trotter - You can update your metadata from [users/zhindes/self_cal_metadata](https://github.com/ni/nidaqmx-python/compare/users/zhindes/self_cal_metadata). The SelfCal changes look identical to me, but some other metadata was updated in the process that has no functional impact on the generated Python API (it is used by other tooling throughout our stack).

zhindes (changes requested):
As mentioned in other places: add a test and update to the latest official metadata

generated/nidaqmx/_base_interpreter.py
Sov-trotter: Added as per #463. Thanks for the pointer

src/codegen/metadata/functions_addon.py
```diff
@@ -133,5 +133,29 @@
     },
     'WriteRaw':{
         'python_codegen_method': 'CustomCode_Write',
+    },
```
zhindes, line 136: revert the changes to this file

generated/nidaqmx/_library_interpreter.py
```diff
@@ -4632,6 +4632,9 @@ def save_task(self, task, save_as, author, options):
             task, save_as, author, options)
         self.check_for_error(error_code)
 
+    def self_cal(self, device_name):
```
zhindes, line 4635: this got generated incorrectly. I think because you still have your `functions_addon.py` changes

zhindes (changes requested):
One more issue (I think), but then we'll be good. Since we're here, can you also add this to the [changelog](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) under **Merged Pull Requests**? A link like this: [460: enabled support for DAQmxSelfCal](https://github.com/ni/nidaqmx-python/pull/460).

I can make a release ASAP once this is in

generated/nidaqmx/_library_interpreter.py
Sov-trotter: Oh missed that. My bad.

src/codegen/metadata/functions_addon.py
Sov-trotter: Let me know if it looks correct now

zhindes (approved):

zhindes:
@Sov-trotter - bah, one last issue, I think.

```
./tests/component/system/test_device.py:33:1: BLK100 Black would make changes.
```

You can run `poetry run black tests/component/system/test_device.py`, and it should fix it for you.

zhindes:
> @Sov-trotter - bah, one last issue, I think.
> 
> ```
> ./tests/component/system/test_device.py:33:1: BLK100 Black would make changes.
> ```
> 
> You can run `poetry run black tests/component/system/test_device.py`, and it should fix it for you.

We're good now. I'll babysit it from here.

zhindes:
> @Sov-trotter its in! I should be able to make a release this afternoon, 🤞

[0.9.0](https://pypi.org/project/nidaqmx/0.9.0/) is out

Sov-trotter:
Thanks for this @zhindes @bkeryan

ClaesFredo:
@zhindes @bkeryan 

Just upgraded to 0.9.0 - thx.  :)

I find the ai_dev_scaling_coeff but not much else. 

Dumb questions? 
a) Has the device calibration date been added? (I assume no & yes, self calibration is more important)
b) will the calibration info of #22  be added? It is great info for QA purposes.

zhindes:
> @zhindes @bkeryan
> 
> Just upgraded to 0.9.0 - thx. :)
> 
> I find the ai_dev_scaling_coeff but not much else.
> 
> Dumb questions? a) Has the device calibration date been added? (I assume no & yes, self calibration is more important) b) will the calibration info of #22 be added? It is great info for QA purposes.

a) No, we did not add that. However, nisyscfg ([pypi](https://pypi.org/project/nisyscfg/), [github](https://github.com/tkrebes/nisyscfg-python/)) is high quality and should have support for that.

b) We do not have any current plans to add broader calibration support. Reflection of calibration metadata should already be covered by nisyscfg, and I don't think calibration adjustment is a feature we will prioritize highly, likely ever.

ClaesFredo:
Hi

Thx for clarifying matters and for the nisycfg info. :)

For what it is worth and, perhaps, as food 4 thought - the device calibration_date and its calibration_due_date is useful as Quality information data, ie to assess the device status.

alra0351:
Hi,
thank you for the new release: I am trying to use the self_cal() function on the NI 9205, but i am getting an error... whtas could be wrong here?

`    for device in nidaqmx.system.System.local().devices:
        if device.product_type == 'NI 9205':
            device.self_cal()`

Output: `    raise DaqError(extended_error_info, error_code)
nidaqmx.errors.DaqError: Some or all of the samples requested have not yet been acquired.
To wait for the samples to become available use a longer read timeout or read later in your program. To make the samples available sooner, increase the sample rate. If your task uses a start trigger,  make sure that your start trigger is configured correctly. It is also possible that you configured the task for external timing, and no clock was supplied. If this is the case, supply an external clock.
Property: DAQmx_Read_RelativeTo
Corresponding Value: DAQmx_Val_CurrReadPos
Property: DAQmx_Read_Offset
Corresponding Value: 0

Status Code: -200284
`

zhindes:
@alra0351 That error being returned from `self_cal` is definitely unexpected, and I doubt it is related to the Python API. Can you try hitting the "Self Calibrate" button in MAX? I expect it will return the same error.

alra0351:
@zhindes thank you for your response. "Self Calibration" from MAX works without any problems. Any further ideas what I can try?

zhindes:
@alra0351 Can you create a new discussion and share the full code?

PR #468, opened 2024-01-05
Add Offset Nulling Calibration API support for nidaqmx-python

DeborahOoi96:

We want to include support for Offset Nulling Calibration API in python. 
This PR adds support for the following C functions in python
[DAQmxPerformBridgeOffsetNullingCalEx ](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ccalibration.cpp&version=GBmain&line=57&lineEnd=58&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) 
[DAQmxPerformThrmcplLeadOffsetNullingCal](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ccalibration.cpp&version=GBmain&line=90&lineEnd=90&lineStartColumn=23&lineEndColumn=62&lineStyle=plain&_a=contents)

Also added regression test coverage to ensure that these functions behave as expected.
To make the regression test for DAQmxPerformThrmcplLeadOffsetNullingCal work, I had to add a new simulated device 9214 in the simulated cDAQ chassis as all other devices do not support performing thermocouple lead offset nulling calibration.

It provides support for the Offset Nulling Calibration functions in python

Regression tests added, all pass successfully in my test machine.
image


generated/nidaqmx/task.py
```diff
@@ -402,6 +402,39 @@ def is_task_done(self):
...
+    def perform_bridge_offset_nulling_cal_ex(self, channel="", skip_unsupported_channels=False):
+        """
+        Performs a bridge offset nulling calibration on the channels in the task.
+        If the task measures both bridge-based sensors and non-bridge-based sensors,
+        use the channels input to specify the names of the channels that measure 
+        bridge-based sensors.
+        """
+
+        self._interpreter.perform_bridge_offset_nulling_cal_ex(
+            self._handle, channel, skip_unsupported_channels)
```
zhindes, line 425: We should only support `perform_bridge_offset_nulling_cal_ex`, but not have the `_ex` suffix. That is an artifact of years of DAQmx history and not something we should expose in the Python API. For example, in LabVIEW, we stopped shipping the original version of the API on the palette for similar reasons. Since this is a new API, no need to have the old one at all. FYI - if its easier, keeping the `_ex` in the `_interpreter` layer is fine. I'm not sure if that is the case or not.

zhindes: Oh, I also don't mind if the interpreter ends up with unused methods, like `perform_bridge_offset_nulling_cal` that nobody uses.

src/handwritten/task.py
```diff
@@ -402,6 +402,39 @@ def is_task_done(self):
...
+    def perform_thrmcpl_lead_offset_nulling_cal(self, channel="", skip_unsupported_channels=False):
+        """
+        Performs thermocouple lead offset nulling calibration on the channels in 
+        the task to compensate for offsets introduced by open thermocouple detection.
+        Keep the measured temperature as constant as possible while performing this 
+        adjustment.
+        """
+
+        self._interpreter.perform_thrmcpl_lead_offset_nulling_cal(
+            self._handle, channel, skip_unsupported_channels)
```
zhindes, line 436: This might need to be handwritten because of my comment about not wanting `_ex`, but if it doesn't have to be, that would be nice. We can talk offline about whether or not that is the case.

generated/nidaqmx/task.py
DeborahOoi96: Thanks for your feedback. I have removed the perform_bridge_offset_nulling_cal and only left the `_ex ` function in the interpreter layer.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -2260,6 +2260,22 @@ def load_task(self, session_name):
...
+                task=task, channel=channel,
+                skip_unsupported_channels=skip_unsupported_channels))
+
+    def perform_thrmcpl_lead_offset_nulling_cal(
+            self, task, channel, skip_unsupported_channels):
+        response = self._invoke(
+            self._client.PerformThrmcplLeadOffsetNullingCal,
+            grpc_types.PerformThrmcplLeadOffsetNullingCalRequest(
+                task=task, channel=channel,
+                skip_unsupported_channels=skip_unsupported_channels))
```
zhindes, line 2277: As mentioned in the scrapigen PR (I think?) these methods don't exist. Let's get them added to grpc-device, and then regenerate the grpc code here. We can move that to a separate work item and PR later, if that's easier.

tests/component/test_task.py
```diff
@@ -26,3 +38,23 @@ def test___tasks_with_different_names___hash___not_equal(generate_task):
...
+
+
+@pytest.mark.device_name("cDAQ1Mod2")
+@pytest.mark.parametrize("skip_unsupported_channels", [True, False])
+def test___perform_thrmcpl_lead_offset_nulling_cal___no_errors(
+    ai_thermocouple_task: nidaqmx.Task, skip_unsupported_channels
+) -> None:
+    ai_thermocouple_task.perform_thrmcpl_lead_offset_nulling_cal(
+        ai_thermocouple_task.channels.name, skip_unsupported_channels
+    )
```
zhindes, line 60: I assume these tests pass? I know the strain cal fails on simulated devices. I think these should work, though.

zhindes (approved):
Looks good! I am fine following up with grpc support later for all the stuff y'all are doing.

generated/nidaqmx/_grpc_interpreter.py
bkeryan: > We can move that to a separate work item and PR later, if that's easier.

You can temporarily mark the new tests with `@pytest.mark.grpc_xfail(reason="...")` where ... is a brief explanation of why the test is expected to fail. (XFAIL = eXpect FAIL)

Cc: @charitylxy

tests/component/test_task.py
DeborahOoi96: Yes these test pass as shown in the PR description :)

src/handwritten/task.py
```diff
@@ -402,6 +402,28 @@ def is_task_done(self):
 
         return is_task_done
 
+    def perform_bridge_offset_nulling_cal(self, channel="", skip_unsupported_channels=False):
+        """
+        Performs a bridge offset nulling calibration on the channels in the task.
+        If the task measures both bridge-based sensors and non-bridge-based sensors,
+        use the channels input to specify the names of the channels that measure 
+        bridge-based sensors.
```
bkeryan, line 410: The first line of the docstring is a one-line summary and should not be part of a larger paragraph.

The summary line should be a command ("Perform" not "Performs") as per https://peps.python.org/pep-0257/#one-line-docstrings

bkeryan (changes requested):

src/handwritten/task.py
```diff
@@ -402,6 +402,28 @@ def is_task_done(self):
 
         return is_task_done
 
+    def perform_bridge_offset_nulling_cal(self, channel="", skip_unsupported_channels=False):
```
bkeryan, line 405: Please add the parameter documentation to the docstring in Google docstring format.

```
Args:
    channel: Whatever

    skip_unsupported_channels: Whatever
```

generated/nidaqmx/_grpc_interpreter.py
DeborahOoi96: Thanks for pointing this out, I have marked it with xfail.

tests/component/test_task.py
```diff
@@ -26,3 +38,23 @@ def test___tasks_with_different_names___hash___not_equal(generate_task):
     task2 = generate_task("MyTask2")
 
     assert hash(task1) != hash(task2)
+
+
+@pytest.mark.device_name("bridgeTester")
+@pytest.mark.parametrize("skip_unsupported_channels", [True, False])
+def test___perform_bridge_offset_nulling_cal___no_errors(
```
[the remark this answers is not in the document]
DeborahOoi96: Thanks for pointing this out, I have marked it with xfail.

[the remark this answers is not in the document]
DeborahOoi96: Thanks for pointing this out, I have added default test cases.

src/handwritten/task.py
DeborahOoi96: Thanks for pointing this out, I have separated it into one-line summary and the rest of the explanation is one empty line below.

src/handwritten/task.py
DeborahOoi96: Thanks for pointing this out, I have added the parameter documentation according to Labview HELP documentation.

generated/nidaqmx/_grpc_interpreter.py
DeborahOoi96: I will address the change to grpc-device and regenerate the grpc code in another PR. Thanks

src/handwritten/task.py
```diff
@@ -402,6 +402,52 @@ def is_task_done(self):
...
+        Args:
+            channel: is a subset of virtual channels in the task that you want to calibrate.
+            Use this input if you do not want to calibrate all the channels in the task or
+            if some channels in the task have open thermocouple detection disabled.
+            If the input is empty, this VI attempts to calibrate all virtual channels in the task.
+
+            skip_unsupported_channels: specifies whether or not to skip channels that do not 
+            support calibration.
+            If skip unsupported channels is TRUE, this VI calibrates only supported channels.
+            If FALSE, this VI calibrates the channels specified by channels. The default is FALSE.
```
bkeryan, line 422: If a parameter has more than one line, you need to indent the rest. See https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html for examples.

This should display 2 parameters, but it's interpreting parts of the other lines as parameters.
image

bkeryan (changes requested):

tests/component/test_task.py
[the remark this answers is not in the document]
bkeryan: https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions?anchor=keep-it-simple

> ### Keep it Simple
> Strive to make tests trivially correct -- there are no tests for tests, so they should be as simple as possible.
> 
> ❌ AVOID obscure code
> ❌ AVOID needless complexity
> ❌ AVOID conditional logic
> ...

Please keep it simple and write a separate test function for this.

For given/when/then naming, it's ok to use the "given" field to distinguish between these two test functions. (It's also ok to add "_with_default_arguments" to the "then" field, if the test already has a precondition.)
- `test___default_arguments___perform_bridge_offset_nulling_cal___no_errors`
- `test___arguments_specified___perform_bridge_offset_nulling_cal___no_errors`

src/handwritten/task.py
DeborahOoi96: Thanks for pointing that out, I have indented it and the result is as shown.
image

tests/component/test_task.py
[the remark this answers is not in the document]
DeborahOoi96: Thanks for clarifying, I did not know of the daqmx functional test coding conventions. I have updated it.

bkeryan (approved):

PR #469, opened 2024-01-05
Add Shunt Cal API support for nidaqmx-python

charitylxy:

This PR adds Shunt Calibration API support for the following functions in nidaqmx-python: 
[DAQmxPerformStrainShuntCalEx](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ccalibration.cpp&version=GBmain&line=158&lineEnd=159&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)
[DAQmxPerformBridgeShuntCalEx](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ccalibration.cpp&version=GBmain&line=239&lineEnd=239&lineStartColumn=1&lineEndColumn=8&lineStyle=plain&_a=contents)
It provide support for Shunt Calibration in python. 

Regression tests have been added and all tests passed in local machine.
image


src/handwritten/task.py
```diff
@@ -402,6 +402,150 @@ def is_task_done(self):
...
+                do not support calibration. If skip unsupported channels is TRUE, this
+                VI calibrates only supported channels. If FALSE, this VI calibrates the 
+                channels specified by channels. The default is FALSE.
+        """
+        self._interpreter.perform_bridge_shunt_cal_ex(
+            self._handle, channel, shunt_resistor_value,
+            shunt_resistor_location, shunt_resistor_select,
+            shunt_resistor_source, bridge_resistance,
+            skip_unsupported_channels)     
+
```
zhindes, line 548: This might need to be handwritten because of my comment about not wanting `_ex`, but if it doesn't have to be, that would be nice. We can talk offline about whether or not that is the case.

src/handwritten/task.py
```diff
@@ -402,6 +402,83 @@ def is_task_done(self):
...
+        calibration instructions.
+
+        Performs shunt calibration for the specified channels using a strain
+        gage sensor.
+
+        Args:
+            task/channel in: specifies name of the task or a list of virtual 
+                channels to which the operation applies. If you provide a
+                list of virtual channels, NI-DAQmx creates a task automatically. 
+            channels: specifies a subset of virtual channels in the task that you
```
bkeryan, line 423: Name mismatch: channel vs. channels, shunt_resistor_value vs. shunt resistor value, etc.

Running Sphinx should catch these. Currently the PR workflow runs Sphinx from Tox, as part of `run_system_tests`, which didn't complete due to issues with the self-hosted GitHub runner.

Please run `poetry run tox -e docs` to test this.

tests/component/test_task.py
```diff
@@ -26,3 +40,61 @@ def test___tasks_with_different_names___hash___not_equal(generate_task):
     task2 = generate_task("MyTask2")
 
     assert hash(task1) != hash(task2)
+
+
+@pytest.mark.device_name("bridgeTester")
+@pytest.mark.parametrize(
+    "shunt_resistor_value, shunt_resistor_location, shunt_resistor_select, shunt_resistor_source, bridge_resistance, skip_unsupported_channels",
+    [
+        (100000, 12465, 0, 0, 0.2, True),
```
bkeryan, line 49: This should pass enums, not bare numbers.

src/handwritten/task.py
```diff
@@ -402,6 +402,83 @@ def is_task_done(self):
...
+                the virtual channel.
+            skip unsupported channels: specifies whether or not to skip channels that 
+                do not support calibration. If skip unsupported channels is TRUE, this
+                VI calibrates only supported channels. If FALSE, this VI calibrates the 
+                channels specified by channels. The default is FALSE.
+        """
+        self._interpreter.perform_bridge_shunt_cal_ex(
+            self._handle, channel, shunt_resistor_value,
+            shunt_resistor_location, shunt_resistor_select,
+            shunt_resistor_source, bridge_resistance,
```
bkeryan, line 479: This should expect enums.

```suggestion
            shunt_resistor_location.value, shunt_resistor_select.value,
            shunt_resistor_source.value, bridge_resistance,
```

bkeryan (changes requested):

src/handwritten/task.py
bkeryan: BTW, the docstring can specify the enum type in parens. Long-term, I would prefer for this to use type hints, but the current codebase uses parens in the docstring.

src/codegen/metadata/enums.py
```diff
@@ -17134,6 +17134,72 @@
             }
         ]
     },
+    'ShuntCalSelect0': {
```
bkeryan, line 17137: Is this even used?

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -2260,6 +2260,36 @@ def load_task(self, session_name):
...
         return response.task, response.new_session_initialized
 
+    def perform_bridge_shunt_cal_ex(
+            self, task, channel, shunt_resistor_value,
+            shunt_resistor_location, shunt_resistor_select,
+            shunt_resistor_source, bridge_resistance,
+            skip_unsupported_channels):
+        response = self._invoke(
+            self._client.PerformBridgeShuntCalEx,
+            grpc_types.PerformBridgeShuntCalExRequest(
```
bkeryan, line 2270: These RPC and requests do not exist in nidaqmx.proto.

generated/nidaqmx/constants.py
```diff
@@ -627,6 +627,20 @@ class ShuntCalSelect(Enum):
     AAND_B = 12515  #: Switches A and B.
```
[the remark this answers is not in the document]
zhindes: Since we're bumping to 1.0 as part of this effort, I think fixes like this are OK. For methods we can use `deprecation.deprecated`, but I don't see anything similar for enums. At the very least we should mention the change in the CHANGELOG.md

generated/nidaqmx/constants.py
```diff
@@ -627,6 +627,20 @@ class ShuntCalSelect(Enum):
...
 
 
+class ShuntCalSource(Enum):
+    DEFAULT = -1  #: Default
+    BUILT_IN = 10200  #: Built-In
+    USER_PROVIDED = 10167  #: User Provided
+
+
+class ShuntElementLocation(Enum):
+    R_1 = 12465
```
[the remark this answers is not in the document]
zhindes: Ugh, no underscore. Like above, since we're going to 1.0, lets fix things like this. However, we should mention in `CHANGELOG.md`

generated/nidaqmx/_grpc_interpreter.py
zhindes: This will cause a runtime error. Do you think that is good enough, or should this explicitly throw some sort of `NotSupported` error?

zhindes (commented):
Brad covered everything quite well. I added some replies to his comments.

generated/nidaqmx/_grpc_interpreter.py
bkeryan: We shouldn't ship code that uses nonexistent methods and types. This would fail with static analysis tools like mypy.

We should do one of the following:
- (preferred) Add these methods to grpc-device and nidaqmx.proto. This will return a gRPC `UNIMPLEMENTED` status with older versions of the NI gRPC Device Server and work with newer versions of the server.
- Update the codegen metadata to indicate that these methods are not supported for gRPC.

tests/component/test_task.py
```diff
@@ -26,3 +40,61 @@ def test___tasks_with_different_names___hash___not_equal(generate_task):
...
+        (100000, 12465, 0, 0, 0.2, False),
+        (100000, 12467, 0, 0, 0.2, True),
+        (100000, 12467, 0, 0, 0.2, False),
+    ],
+)
+def test___perform_bridge_shunt_cal___no_errors(
```
bkeryan, line 55: Are these strain cal functions guaranteed to succeed on simulated devices? This test had one failure that wasn't related to gRPC.

https://github.com/ni/nidaqmx-python/actions/runs/7444178717/job/20285452399?pr=469#step:4:272

```
tests\component\test_task.py:64: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
generated\nidaqmx\task.py:476: in perform_bridge_shunt_cal
    self._interpreter.perform_bridge_shunt_cal_ex(
generated\nidaqmx\_library_interpreter.py:3898: in perform_bridge_shunt_cal_ex
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <nidaqmx._library_interpreter.LibraryInterpreter object at 0x03DF3F08>
error_code = -201493, samps_per_chan_written = None, samps_per_chan_read = None

    def check_for_error(self, error_code, samps_per_chan_written=None, samps_per_chan_read=None):
        if not error_code:
            return
    
        if error_code < 0:
            extended_error_info = self.get_extended_error_info()
    
            if samps_per_chan_read is not None:
                raise DaqReadError(extended_error_info, error_code, samps_per_chan_read)
            elif samps_per_chan_written is not None:
                raise DaqWriteError(extended_error_info, error_code, samps_per_chan_written)
            else:
>               raise DaqError(extended_error_info, error_code)
E               nidaqmx.errors.DaqError: Shunt calibration failed. The calculated gain adjust is out of range. Ensure that the shunt calibration terminals are connected properly and that the shunt resistance and shunt element location settings match how the hardware is wired.
E               Property: DAQmx_AI_Bridge_ShuntCal_ShuntCalAResistance
E               Corresponding Value:  100.0e3
E               Simulated Value with Shunt: -1.249999e-6
E               Measured Value with Shunt: -2.717376e-6
E               Property: DAQmx_AI_Bridge_ShuntCal_GainAdjust
E               Requested Value:  460.002117e-3
E               Maximum Value:  1.500000
E               Minimum Value:  500.0e-3
E               Channel Name: bridgeTester/ai0
E               
E               Task Name: _unnamedTask<18>
E               
E               Status Code: -201493

generated\nidaqmx\_library_interpreter.py:6080: DaqError
```

zhindes: No, ShuntCal will fail with the error you pasted if you don't actually have the hardware configuration wired up correctly. I guess it could randomly pass if the simulated data works out. Its always a sine wave (in ADC codes I think?), which goes haywire with Strain, I believe.  I guess maybe you could randomly get an average close to expected and it would succeed, but I highly doubt it. I think making this test expect that error is probably fine.

src/handwritten/task.py
bkeryan: Hmm, it didn't warn: https://github.com/ni/nidaqmx-python/actions/runs/7444178717/job/20285452399?pr=469#step:4:3942

I wonder why. Perhaps sphinx-autodoc doesn't validate parameters, or perhaps the slashes and spaces are causing sphinx-napoleon to ignore these parameters.

Anyway, in addition to fixing the docstrings, please build the docs and look at them in a web browser to make sure they are formatted right.

generated/nidaqmx/constants.py
```diff
@@ -624,7 +624,7 @@ class SensorPowerType(Enum):
 class ShuntCalSelect(Enum):
     A = 12513  #: Switch A.
     B = 12514  #: Switch B.
-    AAND_B = 12515  #: Switches A and B.
```
zhindes, line 627: Can you update CHANGELOG.md to mention this incompatibility. Don't worry too much about wordsmithing, we can take a final pass before release

generated/nidaqmx/constants.py
```diff
@@ -634,10 +634,10 @@ class ShuntCalSource(Enum):
 
 
 class ShuntElementLocation(Enum):
-    R_1 = 12465
-    R_2 = 12466
-    R_3 = 12467
-    R_4 = 14813
```
zhindes: Can you update CHANGELOG.md to mention this incompatibility. Don't worry too much about wordsmithing, we can take a final pass before release

zhindes (approved):
Nice!

tests/component/test_task.py
bkeryan: I think most applications perform offset nulling before shunt cal. That may also make a difference. But yeah, this test ought to tolerate this error.

src/handwritten/task.py
```diff
@@ -405,34 +405,31 @@ def is_task_done(self):
 
     def perform_strain_shunt_cal(
             self, channel="", shunt_resistor_value=10000,
-            shunt_resistor_location=0, shunt_resistor_select=0,
-            shunt_resistor_source=0, skip_unsupported_channels=False):
+            shunt_resistor_location=ShuntElementLocation.R1, shunt_resistor_select=ShuntCalSelect.A,
+            shunt_resistor_source=ShuntCalSource.BUILT_IN, skip_unsupported_channels=False):
```
bkeryan, line 409: Please match the LabVIEW VI's default values: **100000**, **R3**, A, **DEFAULT**
image

src/handwritten/task.py
```diff
@@ -441,37 +438,34 @@ def perform_strain_shunt_cal(
 
     def perform_bridge_shunt_cal(
             self, channel="", shunt_resistor_value=100000,
-            shunt_resistor_location=0, shunt_resistor_select=0,
-            shunt_resistor_source=0, bridge_resistance=120,
+            shunt_resistor_location=ShuntElementLocation.R1, shunt_resistor_select=ShuntCalSelect.A,
+            shunt_resistor_source=ShuntCalSource.BUILT_IN, bridge_resistance=120,
             skip_unsupported_channels=False):
```
bkeryan, line 443: Please match the LabVIEW VI's default values: 100000, **R3**, A, DEFAULT, 120
image

src/handwritten/task.py
```diff
@@ -405,34 +405,31 @@ def is_task_done(self):
...
-            shunt_resistor_location=0, shunt_resistor_select=0,
-            shunt_resistor_source=0, skip_unsupported_channels=False):
+            shunt_resistor_location=ShuntElementLocation.R1, shunt_resistor_select=ShuntCalSelect.A,
+            shunt_resistor_source=ShuntCalSource.BUILT_IN, skip_unsupported_channels=False):
         """
         Performs shunt calibration for the specified channels of the task. 
-        The instances of this polymorphic VI correspond to the type of bridge 
+        The instances of this method correspond to the type of bridge 
         sensor. Refer to the calibration procedure for your module for detailed
         calibration instructions.
```
bkeryan, line 414: On further thought, it looks like you pulled the 1st and 2nd paragraphs from the poly VI help and the instance VI help, respectively. For Python, it would make sense to combine them.

Also: the 1st line of a Python docstring should be a one-line summary. It should not be part of a larger paragraph.

How about this? (If the 1st line fits in 100 chars)

> Performs shunt calibration for the specified channels using a strain gage sensor.
> 
> Refer to the calibration procedure for your module for detailed calibration instructions.

Ditto for the other method.

tests/component/test_task.py
bkeryan: This still needs done.

src/handwritten/task.py
bkeryan: This still needs done.

tests/component/test_task.py
bkeryan, line 49: Please add another test case for each function that uses the default arguments. This would also catch the missing `.value` for the enum parameters.

tests/component/test_task.py
bkeryan: If the test always expected an error, the answer would be to use `with pytest.raises(...):`

Since we want to tolerate some errors, we probably need to do
```
try:
   task.whatever()
except DaqError as e:
    if e.code != whatever:
        raise
```

tests/component/test_task.py
```diff
@@ -26,3 +40,61 @@ def test___tasks_with_different_names___hash___not_equal(generate_task):
     task2 = generate_task("MyTask2")
 
     assert hash(task1) != hash(task2)
+
+
+@pytest.mark.device_name("bridgeTester")
+@pytest.mark.parametrize(
+    "shunt_resistor_value, shunt_resistor_location, shunt_resistor_select, shunt_resistor_source, bridge_resistance, skip_unsupported_channels",
```
bkeryan, line 47: Add `@pytest.mark.grpc_xfail(reason="Requires NI gRPC Device Server version 2.5 or later")`

bkeryan (changes requested):

src/handwritten/task.py
bkeryan: The summary line should be a command ("Perform" not "Performs") as per https://peps.python.org/pep-0257/#one-line-docstrings

tests/component/test_task.py
```diff
@@ -26,3 +42,103 @@ def test___tasks_with_different_names___hash___not_equal(generate_task):
...
+@pytest.mark.grpc_xfail(
+    reason="Requires NI gRPC Device Server version 2.5 or later", raises=RpcError
+)
+def test___perform_bridge_shunt_cal_default___no_errors(
+    ai_bridge_task: nidaqmx.Task,
+) -> None:
+    try:
+        ai_bridge_task.perform_bridge_shunt_cal(ai_bridge_task.channels.name)
+    except nidaqmx.DaqError as e:
+        if e.error_code != -201493:
```
bkeryan, line 94: Add this import: `from nidaqmx.error_codes import DAQmxErrors`
And replace the hardcoded number with `DAQmxErrors.SHUNT_CAL_FAILED_OUT_OF_RANGE`

CHANGELOG.md
```diff
@@ -20,11 +20,10 @@
 All notable changes to this project will be documented in this file.
 
 ## 0.9.1
-
 * ### Merged Pull Requests
     *  ...
 * ### Major Changes
-    * ...
+    * Fix naming issues. `AAND_B` is now `A_AND_B`,  `R_1` is now `R1`,  `R_2` is now `R2`,  `R_3` is now `R3` and  `R_4` is now `R4`.
```
bkeryan, line 26:
```suggestion
    * Add support for strain calibration (offset nulling and shunt calibration)
    * Fix naming issues:
      * Rename `ShuntCalSelect.AAND_B` to `A_AND_B`.
```

No need to mention R_1 etc. because they didn't exist before this release.

FYI @DeborahOoi96 one bullet can cover both strain cal features

bkeryan (approved):
Approved with suggestions

PR #471, opened 2024-01-19
Add Time Trigger API Support for nidaqmx-python

charitylxy:

[DAQmxCfgTimeStartTrig ](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ctrigger.cpp&version=GBmain&line=209&lineEnd=210&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

Remaining API yet to be added (due to functions missing in daqmxAPISharp.json): 
[DAQmxCfgAnlgMultiEdgeRefTrig](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ctrigger.cpp&version=GBmain&line=389&lineEnd=436&lineStartColumn=1&lineEndColumn=2&lineStyle=plain&_a=contents)
[DAQmxCfgAnlgMultiEdgeStartTrig](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ctrigger.cpp&version=GBmain&line=121&lineEnd=166&lineStartColumn=1&lineEndColumn=2&lineStyle=plain&_a=contents)

Work Item: [[DAQmx] Add Time Triggers API support for nidaqmx-python](https://ni.visualstudio.com/DevCentral/_workitems/edit/2610992)

It provide support for Time Trigger in python.

- Ran nidaqmx-python code generator and validated the API was generated correctly
- TODO: Regression tests for Time Trigger API to be added

### DIFF
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -378,6 +378,11 @@ class GrpcStubInterpreter(BaseInterpreter):
                 active_edge_raw=active_edge, sample_mode_raw=sample_mode,
                 samps_per_chan=samps_per_chan))
 
+    def cfg_time_start_trig(self, task, when, timescale):
+        response = self._invoke(
+            self._client.CfgTimeStartTrig,
+            grpc_types.CfgTimeStartTrigRequest(task=task, when=when, timescale_raw=timescale))
+
     def cfg_watchdog_ao_expir_states(
             self, task, channel_names, expir_state_array, output_type_array):
         response = self._invoke(
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -390,6 +390,18 @@ class LibraryInterpreter(BaseInterpreter):
             task, source, rate, active_edge, sample_mode, samps_per_chan)
         self.check_for_error(error_code)
 
+    def cfg_time_start_trig(self, task, when, timescale):
+        cfunc = lib_importer.windll.DAQmxCfgTimeStartTrig
+        if cfunc.argtypes is None:
+            with cfunc.arglock:
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        lib_importer.task_handle, DateTime, ctypes.c_int]
+
+        error_code = cfunc(
+            task, when, timescale)
+        self.check_for_error(error_code)
+
     def cfg_watchdog_ao_expir_states(
             self, task, channel_names, expir_state_array, output_type_array):
         cfunc = lib_importer.windll.DAQmxCfgWatchdogAOExpirStates
--- a/generated/nidaqmx/_task_modules/triggering/start_trigger.py
+++ b/generated/nidaqmx/_task_modules/triggering/start_trigger.py
@@ -1037,6 +1037,20 @@ class StartTrigger:
         self._interpreter.cfg_dig_pattern_start_trig(
             self._handle, trigger_source, trigger_pattern, trigger_when.value)
 
+    def cfg_time_start_trig(self, when, timescale=Timescale.USE_HOST):
+        """
+        New Start Trigger
+
+        Args:
+            when (nidaqmx.constants.DateTime): Specifies when to
+                trigger.
+            timescale (Optional[nidaqmx.constants.Timescale]): Specifies
+                the start trigger timestamp time scale.
+        """
+
+        self._interpreter.cfg_time_start_trig(
+            self._handle, when, timescale.value)
+
     def disable_start_trig(self):
         """
         Configures the task to start acquiring or generating samples
--- a/generated/nidaqmx/_base_interpreter.py
+++ b/generated/nidaqmx/_base_interpreter.py
@@ -148,6 +148,10 @@ class BaseInterpreter(abc.ABC):
         raise NotImplementedError
 
     @abc.abstractmethod
+    def cfg_time_start_trig(self, task, when, timescale):
+        raise NotImplementedError
+
+    @abc.abstractmethod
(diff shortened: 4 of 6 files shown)


generated/nidaqmx/_task_modules/triggering/start_trigger.py
```diff
@@ -1037,6 +1037,20 @@ def cfg_dig_pattern_start_trig(
         self._interpreter.cfg_dig_pattern_start_trig(
             self._handle, trigger_source, trigger_pattern, trigger_when.value)
 
+    def cfg_time_start_trig(self, when, timescale=Timescale.USE_HOST):
+        """
+        New Start Trigger
+
+        Args:
+            when (nidaqmx.constants.DateTime): Specifies when to
```
zhindes, line 1045: I don't think this is the right type. I added [AbsoluteTime in _lib_time.py](https://github.com/ni/nidaqmx-python/blob/c5fddf5c9dd6be9b0d02bcd96338963176515cf3/src/handwritten/_lib_time.py) that we should accept. We should also accept a `datetime` and a `hightime` from the std library.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -390,6 +390,18 @@ def cfg_samp_clk_timing(
             task, source, rate, active_edge, sample_mode, samps_per_chan)
         self.check_for_error(error_code)
 
+    def cfg_time_start_trig(self, task, when, timescale):
+        cfunc = lib_importer.windll.DAQmxCfgTimeStartTrig
+        if cfunc.argtypes is None:
+            with cfunc.arglock:
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        lib_importer.task_handle, DateTime, ctypes.c_int]
```
zhindes, line 399: When should be an `AbsoluteTime`. `AbsoluteTime` is a ctypes type that should work correctly here, I think.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -378,6 +378,11 @@ def cfg_samp_clk_timing(
                 active_edge_raw=active_edge, sample_mode_raw=sample_mode,
                 samps_per_chan=samps_per_chan))
 
+    def cfg_time_start_trig(self, task, when, timescale):
+        response = self._invoke(
+            self._client.CfgTimeStartTrig,
+            grpc_types.CfgTimeStartTrigRequest(task=task, when=when, timescale_raw=timescale))
```
zhindes, line 384: what type does the grpc lib use for when?

zhindes (changes requested):
I think the types aren't correct here. Some tests would have found a lot of issues, I think?

generated/nidaqmx/_grpc_interpreter.py
bkeryan: google.protobuf.Timestamp
https://github.com/ni/nidaqmx-python/blob/master/src/codegen/protos/nidaqmx.proto#L4690
https://github.com/ni/nidaqmx-python/blob/master/src/handwritten/_grpc_time.py

generated/nidaqmx/_task_modules/triggering/start_trigger.py
bkeryan: The interpreter interface should only use datetime/hightime, not the ctypes and gRPC specific types.

The interpreter implementations should be responsible for converting datetime/hightime to/from AbsoluteTime/google.protobuf.Timestamp.

zhindes: Yup, you're right. I'm just remembering everything we talked about when I added `AbsoluteTime`. `AbsoluteTime` is the type that the lib_interpreter uses to interface with the CAPI. `google.protobuf.Timestamp` for the grpc_interpreter. In addition, the top-level API should accept `datetime` or `hightime` as well. So just pass them down to the interpreter and each will convert to their respective types.

generated/nidaqmx/_task_modules/triggering/start_trigger.py
bkeryan, line 1: In addition to this config function, several time trigger attributes are missing from [start_trigger.py](https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/triggering/start_trigger.py) and similar classes. Are these on your radar?

https://www.ni.com/docs/en-US/bundle/ni-daqmx-properties/page/daqmxprop/daqmxtrigger.html
- Start:Time:When
- Start:Timestamp:Value
- Reference:Timestamp:Value
- More:Arm Start:Time:When
- More:Arm Start:Timestamp:Value

generated/nidaqmx/_task_modules/triggering/start_trigger.py
bkeryan: Some starting points:
- Library
  - https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/library_interpreter/default_c_function_call.py.mako
  - https://github.com/ni/nidaqmx-python/blob/master/src/codegen/utilities/interpreter_helpers.py#L135
- gRPC
  - https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/grpc_interpreter/default_grpc_function_call.py.mako
  - https://github.com/ni/nidaqmx-python/blob/master/src/codegen/utilities/interpreter_helpers.py#L329

generated/nidaqmx/_task_modules/triggering/start_trigger.py
charitylxy: Opps, I might've missed the attributes out, will look into adding those.

PR #473, opened 2024-01-19
Add support for DAQmx calibration info properties

WayneDroid:

This PR is to add calibration info APIs support.
Work Item: https://dev.azure.com/ni/DevCentral/_workitems/edit/2611013

It provides support for calibration info reflection. 

Regression test on test machine -  passing.

```
tests/component/system/test_device.py::test___devices_with_same_name___compare___equal[library_init_kwargs] PASSED                                   [ 33%]
tests/component/system/test_device.py::test___devices_with_same_name___compare___equal[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not inst...) [ 33%]
tests/component/system/test_device.py::test___devices_with_different_names___compare___not_equal[library_init_kwargs] PASSED                         [ 33%]
tests/component/system/test_device.py::test___devices_with_different_names___compare___not_equal[grpc_init_kwargs] SKIPPED (NI gRPC Device Serve...) [ 33%]
tests/component/system/test_device.py::test___devices_with_same_name___hash___equal[library_init_kwargs] PASSED                                      [ 33%]
tests/component/system/test_device.py::test___devices_with_same_name___hash___equal[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not installed)  [ 33%]
tests/component/system/test_device.py::test___devices_with_different_names___hash___not_equal[library_init_kwargs] PASSED                            [ 33%]
tests/component/system/test_device.py::test___devices_with_different_names___hash___not_equal[grpc_init_kwargs] SKIPPED (NI gRPC Device Server n...) [ 33%]
tests/component/system/test_device.py::test___self_test_device___no_errors[library_init_kwargs] PASSED                                               [ 33%]
tests/component/system/test_device.py::test___self_test_device___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not installed)           [ 33%]
tests/component/system/test_device.py::test___restore_last_ext_cal_const___no_errors[library_init_kwargs] PASSED                                     [ 33%]
tests/component/system/test_device.py::test___restore_last_ext_cal_const___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not installed) [ 34%]
tests/component/system/test_device.py::test___self_cal___no_errors[library_init_kwargs] PASSED                                                       [ 34%]
tests/component/system/test_device.py::test___self_cal___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not installed)                   [ 34%]
tests/component/system/test_device_properties.py::test___constructed_device___get_property___returns_value[library_init_kwargs] PASSED               [ 34%]
tests/component/system/test_device_properties.py::test___constructed_device___get_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC De...) [ 34%]
tests/component/system/test_device_properties.py::test___nonexistent_device___get_property___raises_invalid_device_id[library_init_kwargs] PASSED    [ 34%]
tests/component/system/test_device_properties.py::test___nonexistent_device___get_property___raises_invalid_device_id[grpc_init_kwargs] SKIPPED      [ 34%]
tests/component/system/test_device_properties.py::test___device___bool_property___returns_value[library_init_kwargs] PASSED                          [ 34%]
tests/component/system/test_device_properties.py::test___device___bool_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC Device Server...) [ 34%]
tests/component/system/test_device_properties.py::test___device___float_property___returns_value[library_init_kwargs] PASSED                         [ 34%]
tests/component/system/test_device_properties.py::test___device___float_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC Device Serve...) [ 34%]
tests/component/system/test_device_properties.py::test___device___uint_property___returns_value[library_init_kwargs] PASSED                          [ 34%]
tests/component/system/test_device_properties.py::test___device___uint_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC Device Server...) [ 35%]
tests/component/system/test_device_properties.py::test___device___string_property___returns_value[library_init_kwargs] PASSED                        [ 35%]
tests/component/system/test_device_properties.py::test___device___string_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC Device Serv...) [ 35%]
tests/component/system/test_device_properties.py::test___device___enum_property___returns_value[library_init_kwargs] PASSED                          [ 35%]
tests/component/system/test_device_properties.py::test___device___enum_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC Device Server...) [ 35%]
tests/component/system/test_device_properties.py::test___device___list_of_float_property___returns_value[library_init_kwargs] PASSED                 [ 35%]
tests/component/system/test_device_properties.py::test___device___list_of_float_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC Devi...) [ 35%]
tests/component/system/test_device_properties.py::test___device___list_of_enum_property___returns_value[library_init_kwargs] PASSED                  [ 35%]
tests/component/system/test_device_properties.py::test___device___list_of_enum_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC Devic...) [ 35%]
tests/component/system/test_device_properties.py::test___device___list_of_uint_property___returns_value[library_init_kwargs] PASSED                  [ 35%]
tests/component/system/test_device_properties.py::test___device___list_of_uint_property___returns_value[grpc_init_kwargs] SKIPPED (NI gRPC Devic...) [ 35%]
tests/component/system/test_device_properties.py::test___device___get_deprecated_properties___reports_warnings[library_init_kwargs] PASSED           [ 35%]
tests/component/system/test_device_properties.py::test___device___get_deprecated_properties___reports_warnings[grpc_init_kwargs] SKIPPED (NI gRP...) [ 35%]
tests/component/system/test_device_properties.py::test___chassis___get_modules___returns_modules[library_init_kwargs] PASSED                         [ 36%]
tests/component/system/test_device_properties.py::test___chassis___get_modules___returns_modules[grpc_init_kwargs] SKIPPED (NI gRPC Device Serve...) [ 36%]
tests/component/system/test_device_properties.py::test___chassis___get_modules___shared_interpreter[library_init_kwargs] PASSED                      [ 36%]
tests/component/system/test_device_properties.py::test___chassis___get_modules___shared_interpreter[grpc_init_kwargs] SKIPPED (NI gRPC Device Se...) [ 36%]
tests/component/system/test_device_properties.py::test___module___get_chassis___returns_chassis[library_init_kwargs] PASSED                          [ 36%]
tests/component/system/test_device_properties.py::test___module___get_chassis___returns_chassis[grpc_init_kwargs] SKIPPED (NI gRPC Device Server...) [ 36%]
tests/component/system/test_device_properties.py::test___module___get_chassis___shared_interpreter[library_init_kwargs] PASSED                       [ 36%]
tests/component/system/test_device_properties.py::test___module___get_chassis___shared_interpreter[grpc_init_kwargs] SKIPPED (NI gRPC Device Ser...) [ 36%]
tests/component/system/test_device_properties.py::test___ext_cal_last_date_and_time___no_errors[library_init_kwargs] PASSED                          [ 36%]
tests/component/system/test_device_properties.py::test___ext_cal_last_date_and_time___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server...) [ 36%]
tests/component/system/test_device_properties.py::test___self_cal_last_date_and_time___no_errors[library_init_kwargs] PASSED                         [ 36%]
tests/component/system/test_device_properties.py::test___self_cal_last_date_and_time___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Serve...) [ 36%]
tests/component/system/test_device_properties.py::test___device_supports_cal___no_errors[library_init_kwargs] PASSED                                 [ 36%]
tests/component/system/test_device_properties.py::test___device_supports_cal___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not in...) [ 37%]
tests/component/system/test_device_properties.py::test___cal_acc_connection_count__raises_attr_not_supported[library_init_kwargs] PASSED             [ 37%]
tests/component/system/test_device_properties.py::test___cal_acc_connection_count__raises_attr_not_supported[grpc_init_kwargs] SKIPPED (NI gRPC ...) [ 37%]
tests/component/system/test_device_properties.py::test___cal_recommended_acc_connection_count_limit___raises_attr_not_supported[library_init_kwargs] PASSED [ 37%]
tests/component/system/test_device_properties.py::test___cal_recommended_acc_connection_count_limit___raises_attr_not_supported[grpc_init_kwargs] SKIPPED [ 37%]
tests/component/system/test_device_properties.py::test___cal_user_defined_info___no_errors[library_init_kwargs] PASSED                               [ 37%]
tests/component/system/test_device_properties.py::test___cal_user_defined_info___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not ...) [ 37%]
tests/component/system/test_device_properties.py::test___cal_dev_temp___no_errors[library_init_kwargs] PASSED                                        [ 37%]
tests/component/system/test_device_properties.py::test___cal_dev_temp___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not installed)    [ 37%]
tests/component/system/test_device_properties.py::test___ext_cal_last_temp___no_errors[library_init_kwargs] PASSED                                   [ 37%]
tests/component/system/test_device_properties.py::test___ext_cal_last_temp___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not inst...) [ 37%]
tests/component/system/test_device_properties.py::test___ext_cal_recommended_interval___no_errors[library_init_kwargs] PASSED                        [ 37%]
tests/component/system/test_device_properties.py::test___ext_cal_recommended_interval___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Serv...) [ 38%]
tests/component/system/test_device_properties.py::test___self_cal_last_temp___no_errors[library_init_kwargs] PASSED                                  [ 38%]
tests/component/system/test_device_properties.py::test___self_cal_last_temp___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not ins...) [ 38%]
tests/component/system/test_device_properties.py::test___self_cal_supported___no_errors[library_init_kwargs] PASSED                                  [ 38%]
tests/component/system/test_device_properties.py::test___self_cal_supported___no_errors[grpc_init_kwargs] SKIPPED (NI gRPC Device Server not ins...) [ 38%]
```


generated/nidaqmx/system/device.py
```diff
@@ -1069,6 +1071,17 @@ def reset_device(self):
         self._interpreter.reset_device(
             self._name)
 
+    def restore_last_ext_cal_const(self, device_name):
```
zhindes, line 1074: We don't need `device_name` here, we have `self._name`

generated/nidaqmx/system/device.py
```diff
@@ -1069,6 +1071,17 @@ def reset_device(self):
...
+        """
+
```
zhindes, line 1076: missing the docstring

generated/nidaqmx/system/device.py
```diff
@@ -1069,6 +1071,17 @@ def reset_device(self):
...
+    def restore_last_ext_cal_const(self, device_name):
+        """
+        
+
+        Args:
+            device_name (str): 
+        """
+
+        self._interpreter.restore_last_ext_cal_const(
+            self._name, device_name)
```
zhindes, line 1083: extra param here

generated/nidaqmx/system/device.py
```diff
@@ -1161,6 +1174,27 @@ def unreserve_network_device(self):
...
+    def get_self_cal_last_date_and_time(device_name):
+        last_date_and_time = self._interpreter.get_self_cal_last_date_and_time(device_name)
+
+        return datetime(
+            year = last_date_and_time[0],
+            month = last_date_and_time[1],
+            day = last_date_and_time[2],
+            hour = last_date_and_time[3],
+            minute = last_date_and_time[4]
+        )
```
zhindes, line 1197: don't need `device_name` param. missing docstrings

zhindes (changes requested):
Good start!

generated/nidaqmx/_stubs/nidaqmx_pb2.py
bkeryan, line 1: The PR workflows didn't run because this file (and possibly others) has merge conflicts.

Since this is code-generated, do not waste time merging it by hand. I typically resolve codegen merge conflicts by reverting the codegen files, merging main into my branch, and re-running the codegen scripts.

generated/nidaqmx/system/device.py
```diff
@@ -1161,6 +1174,27 @@ def unreserve_network_device(self):
 
     # endregion
 
+    def get_ext_cal_last_date_and_time(device_name):
+        last_date_and_time = self._interpreter.get_ext_cal_last_date_and_time(device_name)
+
+        return datetime(
+            year = last_date_and_time[0],
```
bkeryan, line 1181: https://pep8.org/#other-recommendations

> Don’t use spaces around the = sign when used to indicate a keyword argument or a default parameter value.

I'm a bit surprised that ni-python-styleguide didn't warn about this, but maybe it doesn't enforce this rule.

generated/nidaqmx/system/device.py
```diff
@@ -1161,6 +1174,27 @@ def unreserve_network_device(self):
 
     # endregion
 
+    def get_ext_cal_last_date_and_time(device_name):
```
bkeryan, line 1177: In my opinion, these should be properties, not get methods. 

They are properties in LabVIEW:
https://www.ni.com/docs/en-US/bundle/ni-daqmx-properties/page/daqmxprop/daqmxcalibrationinfo.html

and .NET:
```
class Device
...
        property DateTime SelfCalibrationDateTime { DateTime get(); }
        property DateTime ExternalCalibrationDate { DateTime get(); }
```

```suggestion
    @property
    def ext_cal_last_date_and_time(device_name):
```

bkeryan, line 1177: Instance methods need a `self` parameter.
```suggestion
    def get_ext_cal_last_date_and_time(self):
```

src/codegen/templates/system/device.py.mako
bkeryan, line 1: You should figure out how to get this to generate the cal info properties:

```
<%namespace name="property_template" file="/property_template.py.mako"/>\
%for attribute in attributes:
${property_template.script_property(attribute)}\
%endfor
\
<%namespace name="deprecated_template" file="/property_deprecated_template.py.mako"/>\
${deprecated_template.script_deprecated_property(attributes)}\
<%namespace name="function_template" file="/function_template.py.mako"/>\
%for function_object in functions:
${function_template.script_function(function_object)}
%endfor
\
```

bkeryan: This is getting only the Device attributes:
```
attributes = get_attributes(data, "Device")
```
Try appending `get_attributes(data, "CalibrationInfo")`

generated/nidaqmx/system/device.py
WayneDroid: Thanks for pointing it out, `device_name` is removed.

generated/nidaqmx/system/device.py
WayneDroid: docstring is added.

generated/nidaqmx/system/device.py
WayneDroid: `device_name` is removed.

generated/nidaqmx/system/device.py
WayneDroid: Removed `device_name`, docstrings are added.

generated/nidaqmx/system/device.py
WayneDroid: The get methods are converted to properties.

WayneDroid: Missed out the `self` parameter, it's added.

src/codegen/templates/system/device.py.mako
```diff
@@ -222,6 +224,27 @@ ${function_template.script_function(function_object)}
...
+    def get_self_cal_last_date_and_time(device_name):
+        last_date_and_time = self._interpreter.get_self_cal_last_date_and_time(device_name)
+
+        return datetime(
+            year = last_date_and_time[0],
+            month = last_date_and_time[1],
+            day = last_date_and_time[2],
+            hour = last_date_and_time[3],
+            minute = last_date_and_time[4]
+        )
```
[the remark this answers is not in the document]
WayneDroid: `device_supports_cal` is added as property.
One thing I can't find is the docstring for `device_supports_cal`, unlike other APIs, eg: `DAQmxGetSelfCalLastDateAndTime` in https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/help&_a=contents&version=GBmain. Is it internal method?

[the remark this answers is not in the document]
bkeryan: No, it's in NIDAQmx.h so that means it's public. I guess it's not documented in the DAQmx CVI help, though.

src/codegen/templates/system/device.py.mako
bkeryan: If you want to split this part out into a separate PR, that's fine with me. The hand-written stuff looks good.

generated/nidaqmx/system/device.py
```diff
@@ -1071,16 +1117,15 @@ def reset_device(self):
...
-        Args:
-            device_name (str): 
+        This function nullifies any self-calibration you perform on the
+        device. If you have never performed a self-calibration on the
+        device, this function has no effect.
         """
 
         self._interpreter.restore_last_ext_cal_const(
-            self._name, device_name)
+            self._name)
```
zhindes, line 1128: This looks formatted wrong (I don't think this should be multiple lines), but I think the PR will yell at you

zhindes (approved):
What you have so far looks good to me. I agree with Brad that the other properties should be generated. Also some formatting things that I think the PR will yell at you about. Otherwise you can run `ni-python-styleguide` or `black` to fix most of them.

generated/nidaqmx/system/device.py
WayneDroid: I ran `poetry run ni-python-styleguide lint`, no files changed after running it.
And then I ran `poetry run black generated\nidaqmx\system\device.py`, it fixed the multiple line, however it also formatted other lines as well, eg: capitalize hex codes. Should I push these changes to this PR?

Snippet of the changes:
image

Full changes:
[device.py.diff.txt](https://github.com/ni/nidaqmx-python/files/14018707/device.py.diff.txt)

src/codegen/templates/system/device.py.mako
WayneDroid: I managed to add the calibration info attributes, most likely we don't need separate it to another PR.

> Try appending get_attributes(data, "CalibrationInfo")

Turned out I just need to remove the wanted calibration info attributes from the `EXCLUDED_ATTRIBUTES` list in `attribute_helpers.py`.

generated/nidaqmx/system/device.py
bkeryan: `poetry run ni-python-styleguide lint` runs `black --check`. It doesn't modify any files, and it tolerates some formatting variations.

To reformat files, run `poetry run ni-python-styleguide fix`. Note that this only reformats files that it needs to, so it also tolerates some formatting variations.

Either way, files in the `generated` dir are excluded in both the black config and ni-python-styleguide config, so `poetry run black generated\nidaqmx\system\device.py` should have no effect.

bkeryan: > This looks formatted wrong (I don't think this should be multiple lines), but I think the PR will yell at you

@zhindes It's codegen, so the PR doesn't check formatting.

The reason it's wrapped is because https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/function_template.py.mako#L80 doesn't have a special case to fit all of the parameters on one line.

It's fine.

bkeryan: The build checks are failing because the codegen files in the trunk don't match the output of the codegen script.

```
0s
Run git diff --exit-code
diff --git a/generated/nidaqmx/system/device.py b/generated/nidaqmx/system/device.py
index 92bba3f..545dd32 [1](https://github.com/ni/nidaqmx-python/actions/runs/7622732014/job/20761347475?pr=473#step:9:1)00644
--- a/generated/nidaqmx/system/device.py
+++ b/generated/nidaqmx/system/device.py
@@ -1,6 +1,5 @@
 # Do not edit this file; it was automatically generated.
 
-
 import deprecation
 
 from datetime import datetime
```

Please rerun the codegen script and **don't** run black on the output. The output of the codegen script is fine.

WayneDroid: Ah, it's because I forgot to rerun codegen after I removed extra newline in the device.py.mako.

zhindes: Not checking codegen or modifying it for this specific corner case totally makes sense; my bad!

generated/nidaqmx/system/device.py
```diff
@@ -589,6 +636,68 @@ def bus_type(self):
...
+        self._interpreter.set_cal_info_attribute_uint32(0x2feb, val)
+
+    @property
+    def cal_dev_temp(self):
+        """
+        float: Indicates in degrees Celsius the current temperature of
+            the device.
+        """
+
+        val = self._interpreter.get_cal_info_attribute_double(0x223b)
```
WayneDroid, line 662: @bkeryan @zhindes 
I find this `get_cal_info_attribute_double(0x223b)` is missing `self._name` argument when I write test case, same case on other calibration info attributes. After comparing existing attributes, `handle_parameters` are not available on the calibration info attributes in the [metadate - attributes.py](https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/attributes.py).

```
'handle_parameters': {
                'deviceHandle': {
                    'accessor': 'self._name',
                    'ctypes_data_type': 'ctypes.c_char_p',
                    'cvi_name': 'deviceName'
                }
            },
```

[CalDevTemp attribute](https://github.com/ni/nidaqmx-python/blob/c5fddf5c9dd6be9b0d02bcd96338963176515cf3/src/codegen/metadata/attributes.py#L195)

Is [metadate - attributes.py](https://github.com/ni/nidaqmx-python/blob/master/src/codegen/metadata/attributes.py) handwritten or codegen in another repo?

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
     chassis = device.compact_daq_chassis_device
 
     assert chassis._interpreter is device._interpreter
+
+
+def test___ext_cal_last_date_and_time___no_errors(any_x_series_device: Device) -> None:
+    last_date_and_time = any_x_series_device.ext_cal_last_date_and_time
+
+    assert last_date_and_time is not None
```
zhindes, line 120: At least validate we're getting the right type. I would also like to validate the data a little. A simulated devices seem to return time 0 with the 1904 epoch. A real device should return something sort of reasonable, like make sure the year is >2009? So maybe you create a test with a simulated_x_series and one with a real one? (same for selfcal)

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+    try:
+        current_acc_connection_count = any_x_series_device.cal_acc_connection_count
+
+        any_x_series_device.cal_acc_connection_count = test_value
+        assert test_value == any_x_series_device.cal_acc_connection_count
+
+        any_x_series_device.cal_acc_connection_count = current_acc_connection_count
+
+    except DaqError as e:
+        if e.error_code != DAQmxErrors.ATTR_NOT_SUPPORTED:
```
zhindes, line 147: Looked at the driver, and I confirmed you need a specific real hardware setup for this to be useful. Its a 4353 with a specific calibration accessory. Simulated devices won't support it. So this is good.

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+            raise
+
+
+def test___cal_dev_temp___no_errors(any_x_series_device: Device) -> None:
+    temperature = any_x_series_device.cal_dev_temp
+
+    assert 0.0 != temperature
```
zhindes, line 154: I think simulated returns 0.0 by looking at the code... make this a real_x_series maybe? Or maybe I'm just looking in the wrong place?

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+        if e.error_code != DAQmxErrors.ATTR_NOT_SUPPORTED:
+            raise
+
+
+def test___cal_user_defined_info___no_errors(any_x_series_device: Device) -> None:
+    test_value = "my test value"
+    user_defined_info = any_x_series_device.cal_user_defined_info
+    info_max_size = any_x_series_device.cal_user_defined_info_max_size
+
+    any_x_series_device.cal_user_defined_info = test_value
```
zhindes, line 176: can't `any_x_series_device` be simulated? this wouldn't work

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+
+    except DaqError as e:
+        if e.error_code != DAQmxErrors.ATTR_NOT_SUPPORTED:
+            raise
+
+
+def test___cal_user_defined_info___no_errors(any_x_series_device: Device) -> None:
+    test_value = "my test value"
+    user_defined_info = any_x_series_device.cal_user_defined_info
+    info_max_size = any_x_series_device.cal_user_defined_info_max_size
```
zhindes, line 174: you should limit test_value to max size

zhindes (changes requested):

tests/component/system/test_device_properties.py
zhindes: wait no, last temp is 0. I think this might be something else.

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+        assert test_value == any_x_series_device.cal_acc_connection_count
+
+        any_x_series_device.cal_acc_connection_count = current_acc_connection_count
+
+    except DaqError as e:
+        if e.error_code != DAQmxErrors.ATTR_NOT_SUPPORTED:
+            raise
+
+
+def test___cal_dev_temp___no_errors(any_x_series_device: Device) -> None:
```
zhindes, line 151: I don't see tests for self/ext cal last temp

generated/nidaqmx/system/device.py
zhindes: its codegen'd from scrapigen, I'll point you to it offline

tests/component/system/test_device.py
```diff
@@ -37,3 +37,7 @@ def test___self_test_device___no_errors(any_x_series_device: Device) -> None:
 
 def test___self_cal___no_errors(any_x_series_device: Device) -> None:
     any_x_series_device.self_cal()
+
+
+def test___restore_last_ext_cal_const___no_errors(any_x_series_device: Device) -> None:
```
bkeryan, line 42: Reorder these tests so it runs this one first and self-cal second. Otherwise running the nidaqmx-python tests leaves one of your X Series devices with old calibration constants.

tests/component/system/test_device_properties.py
bkeryan, line 147: Use pytest.raises for error tests, something like this. 
```
with pytest.raises(DaqException) as exc_info:
   _ = any_x_series_device.cal_acc_connection_count

assert exc_info.value.error_code == DAQmxErrors.ATTR_NOT_SUPPORTED
```

Since this is an SC Express specific property and you are testing with X Series, you can assume the attribute always errors.

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+
+    try:
+        current_acc_connection_count = any_x_series_device.cal_acc_connection_count
+
+        any_x_series_device.cal_acc_connection_count = test_value
+        assert test_value == any_x_series_device.cal_acc_connection_count
+
+        any_x_series_device.cal_acc_connection_count = current_acc_connection_count
+
+    except DaqError as e:
```
bkeryan, line 146: This test function is testing too many things. Write separate test functions for get vs. set.

Try to organize the test body into arrange, act, and assert sections separated by one blank line.

Have you seen this? If not, please read it. https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+    assert last_date_and_time is not None
+
+
+def test___device_supports_cal___no_errors(any_x_series_device: Device) -> None:
+    is_cal_supported = any_x_series_device.device_supports_cal
+
+    assert is_cal_supported is True
+
+
+def test___cal_acc_connection_count___no_errors(any_x_series_device: Device) -> None:
```
bkeryan, line 135: The then clause `___no_errors` is not accurate. This test produces and expects an error.

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +112,70 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+
+    assert 0.0 != temperature
+
+
+def test___cal_recommended_acc_connection_count_limit___no_errors(
+    any_x_series_device: Device,
+) -> None:
+    try:
+        acc_connection_count_limit = any_x_series_device.cal_recommended_acc_connection_count_limit
+        print(f"current_acc_connection_count_limit: {acc_connection_count_limit}")
```
bkeryan, line 162: delete debugging code

tests/component/system/test_device_properties.py
bkeryan: yes, there are other fixtures for real/sim X Series: https://github.com/ni/nidaqmx-python/blob/master/tests/conftest.py#L116

bkeryan (changes requested):

tests/component/system/test_device.py
bkeryan: or change these tests to use `sim_x_series_device` to leave the cal constants alone on any real X Series devices

tests/component/system/test_device_properties.py
WayneDroid: I get error -200077 (INVALID_ATTRIBUTE_VALUE) when getting `ext_cal_last_date_and_time` on simulated device, maybe the attribute only available for real device?

image

Cal date on simulated device - `Dev1` on MAX:
image

zhindes: Ah dang, yeah, it must be real-device only. Updating to `real_xseries_device` or whatever its called and doing a bit better validation sounds good

generated/nidaqmx/system/device.py
```diff
@@ -130,6 +131,52 @@ def do_ports(self):
 
     # endregion
 
+    # region Calibration Info property
+
+    @property
+    def ext_cal_last_date_and_time(self):
+        """
+        Indicates the last date and time that the device underwent an
```
bkeryan, line 139: Add the property type.

```suggestion
        datetime: Indicates the last date and time that the device underwent an
```

generated/nidaqmx/system/device.py
```diff
@@ -130,6 +131,52 @@ def do_ports(self):
...
+        external calibration.
```
bkeryan, line 140: PEP-257 says the 1st line of the docstring is a summary. Please run Sphinx and check the output in a web browser to make sure this line wrapping is ok.

generated/nidaqmx/system/device.py
bkeryan, line 139: FYI: Please consider enabling hyperlinks to the stdlib docs for datetime by enabling the intersphinx plugin. See https://github.com/ni/measurementlink-python/pull/430 for an example.

This is not required for this PR. I am posting this suggestion on multiple PRs.

tests/component/system/test_device_properties.py
WayneDroid: I did a quick test, simulated device will return 0; whereas real device will return its current temperature reading. This may vary on different devices and different environment temperature. I'll use `sim_x_series_device` for this test case then.

tests/component/system/test_device_properties.py
WayneDroid: Updated `ext_cal_last_date_and_time` and `self_cal_last_date_and_time` test cases with real x series device.

tests/component/system/test_device_properties.py
WayneDroid: `test_value` is limited to max size in test case.

tests/component/system/test_device_properties.py
WayneDroid: Self/ext cal last temp test cases are added.

tests/component/system/test_device.py
WayneDroid: `self_cal` and `restore_last_ext_cal_const` test cases are updated to test with simulated device.

tests/component/system/test_device_properties.py
WayneDroid: Noted, test case is updated as per suggestion.

tests/component/system/test_device_properties.py
WayneDroid: It's deleted.

tests/component/system/test_device_properties.py
```diff
@@ -112,3 +113,91 @@ def test___module___get_chassis___shared_interpreter(device: Device):
...
+    info_max_size = real_x_series_device.cal_user_defined_info_max_size
+
+    test_value = "my test value"[:info_max_size]
+    real_x_series_device.cal_user_defined_info = test_value
+    value = real_x_series_device.cal_user_defined_info
+
+    assert info_max_size == len(value)
+    assert value in test_value
+
+    real_x_series_device.cal_user_defined_info = user_defined_info
```
bkeryan, line 173: This loses the real X Series device's user defined info in error cases. Put a try block around the arrange/act/assert sections and put the cleanup in a finally block.

FYI, this is testing both read and write in the same test, but in this case there's not an obvious way to split them up.

```suggestion
def test___cal_user_defined_info___no_errors(real_x_series_device: Device) -> None:
    try:
        user_defined_info = real_x_series_device.cal_user_defined_info
        info_max_size = real_x_series_device.cal_user_defined_info_max_size

        test_value = "my test value"[:info_max_size]
        real_x_series_device.cal_user_defined_info = test_value
        value = real_x_series_device.cal_user_defined_info

        assert info_max_size == len(value)
        assert value in test_value
    finally:
        real_x_series_device.cal_user_defined_info = user_defined_info
```

bkeryan (approved):
Approved with suggestions

generated/nidaqmx/system/device.py
zhindes: CONTRIBUTING.md gives details on how to build the documentation

zhindes (approved):
Looks great once Brad's suggestion is in!

generated/nidaqmx/system/device.py
WayneDroid: @zhindes Thanks for pointing how to build the documentation.

After building the documentation, the 1st line of docstring does not wrap in maximum size of browser. And the docstring is wrapped in two lines if browser size is set to half.

browser max size:
image

browser half size:
image

tests/component/system/test_device_properties.py
WayneDroid: `try-finally` block is added in the test case.

WayneDroid:
@zhindes  @bkeryan Any suggestion to resolve the docs build failure? Based on the error message, the pipeline is running Sphinx v4.5.0 whereas the sphinxcontrib.applehelp extension requires Sphinx v5.0.

https://readthedocs.org/projects/nidaqmx-python/builds/23394435/

zhindes:
> @zhindes @bkeryan Any suggestion to resolve the docs build failure? Based on the error message, the pipeline is running Sphinx v4.5.0 whereas the sphinxcontrib.applehelp extension requires Sphinx v5.0.
> 
> https://readthedocs.org/projects/nidaqmx-python/builds/23394435/

That's fixed in mainline. See #487. I think if you rebase your changes on latest master it will be resolved. I'm more worried about you system tests failures.

tests/component/system/test_device.py
```diff
@@ -35,5 +35,9 @@ def test___self_test_device___no_errors(any_x_series_device: Device) -> None:
     any_x_series_device.self_test_device()
 
 
-def test___self_cal___no_errors(any_x_series_device: Device) -> None:
-    any_x_series_device.self_cal()
+def test___restore_last_ext_cal_const___no_errors(sim_x_series_device: Device) -> None:
```
bkeryan, line 38: > nidaqmx.errors.RpcError: StatusCode.UNIMPLEMENTED: This operation is not supported by the NI gRPC Device Server being used. Upgrade NI gRPC Device Server.

Add grpc_xfail to this test:
```suggestion
@pytest.mark.grpc_xfail(
    reason="Requires NI gRPC Device Server version 2.5 or later", raises=RpcError
)
def test___restore_last_ext_cal_const___no_errors(sim_x_series_device: Device) -> None:
```

bkeryan (approved):
Approved with suggestions

tests/component/system/test_device.py
WayneDroid: Thanks for the suggestion, I have marked the test with grpc_xfail as per suggested.

WayneDroid:
> > @zhindes @bkeryan Any suggestion to resolve the docs build failure? Based on the error message, the pipeline is running Sphinx v4.5.0 whereas the sphinxcontrib.applehelp extension requires Sphinx v5.0.
> > https://readthedocs.org/projects/nidaqmx-python/builds/23394435/
> 
> That's fixed in mainline. See #487. I think if you rebase your changes on latest master it will be resolved. I'm more worried about you system tests failures.

Thanks, rebase my branch to latest master resolves the docs build issue.
For the system test failure, it's caused by operation not supported by gRPC device server as what Brad commented. It's resolved after marking `grpc_xfail` on the test case.

maxxboehme (approved):

generated/nidaqmx/system/device.py
WayneDroid: Noted, we could do it in another PR.

generated/nidaqmx/system/device.py
WayneDroid: I think ni-python-styleguide skipped scanning files under `generated` folder as being configured in the `pyproject.toml`.

```
[tool.ni-python-styleguide]
extend_exclude = ".tox,docs,generated,src/codegen/metadata,src/codegen/templates,src/handwritten"
```

bkeryan: Right, we don't run linters on generated code. I forgot that this file is generated

Also, we currently don't run linters on src/handwritten. We ought to fix that at some point, but not in this PR.

generated/nidaqmx/system/device.py
bkeryan: This is resolved, right? After you close this, you can merge the PR.

WayneDroid: Yes, it is resolved. The changes have been reflected `cal_dev_temp`, `ext_cal_last_temp` and `self_cal_last_temp` in the device.py.

image

PR #475, opened 2024-01-24
Add CfgTimeStartTrig to daqmx-python

DeborahOoi96:

[DAQmxCfgTimeStartTrig](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ctrigger.cpp&version=GBmain&line=209&lineEnd=210&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), taking into consideration the AbsoluteTime comments brought up in [Charity's draft PR](https://github.com/ni/nidaqmx-python/pull/471)

It provide support for CfgTimeStartTrig in python.

Ran nidaqmx-python code generator and validated the API was generated correctly
All existing functional test passed.
To keep PRs small I will add trigger functional test coverage in a follow-up PR.
image

### DIFF
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -398,7 +398,7 @@ class LibraryInterpreter(BaseInterpreter):
             with cfunc.arglock:
                 if cfunc.argtypes is None:
                     cfunc.argtypes = [
-                        lib_importer.task_handle, ctypes.CVIAbsoluteTime,
+                        lib_importer.task_handle, _lib_time.AbsoluteTime,
                         ctypes.c_int]
 
         error_code = cfunc(
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -12,7 +12,7 @@ from nidaqmx._base_interpreter import BaseEventHandler, BaseInterpreter
 from nidaqmx._lib import lib_importer, ctypes_byte_str, c_bool32, wrapped_ndpointer
 from nidaqmx.error_codes import DAQmxErrors, DAQmxWarnings
 from nidaqmx.errors import DaqError, DaqReadError, DaqWarning, DaqWriteError
-from nidaqmx._lib_time import AbsoluteTime as LibTimestamp
+from nidaqmx._lib_time import AbsoluteTime
 
 
 _logger = logging.getLogger(__name__)
@@ -393,7 +393,7 @@ class LibraryInterpreter(BaseInterpreter):
 
     def cfg_time_start_trig(self, task, when, timescale):
         cfunc = lib_importer.windll.DAQmxCfgTimeStartTrig
-        when = LibTimestamp.from_datetime(when)
+        when = AbsoluteTime.from_datetime(when)
         if cfunc.argtypes is None:
             with cfunc.arglock:
                 if cfunc.argtypes is None:
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -393,7 +393,6 @@ class LibraryInterpreter(BaseInterpreter):
 
     def cfg_time_start_trig(self, task, when, timescale):
         cfunc = lib_importer.windll.DAQmxCfgTimeStartTrig
-        when = AbsoluteTime.from_datetime(when)
         if cfunc.argtypes is None:
             with cfunc.arglock:
                 if cfunc.argtypes is None:
@@ -402,7 +401,7 @@ class LibraryInterpreter(BaseInterpreter):
                         ctypes.c_int]
 
         error_code = cfunc(
-            task, when, timescale)
+            task, AbsoluteTime.from_datetime(when), timescale)
         self.check_for_error(error_code)
 
     def cfg_watchdog_ao_expir_states(
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -12,6 +12,7 @@ from nidaqmx._base_interpreter import BaseEventHandler, BaseInterpreter
 from nidaqmx._lib import lib_importer, ctypes_byte_str, c_bool32, wrapped_ndpointer
 from nidaqmx.error_codes import DAQmxErrors, DAQmxWarnings
 from nidaqmx.errors import DaqError, DaqReadError, DaqWarning, DaqWriteError
+from nidaqmx._lib_time import AbsoluteTime as LibTimestamp
 
 
 _logger = logging.getLogger(__name__)
@@ -388,6 +389,20 @@ class LibraryInterpreter(BaseInterpreter):
 
         error_code = cfunc(
             task, source, rate, active_edge, sample_mode, samps_per_chan)
+        self.check_for_error(error_code)
+
+    def cfg_time_start_trig(self, task, when, timescale):
+        cfunc = lib_importer.windll.DAQmxCfgTimeStartTrig
+        when = LibTimestamp.from_datetime(when)
+        if cfunc.argtypes is None:
+            with cfunc.arglock:
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        lib_importer.task_handle, ctypes.CVIAbsoluteTime,
+                        ctypes.c_int]
+
+        error_code = cfunc(
+            task, when, timescale)
         self.check_for_error(error_code)
 
     def cfg_watchdog_ao_expir_states(
--- a/generated/nidaqmx/_task_modules/triggering/start_trigger.py
+++ b/generated/nidaqmx/_task_modules/triggering/start_trigger.py
@@ -6,6 +6,7 @@ from nidaqmx.system.physical_channel import _PhysicalChannelAlternateConstructor
 from nidaqmx.constants import (
     Coupling, DigitalPatternCondition, DigitalWidthUnits, Edge, Slope,
     Timescale, TriggerType, WindowTriggerCondition1)
+from datetime import datetime
 
 
 class StartTrigger:
--- a/generated/nidaqmx/_task_modules/triggering/start_trigger.py
+++ b/generated/nidaqmx/_task_modules/triggering/start_trigger.py
@@ -1042,8 +1042,7 @@ class StartTrigger:
         New Start Trigger
 
         Args:
-            when (nidaqmx.constants.DateTime): Specifies when to
-                trigger.
+            when (datetime): Specifies when to trigger.
             timescale (Optional[nidaqmx.constants.Timescale]): Specifies
                 the start trigger timestamp time scale.
         """
--- a/generated/nidaqmx/_task_modules/triggering/start_trigger.py
+++ b/generated/nidaqmx/_task_modules/triggering/start_trigger.py
@@ -1037,6 +1037,20 @@ class StartTrigger:
         self._interpreter.cfg_dig_pattern_start_trig(
             self._handle, trigger_source, trigger_pattern, trigger_when.value)
 
+    def cfg_time_start_trig(self, when, timescale=Timescale.USE_HOST):
+        """
+        New Start Trigger
+
+        Args:
+            when (nidaqmx.constants.DateTime): Specifies when to
+                trigger.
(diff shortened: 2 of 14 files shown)


generated/nidaqmx/_task_modules/triggering/start_trigger.py
```diff
@@ -1037,6 +1037,20 @@ def cfg_dig_pattern_start_trig(
         self._interpreter.cfg_dig_pattern_start_trig(
             self._handle, trigger_source, trigger_pattern, trigger_when.value)
 
+    def cfg_time_start_trig(self, when, timescale=Timescale.USE_HOST):
+        """
+        New Start Trigger
```
bkeryan, line 1042: The first line of a docstring should be a summary. We should figure out why the metadata contains "New Start Trigger" for the description.

https://peps.python.org/pep-0257/

bkeryan (changes requested):

src/codegen/templates/grpc_interpreter/default_grpc_function_call.py.mako
```diff
@@ -41,6 +41,11 @@
         _validate_array_dtype(${parameter_name}, ${parameter_dtype})
     %endfor
 %endif
+%for param in params:
```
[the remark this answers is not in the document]
bkeryan: Correction: you can do the conversion in get_grpc_interpreter_call_params: https://github.com/ni/nidaqmx-python/blob/master/src/codegen/utilities/interpreter_helpers.py#L329

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
```diff
@@ -14,11 +14,22 @@
...
+    argtypes = get_argument_types(function)
+%>\
+%if 'DateTime' in argtypes:
+    <%
+        index = argtypes.index('DateTime')
+        name = function_call_args[index]
+        argtypes[index] = 'ctypes.CVIAbsoluteTime'
+    %>\
+    ${name} = LibTimestamp.from_datetime(${name})
+%endif
```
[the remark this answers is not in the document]
bkeryan: On second thought, generate_interpreter_function_call_args is probably more appropriate: https://github.com/ni/nidaqmx-python/blob/master/src/codegen/utilities/interpreter_helpers.py#L135

get_instantiation_lines_for_output is really for out-parameters.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -390,6 +391,20 @@ def cfg_samp_clk_timing(
...
         self.check_for_error(error_code)
 
+    def cfg_time_start_trig(self, task, when, timescale):
+        cfunc = lib_importer.windll.DAQmxCfgTimeStartTrig
+        when = LibTimestamp.from_datetime(when)
+        if cfunc.argtypes is None:
+            with cfunc.arglock:
+                if cfunc.argtypes is None:
+                    cfunc.argtypes = [
+                        lib_importer.task_handle, ctypes.CVIAbsoluteTime,
```
zhindes, line 401: I'm confused by `ctypes.CVIAbsoluteTime`, where is that defined? `_lib_time.AbsoluteTime` is a ctypes type.

generated/nidaqmx/_task_modules/triggering/start_trigger.py
```diff
@@ -1037,6 +1037,20 @@ def cfg_dig_pattern_start_trig(
...
+
+        Args:
+            when (nidaqmx.constants.DateTime): Specifies when to
```
zhindes, line 1045: `nidaqmx.constants.DateTime` isn't right, it should be either a datetime or hightime

zhindes (changes requested):

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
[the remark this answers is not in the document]
DeborahOoi96: Thanks for the suggestion. 
But since this is an interpreter and the function names being passed are positional arguments, I cannot update function_call_args to do the conversion since that would look like 
`error_code = cfunc(
            task, when=AbsoluteTime.from_datetime(when), timescale)`
and that would result in an error since a positional argument cannot come after a keyword argument.

So I think I need to still have `when = AbsoluteTime.from_datetime(when)` as a separate line in the _library_interpreter file. However I have updated the helpers.py files (generate_interpreter_function_call_args) to do the majority of the parsing jobs and push the name of the dateTime/AbsoluteTime parameters into another list, which will be called and displayed in the mako file.

I have moved the change to argtypes to be directly in to_param_argtype() instead since any special modifications to argument types seem to happen there.

src/codegen/templates/grpc_interpreter/default_grpc_function_call.py.mako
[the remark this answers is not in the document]
DeborahOoi96: Thanks for the suggestion, I have moved the conversion string into get_grpc_interpreter_call_params

generated/nidaqmx/_task_modules/triggering/start_trigger.py
DeborahOoi96: Addressed this comment in a separate PR since this needs changing in grpc-device-scrapigen
https://github.com/ni/grpc-device-scrapigen/pull/213

generated/nidaqmx/_task_modules/triggering/start_trigger.py
DeborahOoi96: This is due to the patch file in nidmxa having incorrect documentation. I have posted a PR on that here https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/640231

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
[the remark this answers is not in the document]
bkeryan: > that would result in an error since a positional argument cannot come after a keyword argument.

Right, you need it to use a positional argument: `error_code = cfunc( task, AbsoluteTime.from_datetime(when), timescale)`

> So I think I need to still have when = AbsoluteTime.from_datetime(when) as a separate line in the _library_interpreter file. 

I disagree. In generate_interpreter_function_call_args , you should specify `f"AbsoluteTime.from_datetime({param.parameter_name})"` instead of `f"{param.parameter_name} = AbsoluteTime.from_datetime({param.parameter_name})"` so that you generate a positional argument rather than a keyword argument.

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
```diff
@@ -7,18 +7,24 @@
     )
     from codegen.utilities.text_wrappers import wrap, docstring_wrap
 
-    function_call_args = generate_interpreter_function_call_args(function)
+    function_call_args, dateTime_args = generate_interpreter_function_call_args(function)
```
bkeryan, line 10: PEP8 says variables are snake_case.

However, this is moot because I think you should delete dateTime_args.

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
```diff
@@ -7,18 +7,24 @@
...
     # samps_per_chan_param includes the keyword argument (samps_per_chan_read=
     # or samps_per_chan_written=)
     samps_per_chan_param = get_samps_per_chan_read_or_write_param(function.base_parameters)
 %>\
         cfunc = lib_importer.${'windll' if function.calling_convention == 'StdCall' else 'cdll'}.DAQmx${function.c_function_name}
+<%
+    argtypes = get_argument_types(function)
+%>\
+%if dateTime_args:
+        ${'\n\t\t'.join(dateTime_args)}
```
bkeryan, line 21: Please don't wrap lines this way.

1. Using \n in a mako template is a red flag. Mako preserves \r\n vs. \n when reading and rendering templates, so many scripts open the output file as binary to avoid newline conversion.
2. This code generator doesn't use tabs, does it?
3. This code generator has a wrap() function for wrapping argument lists.
4. This code generator uses %for for iterating over lines of code.

Also this is moot because I think you should delete dateTime_args.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -182,7 +187,7 @@ def generate_interpreter_function_call_args(function_metadata):
             else:
                 function_call_args.append(param.parameter_name)
 
-    return function_call_args
+    return function_call_args, dateTime_args
```
bkeryan, line 190: I don't like this return value because it's overly specific and not really necessary.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -158,6 +158,11 @@ def generate_interpreter_function_call_args(function_metadata):
             function_call_args.append("None")
         elif is_event_function(function_metadata) and param.parameter_name == "callback_function":
             function_call_args.append("callback_method_ptr")
+        elif param.type == "CVIAbsoluteTime":
```
bkeryan, line 161: Add `and param.direction == "in"`. You will need to use "instantiation lines" to handle the out-parameter for WaitForValidTimestamp.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -158,6 +158,11 @@ def generate_interpreter_function_call_args(function_metadata):
...
+            dateTime_args.append(
+                f"{param.parameter_name} = AbsoluteTime.from_datetime({param.parameter_name})"
+            )
+            function_call_args.append(param.parameter_name)
```
bkeryan, line 165:
```suggestion
            function_call_args.append(f"AbsoluteTime.from_datetime({param.parameter_name})")
```

bkeryan (changes requested):

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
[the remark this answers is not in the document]
bkeryan: Yeah, I think this code is already too complicated and inconsistent, and adding another way to generate variable declarations for parameters makes it worse, not better. Please extend the existing mechanisms.

zhindes (approved):
Only major concern with the generated files is with the docstring, but you're fixing that upstream. I'm fine with waiting to complete this PR until that is ready or coming back to it.

For the code generator, Brad is on it.

generated/nidaqmx/_task_modules/triggering/start_trigger.py
bkeryan, line 1045: FYI: Please consider enabling hyperlinks to the stdlib docs for datetime by enabling the intersphinx plugin. See https://github.com/ni/measurementlink-python/pull/430 for an example.

This is not required for this PR. I am posting this suggestion on multiple PRs.

DeborahOoi96: I think in the interest of small PRs either @WayneDroid or I will address this in a follow up PR for nidaqmx-python. Thanks for the heads up though!

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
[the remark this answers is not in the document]
DeborahOoi96: Thanks for clarifying, I understand what you mean now. I have updated the codegen and the file accordingly to just use existing mechanisms. As this PR only covers CfgStartTimeTrigger, I will handle WaitForValidTimestamp in a separate PR.

src/codegen/utilities/interpreter_helpers.py
DeborahOoi96: Moved this to be further down in `param.direction == "in"` if-else loop

generated/nidaqmx/_task_modules/triggering/start_trigger.py
bkeryan, line 1: Please add `from datetime import datetime` to the top of start_trigger.py.mako so that Sphinx knows this is a reference to the datetime type, not the datetime module.

bkeryan (approved):

generated/nidaqmx/_task_modules/triggering/start_trigger.py
DeborahOoi96: added

PR #479, opened 2024-01-31
Add support for WaitForValidTimestamp to nidaqmx-python

DeborahOoi96:

[DAQmxWaitForValidTimestamp](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicai/source/nicai/ctask.cpp&version=GC83afeb252672d4420622e056eb9aab8930cae7d6&line=1252&lineEnd=1252&lineStartColumn=23&lineEndColumn=49&lineStyle=plain&_a=contents)

It provides support for WaitForValidTimestamp in python.

Ran nidaqmx-python code generator and validated the API was generated correctly
All existing functional test passed.
image
Increasing functional test coverage to cover this function will be done in another PR.

### DIFF
--- a/generated/nidaqmx/_base_interpreter.py
+++ b/generated/nidaqmx/_base_interpreter.py
@@ -1662,6 +1662,10 @@ class BaseInterpreter(abc.ABC):
         raise NotImplementedError
 
     @abc.abstractmethod
+    def wait_for_valid_timestamp(self, task, timestamp_event, timeout):
+        raise NotImplementedError
+
+    @abc.abstractmethod
     def wait_until_task_done(self, task, time_to_wait):
         raise NotImplementedError
 
--- a/src/handwritten/task.py
+++ b/src/handwritten/task.py
@@ -1027,13 +1027,13 @@ class Task:
 
         Args:
             timestamp_event(nidaqmx.constants.TimestampEvent): Specifies the timestamp type to wait on. 
-            timeout (Optional[float]): Specifies the maximum amount of time in
+            timeout (float): Specifies the maximum amount of time in
                 seconds to wait for a valid timestamp.
                 This method returns an error if the time elapses. The
                 default is 10. If you set timeout (sec) to
                 nidaqmx.WAIT_INFINITELY, the method waits indefinitely.
         """
-        self._interpreter.wait_for_valid_timestamp(self._handle, timestamp_event, timeout)
+        self._interpreter.wait_for_valid_timestamp(self._handle, timestamp_event.value, timeout)
 
     def wait_until_done(self, timeout=10.0):
         """
--- a/src/handwritten/task.py
+++ b/src/handwritten/task.py
@@ -1023,7 +1023,7 @@ class Task:
         """
         Wait until the specified timestamp has a value.
 
-        Use this VI to ensure the timestamp has a valid value to prevent an error when querying a timestamp value.
+        Use this method to ensure the timestamp has a valid value to prevent an error when querying a timestamp value.
 
         Args:
             timestamp_event(nidaqmx.constants.TimestampEvent): Specifies the timestamp type to wait on. 
--- a/src/handwritten/task.py
+++ b/src/handwritten/task.py
@@ -1019,6 +1019,24 @@ class Task:
         """
         self._interpreter.stop_task(self._handle)
 
+    def wait_for_valid_timestamp(self, timestamp_event, timeout=10.0):
+        """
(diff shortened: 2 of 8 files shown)


generated/nidaqmx/_base_interpreter.py
bkeryan, line 1: Adding the function to the internal interpreter classes is the first step. The next step is to add methods or properties to the public API.

For each timestamp event, you should edit the corresponding class to add a `wait_for_valid_timestamp()` method or something similar. These methods will hardcode the timestamp event parameter.

These are the events:
```
class TimestampEvent(Enum):
    START_TRIGGER = 12491  #: Start Trigger timestamp.
    REFERENCE_TRIGGER = 12490  #: Reference Trigger timestamp.
    ARM_START_TRIGGER = 14641  #: Arm Start Trigger timestamp.
    FIRST_SAMPLE = 16130  #: First Sample timestamp.
```
so edit these classes:
- https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/triggering/start_trigger.py
- https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/triggering/reference_trigger.py
- https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/triggering/arm_start_trigger.py
- https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_task_modules/timing.py -- this one is special: the method should be named something like wait_for_first_samp_timestamp

bkeryan (changes requested):

generated/nidaqmx/_base_interpreter.py
bkeryan: (An alternative is to expose a single `wait_for_valid_timestamp()` method on the `Task` or `Timing` class and have the user pass in the timestamp event parameter, but I don't think this fits into the design of the Python trigger APIs as well.)

zhindes (commented):
Brad got it all!

DeborahOoi96:
Not done yet, only rebased so far but haven't addressed comments yet.

generated/nidaqmx/_base_interpreter.py
DeborahOoi96: It would be complicated to make this trigger-specific as the python class name for wait_for_valid_timestamp is `Task`. 

The slightly more straightforward approach, codegen-wise, would be to just proceed with your alternate suggestion and add a single method to Task class, and have the user pass in the timestamp event parameter.

It's still possible to make it trigger-specific as per your original suggestion, but in my opinion it would require a significant amount of changes to the codegen. I'm still working on this, haven't pushed my changes yet because it's still under progress. Just wondering if it would be worth all the effort to proceed with this way. What do you think?

bkeryan: The easiest way to implement the trigger-specific approach is to set `'python_codegen_method': 'CustomCode',` and add a hand-written implementation to each trigger template.

I now see that the .NET API uses the alternate approach (`Task.WaitForValidTimestamp` is right next to `Task.WaitUntilDone`), so I guess the alternate approach is ok. @zhindes do you agree?

zhindes (approved):

generated/nidaqmx/_base_interpreter.py
DeborahOoi96: I have update it to be task-based

src/handwritten/task.py
```diff
@@ -1019,6 +1019,24 @@ def stop(self):
...
+            timestamp_event(nidaqmx.constants.TimestampEvent): Specifies the timestamp type to wait on. 
+            timeout (Optional[float]): Specifies the maximum amount of time in
+                seconds to wait for the measurement or generation to complete.
+                This method returns an error if the time elapses. The
+                default is 10. If you set timeout (sec) to
+                nidaqmx.WAIT_INFINITELY, the method waits indefinitely. If you
+                set timeout (sec) to 0, the method checks once and returns
+                an error if the measurement or generation is not done.
+        """
+        self._interpreter.wait_for_valid_timestamp(self._handle, timestamp_event, timeout)
```
bkeryan, line 1038: The interpreter methods expect ints, not enums. A test would have caught this.
```suggestion
        self._interpreter.wait_for_valid_timestamp(self._handle, timestamp_event.value, timeout)
```

bkeryan (changes requested):

maxxboehme (approved):

bkeryan (approved):

PR #481, opened 2024-02-01
Add "When" attribute to start_trigger.py

DeborahOoi96:

Adds the `when` attribute to start_trigger.py. This will enable users to get and set the when attribute when using Start Trigger.
Implemented `GetTrigAttributeTimestamp` and `SetTrigAttributeTimestamp` as it seems start_trig_trig_when uses these functions when getting `when` attribute.
Also removed `_lib_time` from interpreters as this causes errors in unit testing.

Noticed this was not available when I was working on unit test for adding CfgTimeStartTrig function.

Manual validation that exported files are as expected.


src/codegen/utilities/interpreter_helpers.py
```diff
@@ -64,14 +64,12 @@
     "GetSyncPulseTimeWhen",
     "GetTimingAttributeExTimestamp",
     "GetTimingAttributeTimestamp",
-    "GetTrigAttributeTimestamp",
```
DeborahOoi96, line 67: I implemented these functions since it seems `Start_Trig_Trig_When` uses these functions when we set or get that attribute.

src/codegen/templates/_task_modules/triggering/start_trigger.py.mako
```diff
@@ -3,7 +3,7 @@
     from codegen.utilities.function_helpers import get_functions, get_enums_used as functions_enums
     from codegen.utilities.attribute_helpers import get_attributes,  get_enums_used as attribute_enums
     from codegen.utilities.helpers import get_enums_to_import
-    attributes = get_attributes(data, "StartTrigger")
+    attributes = get_attributes(data, "StartTrigger", "TimeStartTrigger")
```
bkeryan, line 6: You shouldn't have to do this here. Scrapigen should translate TimeStartTrigger to StartTrigger like it does for DigitalEdgeStartTrigger, AnalogWindowStartTrigger, etc.

src/codegen/utilities/attribute_helpers.py
```diff
@@ -240,14 +243,17 @@
...
     attributes_metadata = []
     for group_name, attributes in metadata["attributes"].items():
         for id, attribute_data in attributes.items():
             if (
                 "python_class_name" in attribute_data
-                and attribute_data["python_class_name"] == class_name
+                and (
+                    attribute_data["python_class_name"] == class_name
+                    or attribute_data["python_class_name"] == additional_class_name
+                )
```
bkeryan, line 256: I think you should update scrapigen to generate the right `python_class_name` and revert this `additional_class_name` change.

src/codegen/metadata/attributes.py
```diff
@@ -28949,7 +28949,7 @@
             'is_python_object': False,
             'name': 'START_TRIG_TRIG_WHEN',
             'python_class_name': 'TimeStartTrigger',
-            'python_data_type': 'unknown',
+            'python_data_type': 'datetime',
```
bkeryan, line 28952: I assume you will update the rest of these in a follow-up PR.

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -64,14 +64,12 @@
...
     "SetArmStartTrigTrigWhen",
     "SetFirstSampClkWhen",
     "SetStartTrigTrigWhen",
     "SetSyncPulseTimeWhen",
     "SetTimingAttributeExTimestamp",
     "SetTimingAttributeTimestamp",
```
bkeryan, line 72: FYI, you will need these when you add the rest of the attributes, since first_samp_timestamp_val and sync_pulse_when are timing attributes.

bkeryan (changes requested):

src/codegen/templates/_task_modules/triggering/start_trigger.py.mako
DeborahOoi96: I have updated scrapigen to correctly generate StartTrigger instead of TimeStartTrigger

src/codegen/utilities/interpreter_helpers.py
DeborahOoi96: Noted with thanks, I will add support for all of these functions at one shot in another PR

src/codegen/metadata/attributes.py
DeborahOoi96: Correct me if I'm wrong but I think there is not much else to update with StartTrigTrigWhen. If you are referring to the other time trigger attributes in this file then yes, I will update them in a follow up PR.

bkeryan: Yes, I'm referring to the other time trigger attributes. Thanks.

bkeryan (approved):

generated/nidaqmx/_task_modules/triggering/start_trigger.py
bkeryan, line 1: The "check for files dirtied by codegen" is failing because the property order changed and the arm_start property is now generated.

generated/nidaqmx/_task_modules/triggering/start_trigger.py
```diff
@@ -832,6 +832,23 @@ def retriggerable(self, val):
     def retriggerable(self):
         self._interpreter.reset_trig_attribute(self._handle, 0x190f)
 
+    @property
+    def start_trig_trig_when(self):
```
bkeryan, line 836: Sorry I didn't notice this before, but the codegen strips `start_trig_` from the other `StartTrigger` properties.

For example:
- `START_TRIG_TYPE` -> `trig_type` (this one might be special, since it still has trig)
- `ANLG_EDGE_START_TRIG_HYST` -> `anlg_edge_hyst`
- `DIG_EDGE_START_TRIG_SRC` -> `dig_edge_src`

So I think these new properties should be `time_when`.

This will likely require another round of scrapigen changes, so it's ok to do this in a separate PR (maybe with the rest of the time trigger attributes).

DeborahOoi96: I will address this in a separate PR. Thanks.

DeborahOoi96: Addressed in https://github.com/ni/grpc-device-scrapigen/pull/216

PR #488, opened 2024-02-05
Add Unit test for cfg_time_start_trig

DeborahOoi96:

Add unit test coverage for cfg_time_start_trig
Add unit test coverage for new property introduced in https://github.com/ni/nidaqmx-python/pull/481
Created a new file called test_triggers.py because the existing test_triggers.py seems to use legacy testing environment and I foresee we will be adding a lot of new test coverage for time trigger APIs.
Made fixes to get_trig_attribute_timestamp as we found some missing conversion there.
Updated conftest.py to actually pull the simulated device.

Adds unit test coverage

All tests passed including the new tests
image
image
image

### DIFF
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -8,6 +8,7 @@ import warnings
 from typing import Callable, Generic, Optional, TypeVar
 
 import google.protobuf.message
+from google.protobuf.timestamp_pb2 import Timestamp as GrpcTimestamp
 import grpc
 import numpy
 
@@ -380,10 +381,11 @@ class GrpcStubInterpreter(BaseInterpreter):
                 samps_per_chan=samps_per_chan))
 
     def cfg_time_start_trig(self, task, when, timescale):
+        ts = GrpcTimestamp()
         response = self._invoke(
             self._client.CfgTimeStartTrig,
             grpc_types.CfgTimeStartTrigRequest(
-                task=task, when=convert_time_to_timestamp(when),
+                task=task, when=convert_time_to_timestamp(when,ts),
                 timescale_raw=timescale))
 
     def cfg_watchdog_ao_expir_states(
@@ -3133,11 +3135,12 @@ class GrpcStubInterpreter(BaseInterpreter):
                 task=task, attribute_raw=attribute, value=value))
 
     def set_trig_attribute_timestamp(self, task, attribute, value):
+        ts = GrpcTimestamp()
         response = self._invoke(
             self._client.SetTrigAttributeTimestamp,
             grpc_types.SetTrigAttributeTimestampRequest(
                 task=task, attribute_raw=attribute,
-                value=convert_time_to_timestamp(value)))
+                value=convert_time_to_timestamp(value,ts)))
 
     def set_trig_attribute_uint32(self, task, attribute, value):
         response = self._invoke(
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -381,11 +381,10 @@ class GrpcStubInterpreter(BaseInterpreter):
                 samps_per_chan=samps_per_chan))
 
     def cfg_time_start_trig(self, task, when, timescale):
-        ts = GrpcTimestamp()
         response = self._invoke(
             self._client.CfgTimeStartTrig,
             grpc_types.CfgTimeStartTrigRequest(
-                task=task, when=convert_time_to_timestamp(when,ts),
+                task=task, when=convert_time_to_timestamp(when),
                 timescale_raw=timescale))
 
     def cfg_watchdog_ao_expir_states(
@@ -3135,12 +3134,11 @@ class GrpcStubInterpreter(BaseInterpreter):
                 task=task, attribute_raw=attribute, value=value))
 
     def set_trig_attribute_timestamp(self, task, attribute, value):
-        ts = GrpcTimestamp()
         response = self._invoke(
             self._client.SetTrigAttributeTimestamp,
             grpc_types.SetTrigAttributeTimestampRequest(
                 task=task, attribute_raw=attribute,
-                value=convert_time_to_timestamp(value,ts)))
+                value=convert_time_to_timestamp(value)))
 
     def set_trig_attribute_uint32(self, task, attribute, value):
         response = self._invoke(
--- a/generated/nidaqmx/_grpc_time.py
+++ b/generated/nidaqmx/_grpc_time.py
@@ -5,6 +5,8 @@ from typing import Optional, Union
 
 from google.protobuf.timestamp_pb2 import Timestamp as GrpcTimestamp
 
+# 66 years, 17 leap days = 24107 days = 2082844800 seconds
+_BIAS_FROM_1970_EPOCH = 2082844800
 
 _NS_PER_S = 10**9
 _NS_PER_US = 10**3
@@ -44,8 +46,8 @@ def convert_to_local_timezone(expected_time_utc):
 
 def negative_timestamp_to_1904_epoch(negative_timestamp):
     epoch_1904 = ht_datetime(1904, 1, 1)
-    delta = timedelta(seconds=abs(negative_timestamp))
-    datetime_1904 = epoch_1904 - delta
+    seconds_from_1904 = timedelta(seconds = _BIAS_FROM_1970_EPOCH - abs(negative_timestamp))
+    datetime_1904 = epoch_1904 + seconds_from_1904
     return datetime_1904
 
 
--- a/generated/nidaqmx/_grpc_time.py
+++ b/generated/nidaqmx/_grpc_time.py
@@ -14,7 +14,7 @@ _YS_PER_NS = 10**15
 _YS_PER_FS = 10**9
 
 
-def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTimestamp) -> None:
+def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: Optional[GrpcTimestamp] = GrpcTimestamp()) -> GrpcTimestamp:
     utc_dt = dt.astimezone(tz=timezone.utc)
     seconds = int(utc_dt.timestamp())
 
@@ -30,6 +30,7 @@ def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTime
         nanos = utc_dt.microsecond * _NS_PER_US
 
     ts.FromNanoseconds(seconds * _NS_PER_S + nanos)
+    return ts
 
 
 def convert_timestamp_to_time(ts: GrpcTimestamp, tzinfo: Optional[timezone] = None) -> ht_datetime:
--- a/generated/nidaqmx/_grpc_time.py
+++ b/generated/nidaqmx/_grpc_time.py
@@ -60,7 +60,7 @@ def convert_timestamp_to_time(ts: GrpcTimestamp, tzinfo: Optional[timezone] = No
     yoctosecond = remainder_yoctoseconds
 
     if seconds < 0:
-        dt = negative_timestamp_to_1904_epoch(negative_timestamp)
+        dt = negative_timestamp_to_1904_epoch(seconds)
     else:
         dt = ht_datetime.fromtimestamp(seconds, timezone.utc)
 
--- a/generated/nidaqmx/_grpc_time.py
+++ b/generated/nidaqmx/_grpc_time.py
@@ -14,9 +14,11 @@ _YS_PER_NS = 10**15
 _YS_PER_FS = 10**9
 
 
-def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTimestamp) -> None:
+def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: Optional[GrpcTimestamp] = None) -> GrpcTimestamp:
     utc_dt = dt.astimezone(tz=timezone.utc)
     seconds = int(utc_dt.timestamp())
+    if ts is None:
+        ts = GrpcTimestamp()
 
     if isinstance(dt, ht_datetime):
         total_yoctoseconds = dt.yoctosecond
@@ -30,6 +32,7 @@ def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTime
         nanos = utc_dt.microsecond * _NS_PER_US
 
     ts.FromNanoseconds(seconds * _NS_PER_S + nanos)
+    return ts
 
 
 def convert_timestamp_to_time(ts: GrpcTimestamp, tzinfo: Optional[timezone] = None) -> ht_datetime:
--- a/generated/nidaqmx/_grpc_time.py
+++ b/generated/nidaqmx/_grpc_time.py
@@ -1,4 +1,4 @@
-from datetime import timezone
+from datetime import timezone, timedelta
 from datetime import datetime as std_datetime
 from hightime import datetime as ht_datetime
 from typing import Optional, Union
@@ -43,7 +43,7 @@ def convert_to_local_timezone(expected_time_utc):
 
 
 def negative_timestamp_to_1904_epoch(negative_timestamp):
-    epoch_1904 = datetime(1904, 1, 1)
+    epoch_1904 = ht_datetime(1904, 1, 1)
     delta = timedelta(seconds=abs(negative_timestamp))
     datetime_1904 = epoch_1904 - delta
     return datetime_1904
--- a/generated/nidaqmx/_grpc_time.py
+++ b/generated/nidaqmx/_grpc_time.py
@@ -1,13 +1,10 @@
 from datetime import timezone
 from datetime import datetime as std_datetime
 from hightime import datetime as ht_datetime
-from hightime import timedelta as ht_timedelta
 from typing import Optional, Union
 
 from google.protobuf.timestamp_pb2 import Timestamp as GrpcTimestamp
 
-# 66 years, 17 leap days = 24107 days = 2082844800 seconds
-_BIAS_FROM_1970_EPOCH = 2082844800
 
 _NS_PER_S = 10**9
 _NS_PER_US = 10**3
@@ -17,11 +14,9 @@ _YS_PER_NS = 10**15
 _YS_PER_FS = 10**9
 
 
-def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: Optional[GrpcTimestamp] = None) -> GrpcTimestamp:
+def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTimestamp) -> None:
     utc_dt = dt.astimezone(tz=timezone.utc)
     seconds = int(utc_dt.timestamp())
-    if ts is None:
-        ts = GrpcTimestamp()
 
     if isinstance(dt, ht_datetime):
         total_yoctoseconds = dt.yoctosecond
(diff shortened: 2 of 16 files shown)


tests/conftest.py
```diff
@@ -113,6 +113,28 @@ def _x_series_device(device_type, system):
...
+            and len(device.ai_physical_chans) >= 1
+        ):
+            return device
+
+    pytest.skip(
+        "Could not detect a device that meets the requirements to be an Field DAQ fixture of type "
+        f"{device_type}. Cannot proceed to run tests. Import the NI MAX configuration file located "
+        "at nidaqmx\\tests\\max_config\\nidaqmxMaxConfig.ini to create these devices."
+    )
+    return None
```
zhindes, line 135: I think we were a little too loose with these methods. For time-aware, specifically, a real device might act quite a bit differently than a simulated one. I would recommend you use the new method I added `_device_by_product_type` and implement a fixture called `sim_fdaq_device` or maybe even `sim_time_device`

zhindes (changes requested):

tests/component/test_triggers.py
bkeryan, line 1: Organize component/unit tests to mirror the SUT (system under test).

This is the test for nidaqmx._task_modules.triggers, so it should be tests/component/_task_modules/test_triggers.py, in the same directory as test_triggers_properties.py.

tests/component/_task_modules/test_triggers_properties.py
```diff
@@ -99,3 +109,43 @@ def test___ai_task___reset_uint32_property___returns_default_value(ai_voltage_ta
...
     assert ai_voltage_task.triggers.reference_trigger.pretrig_samples == 2
+
+
+def test___ai_voltage_field_daq_task___get_timestamp_property___returns_default_value(
+    ai_voltage_field_daq_task: Task,
+):
+    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
+    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
+    assert when_value.lsb == 0
+    assert when_value.msb == 0
```
bkeryan, line 120: arrange, act, assert
```suggestion
    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
    
    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
    
    assert when_value.lsb == 0
    assert when_value.msb == 0
```

tests/component/_task_modules/test_triggers_properties.py
```diff
@@ -99,3 +109,43 @@ def test___ai_task___reset_uint32_property___returns_default_value(ai_voltage_ta
...
+    ai_voltage_field_daq_task.triggers.start_trigger.trig_when = value_to_test
+    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
+
+    when_value_dt = when_value.to_datetime(timezone.utc)
+    assert when_value_dt.year == 2002
+    assert when_value_dt.month == 1
+    assert when_value_dt.day == 1
+    assert when_value_dt.hour == 0
+    assert when_value_dt.minute == 0
+    assert when_value_dt.second == 0
```
bkeryan, line 138: arrange, act, assert

"act" is "set timestamp property". Reading it back is part of "assert". 
```suggestion
    value_to_test = JAN_01_2002_HIGHTIME
    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
    
    ai_voltage_field_daq_task.triggers.start_trigger.trig_when = value_to_test
    
    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
    when_value_dt = when_value.to_datetime(timezone.utc)
    assert when_value_dt.year == 2002
    assert when_value_dt.month == 1
    assert when_value_dt.day == 1
    assert when_value_dt.hour == 0
    assert when_value_dt.minute == 0
    assert when_value_dt.second == 0
```

tests/component/_task_modules/test_triggers_properties.py
```diff
@@ -99,3 +109,43 @@ def test___ai_task___reset_uint32_property___returns_default_value(ai_voltage_ta
...
+    ai_voltage_field_daq_task: Task,
+):
+    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
+    ai_voltage_field_daq_task.triggers.start_trigger.trig_when = JAN_01_2002_HIGHTIME
+
+    del ai_voltage_field_daq_task.triggers.start_trigger.trig_when
+    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
+
+    assert when_value.lsb == 0
+    assert when_value.msb == 0
```
bkeryan, line 151: arrange, act, assert
```suggestion
    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
    ai_voltage_field_daq_task.triggers.start_trigger.trig_when = JAN_01_2002_HIGHTIME

    del ai_voltage_field_daq_task.triggers.start_trigger.trig_when
    
    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
    assert when_value.lsb == 0
    assert when_value.msb == 0
```

tests/component/test_triggers.py
```diff
@@ -0,0 +1,54 @@
...
+    timescale_value = ai_voltage_field_daq_task.triggers.start_trigger.timestamp_timescale
+
+    when_value_dt = when_value.to_datetime()
+    assert timescale_value == Timescale.USE_HOST
+    assert when_value_dt.year == trigger_time.year
+    assert when_value_dt.month == trigger_time.month
+    assert when_value_dt.day == trigger_time.day
+    assert when_value_dt.hour == trigger_time.hour
+    assert when_value_dt.minute == trigger_time.minute
+    assert when_value_dt.second == trigger_time.second
```
bkeryan, line 35: arrange, act, assert

"act" is "cfg_time_start_trig". Reading it back is part of "assert".
```suggestion
    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
    trigger_time = ht_datetime.now() + timedelta(seconds=10)

    ai_voltage_field_daq_task.triggers.start_trigger.cfg_time_start_trig(trigger_time)
    
    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
    timescale_value = ai_voltage_field_daq_task.triggers.start_trigger.timestamp_timescale
    when_value_dt = when_value.to_datetime()
    assert timescale_value == Timescale.USE_HOST
    assert when_value_dt.year == trigger_time.year
    assert when_value_dt.month == trigger_time.month
    assert when_value_dt.day == trigger_time.day
    assert when_value_dt.hour == trigger_time.hour
    assert when_value_dt.minute == trigger_time.minute
    assert when_value_dt.second == trigger_time.second
```

tests/component/test_triggers.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test___arguments_provided___cfg_time_start_trig___no_errors(
+    ai_voltage_field_daq_task: Task,
+):
+    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
+    trigger_time = ht_datetime.now() + timedelta(seconds=10)
+    timescale = Timescale.USE_HOST
+
+    ai_voltage_field_daq_task.triggers.start_trigger.cfg_time_start_trig(trigger_time, timescale)
```
bkeryan, line 45: This test case would be more effective if the provided argument didn't match the default value used by both the Python API and the driver.

Please specify USE_IO_DEVICE here and assert that the timescale attribute is set to USE_IO_DEVICE.

tests/conftest.py
bkeryan: Agreed, but please spell out `field_daq` instead of abbreviating it `fdaq`.

bkeryan (changes requested):

tests/component/test_triggers.py
DeborahOoi96: Moved

tests/component/_task_modules/test_triggers_properties.py
```diff
@@ -99,3 +109,45 @@ def test___ai_task___reset_uint32_property___returns_default_value(ai_voltage_ta
...
+    ai_voltage_field_daq_task: Task,
+):
+    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
+    ai_voltage_field_daq_task.triggers.start_trigger.trig_when = JAN_01_2002_HIGHTIME
+
+    del ai_voltage_field_daq_task.triggers.start_trigger.trig_when
+
+    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
+    assert when_value.lsb == 0
+    assert when_value.msb == 0
```
bkeryan, line 153: Wait a minute, `hightime` doesn't have `lsb` and `msb` fields. This property is returning the wrong data type.

`set_trig_attribute_timestamp` is correctly converting from `hightime`/`datetime` to `AbsoluteTime`: https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_library_interpreter.py#L5350

`get_trig_attribute_timestamp` should convert from `AbsoluteTime` back to `hightime` before returning, but it's not doing this: https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_library_interpreter.py#L3674

zhindes (approved):

bkeryan (changes requested):

tests/component/_task_modules/test_triggers.py
```diff
@@ -0,0 +1,55 @@
...
+    ai_voltage_field_daq_task: Task,
+):
+    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
+    trigger_time = ht_datetime.now() + timedelta(seconds=10)
+
+    ai_voltage_field_daq_task.triggers.start_trigger.cfg_time_start_trig(trigger_time)
+
+    when_value = ai_voltage_field_daq_task.triggers.start_trigger.trig_when
+    timescale_value = ai_voltage_field_daq_task.triggers.start_trigger.timestamp_timescale
+    when_value_dt = when_value.to_datetime()
```
bkeryan, line 28: Tests should not have to do this.

This is LibraryInterpreter-specific and it should fail with GrpcStubInterpreter.

tests/component/_task_modules/test_triggers.py
```diff
@@ -0,0 +1,55 @@
+from datetime import timedelta
+
+import pytest
+from hightime import datetime as ht_datetime
+
+from nidaqmx._lib_time import AbsoluteTime
```
bkeryan, line 6:
```
Run poetry run ni-python-styleguide lint
./tests/component/_task_modules/test_triggers.py:6:1: F401 'nidaqmx._lib_time.AbsoluteTime' imported but unused
Error: Process completed with exit code 255.
```

tests/component/_task_modules/test_triggers_properties.py
DeborahOoi96: You are right, I have fixed this in the next commit. 

However, this introduced a whole new level of complexity. According to the logic in `AbsoluteTime`, the default value is lsb = 0 and msb = 0, which means it should be 1904-01-01. However, I am getting OS error when trying to use `fromtimestamp()` with the `_BIAS_FROM_1970_EPOCH` which is -2082844800. [According to the documentation ](https://docs.python.org/3/library/datetime.html#:~:text=This%20may%20raise%20OverflowError,ValueError%20on%20localtime()%20failure), it seems like the number is out of scope of the supported range. [Same goes to astimezone()](https://bugs.python.org/issue36759#:~:text=This%20exception%20is,raise%20a%20ValueError%3F). 

So I implemented a small change in _lib_time which helps us address this small corner case. 

Might be good to note that `fromtimestamp()` does pass with -2082844800 when I run it on an online python compiler, so this leads me to believe it could be a python issue. I hit OS error also when I try to run it on Python 3.6.

DeborahOoi96:
`test___arguments_provided___cfg_time_start_trig___no_errors` is failing because timestamp stays at HOST instead of I/O TIME. I suspect this is a simulated device issue, since I get the same behaviour in LabView. I don't have a physical device currently that supports time triggering that can prove my theory, so I will update this test case once I know if it's an expected behavior on simulated device or not.

zhindes:
> `test___arguments_provided___cfg_time_start_trig___no_errors` is failing because timestamp stays at HOST instead of I/O TIME. I suspect this is a simulated device issue, since I get the same behaviour in LabView. I don't have a physical device currently that supports time triggering that can prove my theory, so I will update this test case once I know if it's an expected behavior on simulated device or not.

That doesn't surprise me, but I'm having trouble finding specific driver code that would do this. If you validate in LV, then that's good. Another thing to validate is use NI I/O Trace to ensure the settings going to the driver from Python are what you expect.

tests/component/_task_modules/test_triggers_properties.py
zhindes: Argh, that's unfortunate. I don't like your workaround in `AbsoluteTime`. It is a hack that only works for time 0, which is isn't a relevant use-case. It is unfortunate that simulated devices have this behavior, but I don't think we should work around it. You could either make the test expect the exception or use a real device or both. I'm leaning towards both. For the real-device test:
* Getting a time-aware device is not always easy
* I'm not sure we'd want to force the driver ownership team to validate that on release every time
* **but** writing the test and performing a one-time functional validation with real hardware sounds like a great idea. I think we'll run into more of these, so I'm fine delaying that final validation with real hardware until all the time stuff is in.

bkeryan: There is a simpler way to handle dates before 1970:
- Construct a `datetime` set to the 1904-01-01 epoch
- Add `timedelta(seconds=msb)` to it
- Add `timedelta(seconds=lsb/(2**64))` to it

See https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nidaqmxSharp/source/Api/DaqTypeConversion.cs

DeborahOoi96:
@zhindes Got my hands on a physical cdaq-9189, yes it is able to set timescale to I/O time. So then I guess it is a simulated device issue. I will proceed to use IO Trace later and update the results as well.
image

tests/component/_task_modules/test_triggers_properties.py
DeborahOoi96: @zhindes 

> It is unfortunate that simulated devices have this behavior, but I don't think we should work around it.

It seems that even physical devices's default value for Start Time When is 0 Absolute Time.

Default value on physical 9189
image

Default value on simulated fieddaq
image

zhindes: @DeborahOoi96 Those screenshots aren't doing what you think. They are returning the default value of the Start.Time.When attribute, which is 0. That's a reasonable default and will make your `reset` test a bit of a challenge. However, if you set that to a non-default value, it should maintain it unless there is some weird driver behavior doing something different.

> based on Zach's comment I think he suggests moving towards not having this kind of hack in libtime since it is not a relevant use case.

I definitely don't want a hack in libtime, but what Brad described doesn't seem like a hack to me.

```python
>>> time0 = datetime.datetime(1904,1,1)
>>> td = datetime.timedelta(seconds=10)
>>> print(time0)
1904-01-01 00:00:00
>>> print(td)
0:00:10
>>> time0 + td
datetime.datetime(1904, 1, 1, 0, 0, 10)
>>> time0 - td
datetime.datetime(1903, 12, 31, 23, 59, 50)
```

zhindes: To be more clear, handling `time==0` specifically as some special case is not great. Having some generic code that handles times before the unix 1970 epox is fine.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3133,11 +3135,12 @@ def set_trig_attribute_string(self, task, attribute, value):
                 task=task, attribute_raw=attribute, value=value))
 
     def set_trig_attribute_timestamp(self, task, attribute, value):
+        ts = GrpcTimestamp()
         response = self._invoke(
             self._client.SetTrigAttributeTimestamp,
             grpc_types.SetTrigAttributeTimestampRequest(
                 task=task, attribute_raw=attribute,
-                value=convert_time_to_timestamp(value)))
+                value=convert_time_to_timestamp(value,ts)))
```
bkeryan, line 3144: `convert_time_to_timestamp` returns None, so `value=convert_time_to_timestamp(...)` does not do anything useful.

Feel free to change the signature of `convert_time_to_timestamp` to `def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime]) -> GrpcTimestamp` if that's what you need. Then the calling template doesn't need to declare a local variable `ts`.

bkeryan (changes requested):

generated/nidaqmx/_grpc_time.py
```diff
@@ -14,7 +14,7 @@
 _YS_PER_FS = 10**9
 
 
-def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTimestamp) -> None:
+def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: Optional[GrpcTimestamp] = GrpcTimestamp()) -> GrpcTimestamp:
```
bkeryan, line 17: Using mutable default arguments like this is a very common Python pitfall.
https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

The correct way to do this is to put `ts: Optional[GrpcTimestamp] = None` in the function signature and put this in the function body:
```
if ts is None:
    ts = GrpcTimestamp()
```

BTW, `typing.Optional[T]` means "this parameter is either T or None". This is entirely orthogonal to optional parameters or default arguments. You can have a required parameter that uses `typing.Optional` like `x: Optional[str]`, or you can have an optional parameter that doesn't use `typing.Optional` like `x: str = ""`.

generated/nidaqmx/_lib_time.py
```diff
@@ -58,6 +58,12 @@ def from_datetime(cls, dt: Union[std_datetime, ht_datetime]) -> AbsoluteTime:
...
         return AbsoluteTime(lsb=lsb, msb=timestamp_1904_epoch)
 
+    def convert_to_local_timezone(self, expected_time_utc):
+        current_time_utc = ht_datetime.now(timezone.utc)
+        local_timezone_offset = current_time_utc.astimezone().utcoffset()
+        local_expected_time = expected_time_utc + local_timezone_offset
+        return local_expected_time
+
     def to_datetime(self, tzinfo: Optional[timezone] = None) -> ht_datetime:
         # First, calculate whole seconds by converting from the 1904 to 1970 epoch.
```
bkeryan, line 68: Why does this function need to do anything with 1970? Adding seconds to the 1904 epoch should work before or after 1970.

Can we simplify the fraction seconds and let hightime do the heavy lifting? Does this lose precision or something?
```
_1904_EPOCH =  ht_datetime(1904, 1, 1)
...
dt = _1904_EPOCH + ht_timedelta(seconds=self.msb) + ht_timedelta(self.lsb / 2**64)
```

bkeryan (changes requested):

generated/nidaqmx/_lib_time.py
bkeryan: Please try:
- Always using `ht_datetime(1904, 1, 1) + ht_timedelta(seconds=self.msb)`.
- Add `ht_timedelta(yoctoseconds=total_yoctoseconds)` to it. I think https://github.com/ni/hightime/blob/master/hightime/_timedelta.py should do the divmod to calculate microseconds, femtoseconds, and remainder_yoctoseconds.

zhindes: Huh, yeah, that's a lot more straight-forward. Let the `ht_datetime` class figure out all the epoch conversions. I don't know why this wouldn't work. I made it a lot harder on myself and basically reimplemented the logic those datetime classes already have for handling epochs. OOPS!

generated/nidaqmx/_grpc_time.py
bkeryan, line 1: I'm sorry, you've been working on this for a while, and I think it's time to take a step back and rethink how to tackle this.

Adding a component test for cfg_time_start_trig uncovered the fact that @zhindes's time conversion functions don't work with dates before 1970 because they rely on POSIX timestamps. Most of the discussion on this PR has been about fixing this aspect of the time conversion functions and simplifying them to make better use of the hightime library.

Your new test for cfg_time_start_trig demonstrates the problem, but it doesn't add enough coverage to validate that the problem is entirely fixed. For example, `AbsoluteTime.from_datetime` still uses `datettime.timestamp()`, which uses POSIX time. (The solution here would be to subtract the 1904 epoch from the timestamp and converting the resulting ht_timedelta into msb and lsb.)

At this point, I think it would make sense to put the cfg_time_start_trig test on hold and create a separate PR that does the following:
- Update all of the lib time conversion functions to stop using POSIX times and do timestamp math with the 1904 epoch.
- Where possible, leverage ht_timedelta to calculate seconds, microseconds, etc. 
- Add test coverage for dates before/around/after 1904 to `test_lib_time.py`.

Then when this is done, your cfg_time_start_trig test PR will be small and low-risk.

generated/nidaqmx/_grpc_time.py
```diff
@@ -30,21 +35,41 @@ def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTime
...
+    return new_datetime
+
+
+def timestamp_to_1904_epoch(timestamp, yoctoseconds):
+    epoch_1904 = ht_datetime(1904, 1, 1)
+    if timestamp < 0:
+        seconds_from_1904 = ht_timedelta(seconds = _BIAS_FROM_1970_EPOCH - abs(timestamp))
+    else:
+        seconds_from_1904 = ht_timedelta(seconds = _BIAS_FROM_1970_EPOCH + abs(timestamp))
+    datetime_1904 = epoch_1904 + seconds_from_1904 + ht_timedelta(yoctoseconds=yoctoseconds)
```
bkeryan, line 65: I **think** msb's sign also applies to lsb: for dates before 1904, you should subtract the fractional seconds.

It would be good to find a test case in NI internal time conversion code (e.g. `nitypest`) confirming this.

generated/nidaqmx/_lib_time.py
```diff
@@ -58,33 +59,30 @@ def from_datetime(cls, dt: Union[std_datetime, ht_datetime]) -> AbsoluteTime:
...
-        yoctosecond = remainder_yoctoseconds
-
-        # Start with UTC
-        dt = ht_datetime.fromtimestamp(timestamp_1970_epoch, timezone.utc)
-        # Add in precision
-        dt = dt.replace(microsecond=microsecond, femtosecond=femtosecond, yoctosecond=yoctosecond)
-        # Then convert to requested timezone
-        return dt.astimezone(tz=tzinfo)
+        datetime_1904 = ht_datetime(1904, 1, 1, tzinfo=timezone.utc)
+        dt = datetime_1904 + ht_timedelta(seconds=self.msb) + ht_timedelta(yoctoseconds=total_yoctoseconds)
```
bkeryan, line 84: I **think** we need to subtract the fractional seconds when msb is negative.

bkeryan (changes requested):

generated/nidaqmx/_grpc_time.py
bkeryan: BTW, gRPC timestamps use the POSIX epoch, but they support a range of years 0001 to 9999: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/timestamp.proto

This suggests that instead of using Python's POSIX time conversions, our gRPC time conversions should do timestamp math with the 1970 epoch.

DeborahOoi96: As per our offline discussion I will split this PR to only contain the unit test changes. The lib time changes will be done in another PR and tracked under another user story. I will mark the two tests that touch default values as xfail and hopefully we can proceed with this PR.

generated/nidaqmx/_lib_time.py
DeborahOoi96: After thinking about it for way too long, I don't think we should subtract the fractional seconds. 
Let's say the user enters 1903:12:31:0:0:0:....1ys. 
We would calculate this way: 1904 + (-1 day) + 1ys

To make it easier I translated this into simple math:
If we wanted to get to 5.01 from a value of 10, we would do it like this:
10 + (-5) + (0.01)
 
We will already use negative value for msb, and we truncate the value so we will get a whole number. Then we need to add back the user required fractional seconds.

Please let me know if there are any gaps in this logic

generated/nidaqmx/_grpc_time.py
DeborahOoi96: The same logic for the lib time comment applies here

tests/component/_task_modules/test_triggers_properties.py
bkeryan: Deleting an attribute in nidaqmx-python is equivalent to resetting an attribute to default value in LabVIEW.
image
After you select "Default Value", the wire terminal for the property value goes away:
image

bkeryan (approved):

Issue #38, opened 2017-10-11
No spacing in raw-read function names


epage:
`readall` and `readinto` rather than `read_all` and `read_into`

https://github.com/ni/nidaqmx-python/blob/master/nidaqmx/_task_modules/in_stream.py#L1898

PR #492, opened 2024-02-08
Fix raw-read Function Names Spacing Issue

charitylxy:

This PR aims to address the issue brought up in #38 
- `readall ` modified to `read_all`
- `readinto` modified to `read_into`

To standardize the raw-read function names spacing.  

- Ran nidaqmx-python code generator and validate the generated file 
- Added and modified regression tests for the related test cases and all tests passed in local machine
Test cases for deprecated function:
image
Test cases for new function name:
image

### DIFF
--- a/generated/nidaqmx/_task_modules/in_stream.py
+++ b/generated/nidaqmx/_task_modules/in_stream.py
@@ -1087,7 +1087,7 @@ class InStream:
             return numpy_array[:number_of_channels * samples_read]
         return numpy_array
     
-    @deprecation.deprecated(deprecated_in="1.2.0", details="Use read_all instead.")
+    @deprecation.deprecated(deprecated_in="1.0.0", removed_in="1.2.0", details="Use read_all instead.")
     def readall(self):
         return self.read_all()
 
@@ -1141,7 +1141,7 @@ class InStream:
         """
         return self.read(number_of_samples_per_channel=READ_ALL_AVAILABLE)
 
-    @deprecation.deprecated(deprecated_in="1.2.0", details="Use read_into instead.")
+    @deprecation.deprecated(deprecated_in="1.2.0", removed_in="1.2.0", details="Use read_into instead.")
     def readinto(self, numpy_array):
         return self.read_into(numpy_array)
 
--- a/generated/nidaqmx/_task_modules/in_stream.py
+++ b/generated/nidaqmx/_task_modules/in_stream.py
@@ -1087,7 +1087,7 @@ class InStream:
             return numpy_array[:number_of_channels * samples_read]
         return numpy_array
 
-    def readall(self):
+    def read_all(self):
         """
         Reads all available raw samples from the task or virtual channels
         you specify.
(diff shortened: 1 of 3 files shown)


generated/nidaqmx/_task_modules/in_stream.py
```diff
@@ -1087,7 +1087,7 @@ def read(self, number_of_samples_per_channel=READ_ALL_AVAILABLE):
             return numpy_array[:number_of_channels * samples_read]
         return numpy_array
 
-    def readall(self):
+    def read_all(self):
```
bkeryan, line 1090: This breaks compatibility. Shouldn't you deprecate the old name first?

That would involve adding a stub like this:
```python
@deprecation.deprecated(deprecated_in="0.10.0", details="Use read_all instead.")
def readall(self):
    return self.read_all()
```

Related: what is the plan for removing deprecated methods and properties? We haven't specified a version in which they will be removed.

Cc: @zhindes

zhindes: I like [PEP 387](https://peps.python.org/pep-0387/#backwards-compatibility-rules) rules:

> Wait for the warning to appear in at least two minor Python versions of the same major version, or one minor version in an older major version (e.g. for a warning in Python 3.10.0, you either wait until at least Python 3.12 or Python 4.0 to make the change).

For this, we would be adding the deprecation in 1.0, so we could remove in 1.2. For existing deprecations, we could remove them now (in 1.0). However, I think for simplicity we bucket all existing deprecations together and let them expire in 1.2.

zhindes: @DeborahOoi96 - do you mind creating an issue in this project to remove our current batch of deprecations in 1.2? (Unless Brad has a better idea!)

bkeryan: Once we decide when the deprecations will be removed, we should also add `removed_in="1.2"` to them: https://pypi.org/project/deprecation/

This will update the deprecation messages in the documentation: https://nidaqmx-python.readthedocs.io/en/latest/ai_channel.html#nidaqmx._task_modules.channels.ai_channel.AIChannel.ai_acceld_b_ref

charitylxy: Got it, I've made the necessary changes and added test cases for the deprecated old function name as well.

zhindes (approved):

generated/nidaqmx/_task_modules/in_stream.py
```diff
@@ -1086,8 +1086,12 @@ def read(self, number_of_samples_per_channel=READ_ALL_AVAILABLE):
         if number_of_channels * samples_read != number_of_samples:
             return numpy_array[:number_of_channels * samples_read]
         return numpy_array
-
+    
+    @deprecation.deprecated(deprecated_in="1.2.0", details="Use read_all instead.")
```
bkeryan, line 1090: These should say deprecated in 1.0.0, removed in 1.2.0

bkeryan (approved):
Approved with suggestions

Issue #392, opened 2023-06-02
Indexing `PhysicalChannelCollection` fails for slices and strings containing a list/range of channels


bkeryan:
Most of the collections in `nidaqmx.system` support slices (e.g. `system.devices[:]` or `system.devices[1:3]`) and lists/ranges of names (e.g. `system.devices["Dev1,Dev2"]). Example:
```
PS D:\dev\nidaqmx-python> poetry run python
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import nidaqmx.system
>>> system = nidaqmx.system.System.local()
>>> system.devices[:]
[Device(name=aoTester), Device(name=bridgeTester), Device(name=cdaqChassisTester), Device(name=Dev1), Device(name=nidaqmxMultithreadingTester1), Device(name=nidaqmxMultithreadingTester2), Device(name=nidaqmxMultithreadingTester3), Device(name=nidaqmxMultithreadingTester4), Device(name=tsChassisTester), Device(name=tsPowerTester1), Device(name=tsPowerTester2), Device(name=tsVoltageTester1)]
>>> system.devices[1:3]
[Device(name=bridgeTester), Device(name=cdaqChassisTester)]
>>> system.devices["Dev1,aoTester"]
[Device(name=Dev1), Device(name=aoTester)]
```

`PhysicalChannelCollection.__getitem__` has code to handle slices, but it incorrectly sets `PhysicalChannel._name` to a list of names, which is the wrong data type:
```
>>> device = system.devices[0]
>>> device.ao_physical_chans
<nidaqmx.system._collections.physical_channel_collection.AOPhysicalChannelCollection object at 0x000002249206F100>
>>> list(device.ao_physical_chans)
[PhysicalChannel(name=aoTester/ao0), PhysicalChannel(name=aoTester/ao1), PhysicalChannel(name=aoTester/ao2), PhysicalChannel(name=aoTester/ao3), PhysicalChannel(name=aoTester/ao4), PhysicalChannel(name=aoTester/ao5), PhysicalChannel(name=aoTester/ao6), PhysicalChannel(name=aoTester/ao7)]
>>> device.ao_physical_chans[:]
[...]

PR #494, opened 2024-02-14
 Add Digital Stream Reader tests

zhindes:

This ensures we have a test per method for each of the Digital stream readers classes. The existing and new tests all have channel-specific data validation by taking advantage of the behavior of simulated DAQ digital data. This ensures we don't inadvertently swap channels within the Python or underlying C bindings.

This will give us confidence that we haven't broken anything as we refactor parts of the Python and driver stack.

```
$ poetry run ni-python-styleguide fix tests && poetry run ni-python-styleguide lint tests && poetry run pytest -rsP -k test___digital
=============================== test session starts ===============================
platform win32 -- Python 3.10.10, pytest-7.4.2, pluggy-1.3.0
rootdir: C:\dev\nidaqmx-python
configfile: pyproject.toml
testpaths: tests
plugins: cov-4.1.0, mock-3.11.1
collected 1641 items / 1585 deselected / 56 selected

tests\component\test_stream_readers.py ..................................... [ 66%]
...................                                                          [100%]

===================================== PASSES ======================================
======================= 56 passed, 1585 deselected in 2.44s =======================
```

### DIFF
--- a/tests/component/test_stream_readers.py
+++ b/tests/component/test_stream_readers.py
@@ -6,6 +6,7 @@ import numpy
 import pytest
 
 import nidaqmx
+from nidaqmx.constants import LineGrouping
 from nidaqmx.stream_readers import (
     AnalogMultiChannelReader,
     AnalogSingleChannelReader,
@@ -14,6 +15,7 @@ from nidaqmx.stream_readers import (
     PowerMultiChannelReader,
     PowerSingleChannelReader,
 )
+from nidaqmx.utils import flatten_channel_string
 
 
 # Simulated DAQ voltage data is a noisy sinewave within the range of the minimum and maximum values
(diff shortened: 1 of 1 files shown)


tests/component/test_stream_readers.py
```diff
@@ -545,3 +547,591 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+        flatten_channel_string(sim_6363_device.di_lines[:32].name),
+        line_grouping=LineGrouping.CHAN_PER_LINE,
+    )
+    return task
+
+
+def _get_expected_digital_data_for_sample(num_lines: int, sample_number: int) -> int:
+    # Simulated digital signals "count" from 0 in binary within each group of 8 lines.
+    result = sample_number
+    for _ in range(num_lines // 8):
```
bkeryan, line 661: `//` is floor division. Should this use `(num_lines + 7) // 8` to round up? I would expect this to affect the tests that use line0:3.

tests/component/test_stream_readers.py
```diff
@@ -545,3 +547,591 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+    result = []
+    # again, port data is strange. it is line-major with the same odd masking/port format as above
+    for line in range(num_lines):
+        result.append(
+            [datum & (1 << line) for datum in _get_expected_digital_data(num_lines, num_samples)]
+        )
+    return result
+
+
+def _bool_array_to_int(bool_array: numpy.ndarray) -> int:
```
bkeryan, line 691: `numpy.typing.NDArray[numpy.bool_]`

tests/component/test_stream_readers.py
```diff
@@ -545,3 +547,591 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+    return result
+
+
+def test___digital_single_channel_reader___read_one_sample_one_line___returns_valid_samples(
+    di_single_line_task: nidaqmx.Task,
+) -> None:
+    reader = nidaqmx.stream_readers.DigitalSingleChannelReader(di_single_line_task.in_stream)
+    num_lines = di_single_line_task.channels.di_num_lines
+    samples_to_read = 10
+    data = [reader.read_one_sample_one_line() for _ in range(samples_to_read)]
```
bkeryan, line 705: arrange, act, assert

This applies to every test case, they should all be split up into 3 sections (unless they omit a section)
```suggestion
    samples_to_read = 10
    
    data = [reader.read_one_sample_one_line() for _ in range(samples_to_read)]
```

tests/component/test_stream_readers.py
```diff
@@ -545,3 +547,591 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+    reader = nidaqmx.stream_readers.DigitalMultiChannelReader(di_single_line_task.in_stream)
+    num_lines = di_single_line_task.channels.di_num_lines
+    samples_to_read = 10
+    data = numpy.full(num_lines, False, dtype=numpy.bool_)
+
+    for sample_number in range(samples_to_read):
+        reader.read_one_sample_one_line(data)
+        assert _bool_array_to_int(data) == _get_expected_digital_data_for_sample(
+            num_lines, sample_number
+        )
```
bkeryan, line 722: arrange, act, assert

I like how some of these tests cases accumulate data in "act" and validate it in "assert". Can you do the same thing with test cases that read into a preallocated array? This avoids mixing calls to the SUT with validation code, and also when it fails it prints everything, not just the first sample that didn't match.

```
def read_and_copy(read_func, array):
    read_func(array)
    return array.copy()
```

```
sample = numpy.full(num_lines, False, dtype=numpy.bool_)

data = [read_and_copy(reader.read_one_sample_one_line, sample) for _ in range(samples_to_read))]

assert [_bool_array_to_int(sample) for sample in data] == [_get_expected_digital_data_for_sample(num_lines, sample_number) for sample_number in range(samples_to_read)]
```

The assert could also benefit from more helper functions.

tests/component/test_stream_readers.py
```diff
@@ -545,3 +547,591 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+
+
+def test___digital_single_channel_reader___read_one_sample_multi_line_with_wrong_dtype___raises_error_with_correct_dtype(
+    di_single_channel_multi_line_task: nidaqmx.Task,
+) -> None:
+    reader = nidaqmx.stream_readers.DigitalSingleChannelReader(
+        di_single_channel_multi_line_task.in_stream
```
bkeryan, line 759: [style] `from nidaqmx.stream_readers import DigitalSingleChannelReader, DigitalMultiChannelReader` would help some of these fit on one line.

tests/component/test_stream_readers.py
```diff
@@ -545,3 +547,591 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+
+@pytest.fixture
+def di_multi_channel_port_byte_task(
+    task: nidaqmx.Task, sim_6363_device: nidaqmx.system.Device
+) -> nidaqmx.Task:
+    # 6363 port 1/2 have 8 lines each, but we need to skip port 0, which has 32 lines
+    task.di_channels.add_di_chan(
+        flatten_channel_string(sim_6363_device.di_lines[32:40].name),
+        line_grouping=LineGrouping.CHAN_PER_LINE,
+    )
```
bkeryan, line 602: Using port reads with CHAN_PER_LINE is kind of weird, as you noticed. It's less weird to put each port in a separate channel. Then the port read returns [0xFF, 0xFF] instead of [0x01, 0x02, ..., 0x80, 0x01, 0x02, ..., 0x80].

```
task.di_channels.add_di_chan(
    sim_6363_device.di_ports[1].name,
    line_grouping=LineGrouping.CHAN_FOR_ALL_LINES,
)
task.di_channels.add_di_chan(
    sim_6363_device.di_ports[2].name,
    line_grouping=LineGrouping.CHAN_FOR_ALL_LINES,
)
```

tests/component/test_stream_readers.py
```diff
@@ -545,3 +547,591 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+    # 6363 port 0 has 32 lines
+    task.di_channels.add_di_chan(
+        flatten_channel_string(sim_6363_device.di_lines[:32].name),
+        line_grouping=LineGrouping.CHAN_PER_LINE,
+    )
+    return task
+
+
+def _get_expected_digital_data_for_sample(num_lines: int, sample_number: int) -> int:
+    # Simulated digital signals "count" from 0 in binary within each group of 8 lines.
```
bkeryan, line 659: Reading 10 samples counts from 0b00000000 to 0b00001010, so it only exercises the lower 4 bits.

tests/component/test_stream_readers.py
```diff
@@ -545,3 +547,591 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+    for bit in reversed(bool_array):
+        result = (result << 1) | bit
+    return result
+
+
+def test___digital_single_channel_reader___read_one_sample_one_line___returns_valid_samples(
+    di_single_line_task: nidaqmx.Task,
+) -> None:
+    reader = nidaqmx.stream_readers.DigitalSingleChannelReader(di_single_line_task.in_stream)
+    num_lines = di_single_line_task.channels.di_num_lines
```
bkeryan, line 703: This looks like it's getting the total number of lines, but `task.channels.di_num_lines` returns the number of lines per channel and errors if they're not the same. It happens to work because there is only one channel.

I see the multi-channel multi-line test cases are setting `num_lines = task.number_of_channels`, which makes a different assumption: that each channel contains one line.

I think this is kind of misleading, but more importantly it highlights the lack of coverage of less trivial multi-line multi-channel use cases. DAQmx lets you group an arbitrary number of lines into each channel. It's easy to go overboard here, but perhaps we could at least test the most obvious case, chan0="Dev1/port0" (32) chan1="Dev1/port1" (8) chan2="Dev1/port2" (8). 

It's ok to split this into a separate PR.

bkeryan (changes requested):

tests/component/test_stream_readers.py
zhindes: The starting result (outside the loop) works fine for the first 8 lines. When more than 8 lines, we gotta start shifting in new values. BUT.... I think your suggestion is more obvious and cleaner code, I like it!

tests/component/test_stream_readers.py
zhindes: Holy hell this is way easier. Why did I make it so hard on myself :D

tests/component/test_stream_readers.py
zhindes: This test case naturally appeared out of my refactor to move to ports instead of lines for multi-channel tests. `di_multi_channel_port_uint32_task` is your exact example.

tests/component/test_stream_readers.py
```diff
@@ -545,3 +550,598 @@ def test___power_binary_reader___read_many_sample_with_wrong_dtype___raises_erro
...
+    result = 0
+    # Simulated data is little-endian
+    for bit in reversed(bool_array):
+        result = (result << 1) | bit
+    return result
+
+
+def _read_and_copy(
+    read_func: Callable[[numpy.typing.NDArray], None], array: numpy.typing.NDArray
+) -> numpy.ndarray:
```
bkeryan, line 721: FYI, the type checker doesn't know all three of these ndarrays have the same dtype. Here's how to tell it that (I think):

```
_DType = TypeVar("_DType", bound=numpy.typing.DTypeLike)

def _read_and_copy(
    read_func: Callable[[numpy.typing.NDArray[_DType]], None], array: numpy.typing.NDArray[_DType]
) -> numpy.typing.NDArray[_DType]:
```

Python 3.12 makes this less confusing by removing the separate TypeVar declaration, but we can't use this syntax until 3.11 is EOLed:
```
def _read_and_copy[DType: numpy.typing.DTypeLike](
    read_func: Callable[[numpy.typing.NDArray[DType]], None], array: numpy.typing.NDArray[DType]
) -> numpy.typing.NDArray[DType]:
```

bkeryan (approved):

tests/component/test_stream_readers.py
zhindes: Neat! That matches what I see in the documentation, and it seems to work.

tests/component/test_stream_readers.py
```diff
@@ -562,49 +562,100 @@ def di_single_channel_multi_line_task(
     task: nidaqmx.Task, sim_6363_device: nidaqmx.system.Device
 ) -> nidaqmx.Task:
     task.di_channels.add_di_chan(
-        flatten_channel_string(sim_6363_device.di_lines[:2].name),
+        flatten_channel_string(sim_6363_device.di_lines[:4].name),
```
bkeryan: FYI, I tried using this technique in another test and it doesn't really work right. It relies on the GIGO behavior described in https://github.com/ni/nidaqmx-python/issues/392

`di_lines[:4]` returns a single `DIChannel` object. Its `name` property should be a flattened channel string, but it's actually a Python list of channel names.

PR #497, opened 2024-02-16
Add Counter Stream Reader tests 

zhindes:

This ensures we have a test per method within the Counter stream readers class. These tests use an onboard timebase of a real X Series to validate we get correct data from either the Python or underlying C bindings.

This will give us confidence that we haven't broken anything as we refactor parts of the Python and driver stack.

```
$ poetry run ni-python-styleguide fix tests && poetry run ni-python-styleguide lint tests && poetry run pytest -rsP -k test___counter
=============================== test session starts ===============================
platform win32 -- Python 3.10.10, pytest-7.4.2, pluggy-1.3.0
rootdir: C:\dev\nidaqmx-python
configfile: pyproject.toml
testpaths: tests
plugins: cov-4.1.0, mock-3.11.1
collected 1653 items / 1611 deselected / 42 selected

tests\acceptance\test_stream_readers.py .................................... [ 85%]
......                                                                       [100%]

===================================== PASSES ======================================
======================= 42 passed, 1611 deselected in 6.17s =======================
```

### DIFF
--- a/tests/acceptance/test_stream_readers.py
+++ b/tests/acceptance/test_stream_readers.py
@@ -285,7 +285,7 @@ def test___counter_reader___read_many_sample_pulse_time___returns_valid_samples(
     ],
 )
 def test___counter_reader___read_many_sample_pulse_time_with_wrong_dtype___raises_error_with_correct_dtype(
-    ci_pulse_time_task: nidaqmx.Task, high_time_dtype: numpy.generic, low_time_dtype: numpy.generic
+    ci_pulse_time_task: nidaqmx.Task, high_time_dtype: numpy.typing.DTypeLike, low_time_dtype: numpy.typing.DTypeLike
 ) -> None:
     reader = CounterReader(ci_pulse_time_task.in_stream)
     samples_to_read = 10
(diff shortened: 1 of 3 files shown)


tests/acceptance/test_stream_readers.py
```diff
@@ -0,0 +1,357 @@
...
+    "high_ticks_dtype, high_ticks_default, low_ticks_dtype, low_ticks_default",
+    [
+        (numpy.float64, math.inf, numpy.uint32, numpy.iinfo(numpy.uint32).min),
+        (numpy.uint32, numpy.iinfo(numpy.uint32).min, numpy.float64, math.inf),
+        (numpy.float64, math.inf, numpy.float64, math.inf),
+    ],
+)
+def test___counter_reader___read_many_sample_pulse_ticks_with_wrong_dtype___raises_error_with_correct_dtype(
+    ci_pulse_ticks_task: nidaqmx.Task,
+    high_ticks_dtype: numpy.generic,
```
bkeryan, line 342: I think the dtypes should be `numpy.typing.DTypeLike`

tests/acceptance/test_stream_readers.py
bkeryan, line 1: Buffered pulse input supports multi-channel. Does the Python API support that? Do you have to reshape the NumPy array yourself?

In LV, there is a `Counter 2D DBL NChan NSamp` poly VI instance.
In C, the `numSampsPerChan` and `arraySizeInSamps` parameters are separate so that you can specify an array with more than one channel.

This seems like a missing API feature, so it's outside the scope of this PR.

bkeryan (changes requested):

tests/acceptance/test_stream_readers.py
zhindes: Oh wow, I had forgotten about that. We also added multi-channel double reads for 9361 to do multiple duty-cycle reads in the same task. You're right that Python doesn't support multi-channel counters quite right. I think `task.read` probably works already for doubles, but not pulses. There isn't a `ChannelReader` variant that looks appropriate for any of that, though. #498 498

bkeryan (approved):

PR #504, opened 2024-02-20
Update libtime to accept time before 1970

DeborahOoi96:

Modify _lib_time.py and _grpc_time.py to accept time before 1970.
Add unit test coverage to test for times before 1904 and on 1904.

Adding [a component test for cfg_time_start_trig](https://github.com/ni/nidaqmx-python/pull/488) uncovered the fact that @zhindes's time conversion functions don't work with dates before 1970 because they rely on POSIX timestamps. 
We want to fix this aspect of the time conversion functions and simplify them to make better use of the hightime library. 
We also want to add enough coverage to validate that the problem is entirely fixed. For example, AbsoluteTime.from_datetime still uses datetime.timestamp(), which uses POSIX time.

Built successfully. All test passed including newly added tests
image

### DIFF
--- a/generated/nidaqmx/_time.py
+++ b/generated/nidaqmx/_time.py
@@ -18,8 +18,8 @@ def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datet
     if tzinfo is None:
         tzinfo = get_localzone()
 
-    # use pytz.tzinfo.BaseTzInfo here to account for daylight savings
-    if hasattr(tzinfo, "zone") or isinstance(tzinfo, ZoneInfo):
+    # use ZoneInfo here to account for daylight savings
+    if isinstance(tzinfo, ZoneInfo):
         localized_time = expected_time_utc.replace(tzinfo=tzinfo)
         desired_expected_time = tzinfo.fromutc(localized_time)
         return(desired_expected_time)
--- a/generated/nidaqmx/_time.py
+++ b/generated/nidaqmx/_time.py
@@ -6,7 +6,10 @@ from datetime import tzinfo as dt_tzinfo
 from datetime import datetime as std_datetime
 from hightime import datetime as ht_datetime
 from typing import Optional, Union
-from backports.zoneinfo import ZoneInfo
+try:
+    import zoneinfo
+except ImportError:
+    from backports import zoneinfo
 
 # theoretically the same as astimezone(), but with support for dates before 1970
 def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[dt_tzinfo] = None) -> Union[std_datetime, ht_datetime]:
@@ -19,7 +22,7 @@ def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datet
         tzinfo = get_localzone()
 
     # use ZoneInfo here to account for daylight savings
-    if isinstance(tzinfo, ZoneInfo):
+    if isinstance(tzinfo, zoneinfo.ZoneInfo):
         localized_time = expected_time_utc.replace(tzinfo=tzinfo)
         desired_expected_time = tzinfo.fromutc(localized_time)
         return(desired_expected_time)
--- a/generated/nidaqmx/_time.py
+++ b/generated/nidaqmx/_time.py
@@ -1,46 +1,33 @@
 from __future__ import annotations
 
-import pytz
 from tzlocal import get_localzone
 from datetime import timezone
+from datetime import tzinfo as dt_tzinfo
 from datetime import datetime as std_datetime
 from hightime import datetime as ht_datetime
 from typing import Optional, Union
+from zoneinfo import ZoneInfo
 
 # theoretically the same as astimezone(), but with support for dates before 1970
-def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[timezone] = None):
+def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[dt_tzinfo] = None):
     # if timezone matches, no need to do conversion
     if expected_time_utc.tzinfo is tzinfo:
         return expected_time_utc
 
     # if timezone is not defined, use system timezone
     if tzinfo is None:
-        local_timezone = get_localzone()
-        tzinfo = pytz.timezone(str(local_timezone))
+        tzinfo = get_localzone()
 
-    # use pytz here to account for daylight savings
-    if isinstance(tzinfo, pytz.tzinfo.BaseTzInfo):
-        target_tz = pytz.timezone(str(tzinfo))
-        # we need to make expected_time_utc naive in order to use localize
-        expected_time_utc = expected_time_utc.replace(tzinfo=None)
-        localized_time = target_tz.localize(expected_time_utc)
-        desired_expected_time = target_tz.fromutc(localized_time)
+    # use pytz.tzinfo.BaseTzInfo here to account for daylight savings
+    if hasattr(tzinfo, "zone") or isinstance(tzinfo, ZoneInfo):
+        localized_time = expected_time_utc.replace(tzinfo=tzinfo)
+        desired_expected_time = tzinfo.fromutc(localized_time)
         return(desired_expected_time)
 
     # if the tzinfo passed in is a timedelta function, then we don't need to consider daylight savings
-    elif isinstance(tzinfo, timezone) and tzinfo.utcoffset(None) is not None:
+    elif isinstance(tzinfo, dt_tzinfo) and tzinfo.utcoffset(None) is not None:
         current_time_utc = ht_datetime.now(timezone.utc)
         desired_timezone_offset = current_time_utc.astimezone(tz=tzinfo).utcoffset()
         desired_expected_time = expected_time_utc + desired_timezone_offset
-        new_datetime = ht_datetime(
-            desired_expected_time.year,
-            desired_expected_time.month,
-            desired_expected_time.day,
-            desired_expected_time.hour,
-            desired_expected_time.minute,
-            desired_expected_time.second,
-            desired_expected_time.microsecond,
-            desired_expected_time.femtosecond,
-            desired_expected_time.yoctosecond,
-            tzinfo = tzinfo)
+        new_datetime = desired_expected_time.replace(tzinfo=tzinfo)
         return new_datetime
--- a/generated/nidaqmx/_time.py
+++ b/generated/nidaqmx/_time.py
@@ -1,0 +1,46 @@
+from __future__ import annotations
+
+import pytz
+from tzlocal import get_localzone
+from datetime import timezone
+from datetime import datetime as std_datetime
+from hightime import datetime as ht_datetime
+from typing import Optional, Union
+
+# theoretically the same as astimezone(), but with support for dates before 1970
+def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[timezone] = None):
+    # if timezone matches, no need to do conversion
+    if expected_time_utc.tzinfo is tzinfo:
+        return expected_time_utc
+
+    # if timezone is not defined, use system timezone
+    if tzinfo is None:
+        local_timezone = get_localzone()
+        tzinfo = pytz.timezone(str(local_timezone))
+
+    # use pytz here to account for daylight savings
+    if isinstance(tzinfo, pytz.tzinfo.BaseTzInfo):
+        target_tz = pytz.timezone(str(tzinfo))
+        # we need to make expected_time_utc naive in order to use localize
+        expected_time_utc = expected_time_utc.replace(tzinfo=None)
+        localized_time = target_tz.localize(expected_time_utc)
+        desired_expected_time = target_tz.fromutc(localized_time)
+        return(desired_expected_time)
+
+    # if the tzinfo passed in is a timedelta function, then we don't need to consider daylight savings
+    elif isinstance(tzinfo, timezone) and tzinfo.utcoffset(None) is not None:
+        current_time_utc = ht_datetime.now(timezone.utc)
+        desired_timezone_offset = current_time_utc.astimezone(tz=tzinfo).utcoffset()
+        desired_expected_time = expected_time_utc + desired_timezone_offset
+        new_datetime = ht_datetime(
+            desired_expected_time.year,
+            desired_expected_time.month,
+            desired_expected_time.day,
+            desired_expected_time.hour,
+            desired_expected_time.minute,
+            desired_expected_time.second,
+            desired_expected_time.microsecond,
+            desired_expected_time.femtosecond,
+            desired_expected_time.yoctosecond,
+            tzinfo = tzinfo)
+        return new_datetime
--- a/generated/nidaqmx/_time.py
+++ b/generated/nidaqmx/_time.py
@@ -9,7 +9,7 @@ from typing import Optional, Union
 from zoneinfo import ZoneInfo
 
 # theoretically the same as astimezone(), but with support for dates before 1970
-def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[dt_tzinfo] = None):
+def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[dt_tzinfo] = None) -> Union[std_datetime, ht_datetime]:
     # if timezone matches, no need to do conversion
     if expected_time_utc.tzinfo is tzinfo:
         return expected_time_utc
@@ -25,9 +25,13 @@ def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datet
         return(desired_expected_time)
 
     # if the tzinfo passed in is a timedelta function, then we don't need to consider daylight savings
-    elif isinstance(tzinfo, dt_tzinfo) and tzinfo.utcoffset(None) is not None:
+    elif tzinfo.utcoffset(None) is not None:
         current_time_utc = ht_datetime.now(timezone.utc)
         desired_timezone_offset = current_time_utc.astimezone(tz=tzinfo).utcoffset()
         desired_expected_time = expected_time_utc + desired_timezone_offset
         new_datetime = desired_expected_time.replace(tzinfo=tzinfo)
         return new_datetime
+
+    # if the tzinfo passed in is none of the above, fall back to original astimezone()
+    else:
+        return expected_time_utc.astimezone(tzinfo)
--- a/generated/nidaqmx/_time.py
+++ b/generated/nidaqmx/_time.py
@@ -6,7 +6,7 @@ from datetime import tzinfo as dt_tzinfo
 from datetime import datetime as std_datetime
 from hightime import datetime as ht_datetime
 from typing import Optional, Union
-from zoneinfo import ZoneInfo
+from backports.zoneinfo import ZoneInfo
 
 # theoretically the same as astimezone(), but with support for dates before 1970
 def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[dt_tzinfo] = None) -> Union[std_datetime, ht_datetime]:
--- a/src/handwritten/_grpc_time.py
+++ b/src/handwritten/_grpc_time.py
@@ -16,10 +16,30 @@ _YS_PER_US = 10**18
 _YS_PER_NS = 10**15
 _YS_PER_FS = 10**9
 
+def convert_to_desired_timezone(expected_time_utc, tzinfo):
+    current_time_utc = ht_datetime.now(timezone.utc)
+    desired_timezone_offset = current_time_utc.astimezone(tz=tzinfo).utcoffset()
+    desired_expected_time = expected_time_utc + desired_timezone_offset
+    new_datetime = ht_datetime(
+        desired_expected_time.year,
+        desired_expected_time.month,
+        desired_expected_time.day,
+        desired_expected_time.hour,
+        desired_expected_time.minute,
+        desired_expected_time.second,
+        desired_expected_time.microsecond,
+        desired_expected_time.femtosecond,
+        desired_expected_time.yoctosecond,
+        tzinfo = tzinfo)
+    return new_datetime
+
 
 def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: Optional[GrpcTimestamp] = None) -> GrpcTimestamp:
-    utc_dt = dt.astimezone(tz=timezone.utc)
-    seconds = int(utc_dt.timestamp())
+    epoch_1970 = ht_datetime(1970, 1, 1, tzinfo=timezone.utc)
+    if dt < epoch_1970:
+        seconds_since_1970 = -int(abs(dt - epoch_1970).total_seconds())
+    else:
+        seconds_since_1970 = int(abs(dt - epoch_1970).total_seconds())
     if ts is None:
         ts = GrpcTimestamp()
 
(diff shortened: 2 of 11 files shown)


tests/unit/test_lib_time.py
```diff
@@ -102,6 +154,45 @@ def test___datetime_with_microseconds___convert_to_timestamp___is_reversible(
     assert roundtrip_dt_microsecond == microsecond
 
 
+@pytest.mark.parametrize(
+    "base_dt, microsecond, subseconds",
+    [
+        (JAN_01_1850_DATETIME, 0, 0),
+        (JAN_01_1850_DATETIME, 1, 0x10C6F7A0B5EE),
+        (JAN_01_1850_DATETIME, 250000, 0x4000000000000000),
+        (JAN_01_1850_DATETIME, 500000, 0x8000000000000000),
```
bkeryan, line 163: These subseconds values seem to be based on the premise that the NI-BTF fractional seconds always count forwards (like Protobuf timestamps), even when the whole seconds are negative. Unfortunately, https://www.ni.com/docs/en-US/bundle/labwindows-cvi/page/cvi/libref/ni-btf.htm doesn't say whether the positive fractions of a second follow the sign of the whole seconds.

Do we have any documentation or tests confirming that this is the case?

@zhindes

tests/unit/test_lib_time.py
```diff
@@ -102,6 +154,45 @@ def test___datetime_with_microseconds___convert_to_timestamp___is_reversible(
...
+    if microsecond:
+        assert to_ts.msb == JAN_01_1850_LIB.msb + 1
+    else:
+        assert to_ts.msb == JAN_01_1850_LIB.msb
+    assert to_ts.lsb == subseconds
+    # comparison is tricky since imprecision in the conversion to NI-BTF are
+    # caught by the higher precision values in hightime, so we round here.
+    roundtrip_dt_microsecond = roundtrip_dt.microsecond
+    if roundtrip_dt.femtosecond > LibTimestamp.MAX_FS / 2:
+        roundtrip_dt_microsecond += 1
```
bkeryan, line 191: Can you use `pytest.approx` to do the rounding?

src/handwritten/_grpc_time.py
```diff
@@ -13,10 +16,32 @@
...
+        desired_expected_time.yoctosecond,
+        tzinfo = tzinfo)
+    return new_datetime
 
-def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTimestamp) -> None:
-    utc_dt = dt.astimezone(tz=timezone.utc)
-    seconds = int(utc_dt.timestamp())
+
+def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: Optional[GrpcTimestamp] = None) -> GrpcTimestamp:
+    epoch_1970 = ht_datetime(1970, 1, 1, tzinfo=timezone.utc)
```
bkeryan, line 38: I don't think we have to construct this epoch variable every time. Please extract it into a constant.

src/handwritten/_grpc_time.py
```diff
@@ -13,10 +16,32 @@
...
+    if dt < epoch_1970:
+        seconds_since_1970 = -int(abs(dt - epoch_1970).total_seconds())
+    else:
+        seconds_since_1970 = int(abs(dt - epoch_1970).total_seconds())
```
bkeryan, line 42: Are you sure the conditional is necessary? The sign appears to be correct without it.

```
>>> (datetime.datetime(1969, 1, 1) - datetime.datetime(1970, 1, 1)).total_seconds()
-31536000.0
>>> (datetime.datetime(1971, 1, 1) - datetime.datetime(1970, 1, 1)).total_seconds()
31536000.0
```

src/handwritten/_grpc_time.py
```diff
@@ -27,24 +52,21 @@ def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: GrpcTime
...
     else:
-        nanos = utc_dt.microsecond * _NS_PER_US
+        nanos = dt.microsecond * _NS_PER_US
 
-    ts.FromNanoseconds(seconds * _NS_PER_S + nanos)
+    ts.FromNanoseconds(seconds_since_1970 * _NS_PER_S + nanos)
+    return ts
 
 
+def timestamp_to_1970_epoch(timestamp, yoctoseconds):
```
bkeryan, line 61: If you extract the epoch into a constant, this becomes a one-liner, so consider inlining it into `convert_timestamp_to_time`.

If not: `_` prefix, type hints

src/handwritten/_grpc_time.py
```diff
@@ -13,10 +16,32 @@
 _YS_PER_NS = 10**15
 _YS_PER_FS = 10**9
 
+def convert_to_desired_timezone(expected_time_utc, tzinfo):
```
bkeryan, line 19: Add a `_` prefix to helper functions to keep them private. 

Please write type hints for helper functions.

src/handwritten/_lib_time.py
```diff
@@ -29,21 +30,29 @@ class AbsoluteTime(ctypes.Structure):
     MAX_FS = 10**9
     MAX_YS = 10**9
 
+    def convert_to_desired_timezone(self, expected_time_utc, tzinfo):
```
bkeryan, line 33: ditto, re: helpers

src/handwritten/_lib_time.py
```diff
@@ -29,21 +30,29 @@ class AbsoluteTime(ctypes.Structure):
...
     @classmethod
     def from_datetime(cls, dt: Union[std_datetime, ht_datetime]) -> AbsoluteTime:
-        utc_dt = dt.astimezone(tz=timezone.utc)
-
-        # First, calculate whole seconds by converting from the 1970 to 1904 epoch.
-        timestamp_1970_epoch = utc_dt.timestamp()
-        was_negative = timestamp_1970_epoch < 0
-        timestamp_1904_epoch = int(timestamp_1970_epoch + AbsoluteTime._BIAS_FROM_1970_EPOCH)
+        epoch_1904 = ht_datetime(1904, 1, 1, tzinfo=timezone.utc)
+        seconds_since_1904 = int(abs(dt - epoch_1904) / ht_timedelta(seconds=1))
```
bkeryan, line 53: ditto, re: epoch constant, sign

bkeryan, line 53: Why does the lib version divide by 1 second and the grpc version use `.total_seconds()`?

bkeryan (changes requested):

tests/unit/test_lib_time.py
bkeryan: BTW, I am 90% sure that for CVI time, lsb implicitly follows the sign of msb. Why? Because these types are defined as 128-bit fixed point numbers with 64-bit int part and 64-bit fractional part.

If we don't have any docs/tests confirming this, it would be helpful to add a test case for this to the internal "nitypest" component.

zhindes: It would be pretty insane for a generic fixed point type to not have the sign behavior you describe...

tests/unit/test_grpc_time.py
```diff
@@ -113,6 +178,40 @@ def test___datetime_with_microseconds___convert_to_timestamp___is_reversible(
...
+    to_ts = GrpcTimestamp()
+    grpc_time.convert_time_to_timestamp(from_dt, to_ts)
+    roundtrip_dt = grpc_time.convert_timestamp_to_time(to_ts, tzinfo=timezone.utc)
+
+    if microsecond:
+        assert to_ts.seconds == JAN_01_1850_TIMESTAMP_1970_EPOCH + 1
+    else:
+        assert to_ts.seconds == JAN_01_1850_TIMESTAMP_1970_EPOCH
+    assert to_ts.nanos == nanoseconds
+    assert roundtrip_dt.microsecond == microsecond
```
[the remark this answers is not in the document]
DeborahOoi96: You are right. I cracked my head on this. The thing that threw me off was the calculation of `seconds_since_1970`.

`seconds_since_1970 = int((dt - _EPOCH_1970) / ht_timedelta(seconds=1))` 

If the user set: Jan 1, 1850 + 1us, according to the logic in `convert_time_to_timestamp()` the timestamp returned will be -3786825599. I ran it through an online compiler to be sure.
image

But if the user sets Jan 1, 1850 + 1femtosec (assuming it's hightime), the timestamp returned will be -3786825600. It's weird and I can't find any flaw in the hightime sub func definition, but that is what's returned.
image
Maybe the number is too small to be reflected in the seconds? Or maybe the changes somehow were not propogated up correctly to seconds. 

Either way, I added a workaround for this special case in grpc_time.

tests/unit/test_lib_time.py
DeborahOoi96: I'm just going off from what @zhindes documented in AbsoluteTime which says that lsb is `positive fractions (2^-64) of a second`, and is also mentioned in the NI_BTF documentation
image

which I assume is referred to from here
https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nicaistub/source/nicaistub/nicaistub_typedefs.h&version=GCde6458e0b93bf6d6c064623bb74cef4ab9f5cc84&line=80&lineEnd=84&lineStartColumn=1&lineEndColumn=11&lineStyle=plain&_a=contents
I also noticed that cvi has it as an [unsigned int](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/cvi_product/cvi/Components/cvi_Headers/source/cvitdms.h&version=GCde6458e0b93bf6d6c064623bb74cef4ab9f5cc84&line=46&lineEnd=46&lineStartColumn=26&lineEndColumn=42&lineStyle=plain&_a=contents) so maybe this proves that it's always positive?

src/handwritten/_lib_time.py
```diff
@@ -29,21 +30,30 @@ class AbsoluteTime(ctypes.Structure):
     MAX_FS = 10**9
     MAX_YS = 10**9
 
+    _EPOCH_1904 = ht_datetime(1904, 1, 1, tzinfo=timezone.utc)
+
+    def _convert_to_desired_timezone(self, expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[timezone] = None):
```
zhindes, line 35: I think this method is identical in grpc and lib, so you could move it to a new file, `_time.py`.

src/handwritten/_lib_time.py
```diff
@@ -29,21 +30,30 @@ class AbsoluteTime(ctypes.Structure):
...
+            desired_expected_time.month,
+            desired_expected_time.day,
+            desired_expected_time.hour,
+            desired_expected_time.minute,
+            desired_expected_time.second,
+            desired_expected_time.microsecond,
+            desired_expected_time.femtosecond,
+            desired_expected_time.yoctosecond,
+            tzinfo = tzinfo)
+        return new_datetime
```
zhindes, line 50: This looks like a bespoke implementation of [datetime.astimezone](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone) (hightime has the same method). That is dangerous...

src/handwritten/_lib_time.py
zhindes: Except maybe it isn't required at all, see next comment.

zhindes (changes requested):

src/handwritten/_lib_time.py
zhindes: Ahhh, ok. Add a comment to that effect :)

tests/unit/test_grpc_time.py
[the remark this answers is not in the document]
DeborahOoi96: > You are right. I cracked my head on this. The thing that threw me off was the calculation of `seconds_since_1970`.
> 
> 
> 
> `seconds_since_1970 = int((dt - _EPOCH_1970) / ht_timedelta(seconds=1))` 
> 
> 
> 
> If the user set: Jan 1, 1850 + 1us, according to the logic in `convert_time_to_timestamp()` the timestamp returned will be -3786825599. I ran it through an online compiler to be sure.
> 
> image
> 
> 
> 
> But if the user sets Jan 1, 1850 + 1femtosec (assuming it's hightime), the timestamp returned will be -3786825600. It's weird and I can't find any flaw in the hightime sub func definition, but that is what's returned.
> 
> image
> 
> Maybe the number is too small to be reflected in the seconds? Or maybe the changes somehow were not propogated up correctly to seconds. 
> 
> 
> 
> Either way, I added a workaround for this special case in grpc_time.
> 
> 
> 
> 

@zhindes fyi

src/handwritten/_grpc_time.py
```diff
@@ -13,10 +16,31 @@
...
+    return new_datetime
+
 
 def convert_time_to_timestamp(dt: Union[std_datetime, ht_datetime], ts: Optional[GrpcTimestamp] = None) -> GrpcTimestamp:
-    utc_dt = dt.astimezone(tz=timezone.utc)
-    seconds = int(utc_dt.timestamp())
+    seconds_since_1970 = int((dt - _EPOCH_1970) / ht_timedelta(seconds=1))
+    # We need to add one more negative second if applicable to compensate for a non-zero microsecond.
+    if dt.microsecond and (dt < _EPOCH_1970):
+        seconds_since_1970 -=1
```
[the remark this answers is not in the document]
DeborahOoi96: I have tagged you in Brad's comment where the discussion for this took place. Please refer to that for further (complicated) elaboration on why I added this in grpc_time

tests/unit/test_lib_time.py
bkeryan: The lsb is always positive because the sign bit is in the msb. This does not tell you whether the lsb counts forwards or backwards in time.

I wrote some new nitypest test cases and convinced myself that the lsb always counts forwards in time because it's the lower half of a 128-bit two's complement number. It's like how adding 1 to 0xFFFFFFF0 (-16) yields 0xFFFFFFF1 (-15), which is less negative.

Sorry about the confusion.

src/handwritten/_lib_time.py
bkeryan: More questions about this timezone conversion...

Does the timezone offset account for daylight saving time? If so, the timezone offset now is not necessarily the timezone offset for some past or future date. For example, the offset would change by 1 hour on March 10th in the US.

Is this timezone conversion even used?
```
PS D:\dev\nidaqmx-python> rg "convert_timestamp_to_time|to_datetime" .\generated\nidaqmx\
.\generated\nidaqmx\_grpc_time.py
38:def convert_timestamp_to_time(ts: GrpcTimestamp, tzinfo: Optional[timezone] = None) -> ht_datetime:

.\generated\nidaqmx\_grpc_interpreter.py
21:from nidaqmx._grpc_time import convert_time_to_timestamp, convert_timestamp_to_time
2234:        return convert_timestamp_to_time(response.value)
3339:        return convert_timestamp_to_time(response.timestamp)

.\generated\nidaqmx\_lib_time.py
61:    def to_datetime(self, tzinfo: Optional[timezone] = None) -> ht_datetime:

.\generated\nidaqmx\_library_interpreter.py
3819:        return value.to_datetime()
5845:        return timestamp.to_datetime()
```

In the C API, I think the time trigger attributes like `DAQmxSetSyncPulseTimeWhen` pass the timestamp through without doing any timezone conversions. However, the channel calibration attributes like `DAQmxGetAIChanCalCalDate` convert to/from local time. They also break everything out into year, month, day, hour, minutes, so they would not use this code.

tests/unit/test_grpc_time.py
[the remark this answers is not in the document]
bkeryan: If this is a bug in hightime, please create a GitHub issue in the hightime repo.

src/handwritten/_lib_time.py
bkeryan: I still don't understand why this is different.

tests/unit/test_grpc_time.py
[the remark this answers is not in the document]
DeborahOoi96: hightime issue created: https://github.com/ni/hightime/issues/38

src/handwritten/_lib_time.py
DeborahOoi96: > Does the timezone offset account for daylight saving time?

Very good question, let me look into this a bit further before getting back to you.

> Is this timezone conversion even used?

Yes it is used, in python's `to_datetime` and grpc's `convert_timestamp_to_time`. You may refer to my reply in the [WaitForValidTimestamp PR](https://github.com/ni/nidaqmx-python/pull/510#discussion_r1498963296) to check that it's used. 

>In the C API, I think the time trigger attributes like DAQmxSetSyncPulseTimeWhen pass the timestamp through without doing any timezone conversions. 

If what they take is a timestamp value, then I believe the timestamp value in itself should already have taken into consideration the timezone conversion.

> the channel calibration attributes like DAQmxGetAIChanCalCalDate convert to/from local time. They also break everything out into year, month, day, hour, minutes, so they would not use this code. Also, I suspect that channel calibration dates may not return values before 1970.

It seems like we currently don't have support for these attributes in python, would this also be part of the things to port over in daqmx-python? cc @WayneDroid

src/handwritten/_lib_time.py
DeborahOoi96: You are right, total_seconds() is the better option. I have updated it to use total_seconds()

src/handwritten/_lib_time.py
bkeryan: > > Does the timezone offset account for daylight saving time?
> 
> Very good question, let me look into this a bit further before getting back to you.

One way to look at this is to write a unit test that compares the result of the standard `.astimezone()` method with your function for various dates in the current year (such as the first of each month, which should include some DST and some non-DST dates in regions that have DST).

> > Is this timezone conversion even used?
>
> Yes it is used, in python's to_datetime and grpc's convert_timestamp_to_time. You may refer to my reply in the https://github.com/ni/nidaqmx-python/pull/510#discussion_r1498963296 to check that it's used.

Thanks. The thing I missed is that the default arg `tzinfo=None` means "convert to local time", not "don't convert timezone".

> > the channel calibration attributes like DAQmxGetAIChanCalCalDate convert to/from local time. They also break everything out into year, month, day, hour, minutes, so they would not use this code. Also, I suspect that channel calibration dates may not return values before 1970.
>
> It seems like we currently don't have support for these attributes in python, would this also be part of the things to port over in daqmx-python? cc @WayneDroid

In AB#1645316, @zhindes put channel calibration in the "I wouldn't add, but might be useful to add someday" category.

bkeryan (approved):

src/handwritten/_lib_time.py
WayneDroid: @DeborahOoi96 @bkeryan Yup, Channel Calibration properties are not planned to be included in 1.0 release.

DeborahOoi96: @bkeryan Unfortunately, the timezone offset did not account for daylight saving time. :( I have modified it now to accommodate daylight savings times but this involves adding a new dependency, `pytz` in the unit tests. This seems to be breaking PR build due to poetry.lock not being updated,  but I am facing this certification error when trying to run poetry lock. 
image

May I know how to update the dependency in daqmx-python?

As for what went wrong in my previous time conversion function:
1. It used the *local time offset* to calculate the expected datetime to be returned to the user. Which means if I were to run the function in a DST-supported region **at a time in which DST is supported**, it should theoretically work (since astimezone does account for DST).
2. But for the sake of unit test, we want to check if the offset for a future/past date is the same as the offset from a present date, even if they are in the same DST region. This would vary in the cases of DST and be inconsistent.

So I used pytz.localize to help me with this. It will directly convert the `expected_time_utc` to the expected timezone while considering DST.

DeborahOoi96: But adding a new dependency, `pytz` in the unit tests seems to be breaking PR build due to poetry.lock not being updated,  but I am facing this certification error when trying to run poetry lock. 
image

May I know how to update the dependency in daqmx-python?

src/handwritten/_lib_time.py
DeborahOoi96: Moved to _time.py

generated/nidaqmx/_time.py
```diff
@@ -0,0 +1,46 @@
...
+# theoretically the same as astimezone(), but with support for dates before 1970
+def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[timezone] = None):
+    # if timezone matches, no need to do conversion
+    if expected_time_utc.tzinfo is tzinfo:
+        return expected_time_utc
+
+    # if timezone is not defined, use system timezone
+    if tzinfo is None:
+        local_timezone = get_localzone()
+        tzinfo = pytz.timezone(str(local_timezone))
```
bkeryan, line 19: This uses `tzlocal.get_localzone()`, which returns a `zoneinfo.ZoneInfo`. Then it uses `pytz.timezone()` to convert the `ZoneInfo` into a `datetime.tzinfo`. However, [zoneinfo](https://docs.python.org/3/library/zoneinfo.html) says

> [ZoneInfo](https://docs.python.org/3/library/zoneinfo.html#zoneinfo.ZoneInfo) is a concrete implementation of the [datetime.tzinfo](https://docs.python.org/3/library/datetime.html#datetime.tzinfo) abstract base class, and is intended to be attached to tzinfo, either via the constructor, the [datetime.replace](https://docs.python.org/3/library/datetime.html#datetime.datetime.replace) method or [datetime.astimezone](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone):

Do you really need `pytz`? Based on the docs, I think you can set `tzinfo = get_localzone()`.

generated/nidaqmx/_time.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # use pytz here to account for daylight savings
+    if isinstance(tzinfo, pytz.tzinfo.BaseTzInfo):
+        target_tz = pytz.timezone(str(tzinfo))
+        # we need to make expected_time_utc naive in order to use localize
+        expected_time_utc = expected_time_utc.replace(tzinfo=None)
+        localized_time = target_tz.localize(expected_time_utc)
+        desired_expected_time = target_tz.fromutc(localized_time)
```
bkeryan, line 27: `pytz`'s localize method seems to call `dt.replace(tz=self)`: https://github.com/stub42/pytz/blob/master/src/pytz/__init__.py#L242

Do you really need `pytz`?

generated/nidaqmx/_time.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+import pytz
+from tzlocal import get_localzone
+from datetime import timezone
```
bkeryan, line 5: I think you want `datetime.tzinfo`, not `datetime.timezone`.

https://docs.python.org/3/library/datetime.html#datetime.tzinfo

> This is an abstract base class, meaning that this class should not be instantiated directly. Define a subclass of [tzinfo](https://docs.python.org/3/library/datetime.html#datetime.tzinfo) to capture information about a particular time zone.
>
> ...
>
> The datetime module provides [timezone](https://docs.python.org/3/library/datetime.html#datetime.timezone), a simple concrete subclass of [tzinfo](https://docs.python.org/3/library/datetime.html#datetime.tzinfo) which can represent timezones with fixed offset from UTC such as UTC itself or North American EST and EDT.

We should use the abstract base class `datetime.tzinfo` in function signatures to allow a more sophisicated implementation such as `zoneinfo.ZoneInfo`.

generated/nidaqmx/_time.py
```diff
@@ -0,0 +1,46 @@
...
+            desired_expected_time.year,
+            desired_expected_time.month,
+            desired_expected_time.day,
+            desired_expected_time.hour,
+            desired_expected_time.minute,
+            desired_expected_time.second,
+            desired_expected_time.microsecond,
+            desired_expected_time.femtosecond,
+            desired_expected_time.yoctosecond,
+            tzinfo = tzinfo)
```
bkeryan, line 45: Can this case use `desired_expected_time.replace(tzinfo=tzinfo)`? It looks equivalent.

If you replace `pytz`'s `localize` method with `.replace(tzinfo=...)` don't these cases end up being the same?

bkeryan (changes requested):

generated/nidaqmx/_time.py
DeborahOoi96: we still need datetime.timezone to pass in timezone.utc in `current_time_utc = ht_datetime.now(timezone.utc)`, but I have modified the function parameter to be datetime.tzinfo instead of timezone

DeborahOoi96:
`tzlocal` dependency added in #521. All tests passes on my local machine.

src/handwritten/_time.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # if timezone is not defined, use system timezone
+    if tzinfo is None:
+        tzinfo = get_localzone()
+
+    # use ZoneInfo here to account for daylight savings
+    if isinstance(tzinfo, ZoneInfo):
+        localized_time = expected_time_utc.replace(tzinfo=tzinfo)
+        desired_expected_time = tzinfo.fromutc(localized_time)
+        return(desired_expected_time)
```
bkeryan, line 25: FYI, check for understanding: the original time is in UTC. `replace(tzinfo=)` changes the timezone without adjusting the time. `tzinfo.fromutc()` adjusts the time. So this emulates `astimezone` without hitting the 1970 case.

src/handwritten/_time.py
```diff
@@ -0,0 +1,33 @@
...
+from tzlocal import get_localzone
+from datetime import timezone
+from datetime import tzinfo as dt_tzinfo
+from datetime import datetime as std_datetime
+from hightime import datetime as ht_datetime
+from typing import Optional, Union
+from zoneinfo import ZoneInfo
+
+# theoretically the same as astimezone(), but with support for dates before 1970
+def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[dt_tzinfo] = None):
```
bkeryan, line 12:
```suggestion
def _convert_to_desired_timezone(expected_time_utc: Union[std_datetime, ht_datetime], tzinfo: Optional[dt_tzinfo] = None) -> Union[std_datetime, ht_datetime]:
```

src/handwritten/_time.py
```diff
@@ -0,0 +1,33 @@
...
+        tzinfo = get_localzone()
+
+    # use ZoneInfo here to account for daylight savings
+    if isinstance(tzinfo, ZoneInfo):
+        localized_time = expected_time_utc.replace(tzinfo=tzinfo)
+        desired_expected_time = tzinfo.fromutc(localized_time)
+        return(desired_expected_time)
+
+    # if the tzinfo passed in is a timedelta function, then we don't need to consider daylight savings
+    elif isinstance(tzinfo, dt_tzinfo) and tzinfo.utcoffset(None) is not None:
```
bkeryan, line 28: FYI, since tzinfo is the abstract base class, `isinstance(tzinfo, dt_tzinfo)` will be true unless the caller passes in a non-tzinfo or get_localzone() returns a non-tzinfo. You can probably remove this isinstance check. It's not a big deal though.

src/handwritten/_time.py
```diff
@@ -0,0 +1,33 @@
...
+        current_time_utc = ht_datetime.now(timezone.utc)
+        desired_timezone_offset = current_time_utc.astimezone(tz=tzinfo).utcoffset()
+        desired_expected_time = expected_time_utc + desired_timezone_offset
+        new_datetime = desired_expected_time.replace(tzinfo=tzinfo)
+        return new_datetime
```
bkeryan, line 33: If we don't hit the cases for ZoneInfo or utcoffset, then what? If we fall off the end of the function, then it implicitly returns None. Would it be better to raise an exception or return the original datetime?

bkeryan (approved):
Approved with suggestions:
- Consider adding a final else case to the timezone conversion function
- Type hint timezone conversion return value

src/handwritten/_time.py
DeborahOoi96: yep you got it right :)

src/handwritten/_time.py
bkeryan: Maybe the else case should fall back to `astimezone()`.

PR #505, opened 2024-02-20
tests: Fix mypy errors and run mypy during PR build

bkeryan:

Fix mypy errors in tests. For details, see commits.

Update PR build workflow to install extras and run mypy (tests only, for now).

Fix mypy errors that I found when I added the nidaqmx-python test fixtures to an internal test component.

Ooch towards being able to run mypy on the entire component.

Ran `poetry run mypy tests` and `poetry run pytest -v`


tests/unit/test_grpc_time.py
```diff
@@ -10,14 +10,10 @@
...
 
 try:
-    from google.protobuf.timestamp_pb2 import Timestamp as GrpcTimestamp
-
     import nidaqmx._grpc_time as grpc_time
     import nidaqmx._stubs.nidaqmx_pb2 as nidaqmx_pb2
 except ImportError:
-    GrpcTimestamp = None
-    grpc_time = None
-    nidaqmx_pb2 = None
```
bkeryan, line 20: We only need to assign things to None when we want to check them at run time.

tests/component/test_stream_readers_di.py
```diff
@@ -179,12 +179,12 @@ def _bool_array_to_int(bool_array: numpy.typing.NDArray[numpy.bool_]) -> int:
     return result
 
 
-_DType = TypeVar("_DType", bound=numpy.typing.DTypeLike)
+_D = TypeVar("_D", bound=numpy.generic, covariant=True)
```
bkeryan, line 182: @zhindes It turns out numpy.generic was right after all, but we want a type var to make the return type preserve the parameter type.

This also implicitly relies on the fact that the DAQmx read methods are untyped (Any).

tests/_grpc_utils.py
```diff
@@ -48,22 +50,23 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self._stdout_thread.join()
 
     def _get_grpc_server_exe(self):
-        if os.name != "nt":
-            pytest.skip("Only supported on Windows")
-        import winreg
+        if sys.platform == "win32":
+            import winreg
```
bkeryan, line 54: mypy recognizes `sys.platform` but not necessarily `os.name`

Also this relied on mypy knowing that pytest.skip never returns, but apparently it does know that: 
https://github.com/pytest-dev/pytest/blob/8a410d0ba60bca7073bf18a1503c59a3b5445780/src/_pytest/outcomes.py#L120

bkeryan: ...which explains all the "unreachable code" warnings for return statements after a pytest.skip

zhindes (approved):

maxxboehme (approved):

PR #510, opened 2024-02-21
Add functional test for wait_for_valid_timestamp

DeborahOoi96:

Add unit test coverage for wait_for_valid_timestamp

Existing and newly added tests all pass
image

### DIFF
--- a/tests/component/_task_modules/test_triggers.py
+++ b/tests/component/_task_modules/test_triggers.py
@@ -1,12 +1,12 @@
 from datetime import timezone
 
-import nidaqmx
 import pytest
 from hightime import datetime as ht_datetime, timedelta as ht_timedelta
 
-from nidaqmx.constants import Timescale, TimestampEvent, TriggerType, Edge
-from nidaqmx.task import Task
+import nidaqmx
+from nidaqmx.constants import Edge, Timescale, TimestampEvent, TriggerType
 from nidaqmx.error_codes import DAQmxErrors
+from nidaqmx.task import Task
 
 
 @pytest.fixture()
@@ -17,9 +17,7 @@ def ai_voltage_task(task, sim_time_aware_9215_device):
 
 
 @pytest.fixture()
-def ci_count_edges_task(
-    task
-) -> nidaqmx.Task:
+def ci_count_edges_task(task) -> nidaqmx.Task:
     chan = task.ci_channels.add_ci_count_edges_chan(
         "cdaqChassisTester/_ctr0",
     )
@@ -109,6 +107,9 @@ def test___first_sample_trigger___wait_for_valid_timestamp___no_errors(
     ai_voltage_task.start()
 
     ai_voltage_task.wait_for_valid_timestamp(TimestampEvent.FIRST_SAMPLE)
+
+    assert ai_voltage_field_daq_task.timing.first_samp_timestamp_enable == True
+    assert ai_voltage_field_daq_task.timing.first_samp_timestamp_timescale == Timescale.USE_HOST
 
 
 def test___timestamp_not_enabled___wait_for_valid_timestamp___throw_error(
--- a/tests/component/_task_modules/test_triggers.py
+++ b/tests/component/_task_modules/test_triggers.py
@@ -19,9 +19,7 @@ def ai_voltage_task(task, sim_time_aware_9215_device):
 @pytest.fixture()
 def ci_count_edges_task(task, sim_9185_device) -> nidaqmx.Task:
     device_name = sim_9185_device.name
-    chan = task.ci_channels.add_ci_count_edges_chan(
-        device_name + "/_ctr0"
-    )
+    chan = task.ci_channels.add_ci_count_edges_chan(device_name + "/_ctr0")
     chan.ci_count_edges_term = "/" + device_name + "/te0/SampleClock"
     chan.ci_count_edges_active_edge = Edge.RISING
     return task
--- a/tests/component/_task_modules/test_triggers.py
+++ b/tests/component/_task_modules/test_triggers.py
@@ -17,11 +17,12 @@ def ai_voltage_task(task, sim_time_aware_9215_device):
 
 
 @pytest.fixture()
-def ci_count_edges_task(task) -> nidaqmx.Task:
+def ci_count_edges_task(task, sim_9185_device) -> nidaqmx.Task:
+    device_name = sim_9185_device.name
     chan = task.ci_channels.add_ci_count_edges_chan(
-        "cdaqChassisTester/_ctr0",
+        device_name + "/_ctr0"
     )
-    chan.ci_count_edges_term = "/cdaqChassisTester/te0/SampleClock"
+    chan.ci_count_edges_term = "/" + device_name + "/te0/SampleClock"
     chan.ci_count_edges_active_edge = Edge.RISING
     return task
 
--- a/tests/component/_task_modules/test_triggers.py
+++ b/tests/component/_task_modules/test_triggers.py
@@ -19,8 +19,8 @@ def ai_voltage_task(task, sim_time_aware_9215_device):
 @pytest.fixture()
 def ci_count_edges_task(task, sim_9185_device) -> nidaqmx.Task:
     device_name = sim_9185_device.name
-    chan = task.ci_channels.add_ci_count_edges_chan(device_name + "/_ctr0")
-    chan.ci_count_edges_term = "/" + device_name + "/te0/SampleClock"
+    chan = task.ci_channels.add_ci_count_edges_chan(f"{sim_9185_device.name}/_ctr0")
+    chan.ci_count_edges_term = f"/{sim_9185_device.name}/te0/SampleClock"
     chan.ci_count_edges_active_edge = Edge.RISING
     return task
 
--- a/tests/component/_task_modules/test_triggers.py
+++ b/tests/component/_task_modules/test_triggers.py
@@ -107,8 +107,8 @@ def test___first_sample_trigger___wait_for_valid_timestamp___no_errors(
 
     ai_voltage_task.wait_for_valid_timestamp(TimestampEvent.FIRST_SAMPLE)
 
-    assert ai_voltage_field_daq_task.timing.first_samp_timestamp_enable
-    assert ai_voltage_field_daq_task.timing.first_samp_timestamp_timescale == Timescale.USE_HOST
+    assert ai_voltage_task.timing.first_samp_timestamp_enable
+    assert ai_voltage_task.timing.first_samp_timestamp_timescale == Timescale.USE_HOST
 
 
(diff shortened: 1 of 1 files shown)


DeborahOoi96:
I'm not quite sure how to test this, so I just went with no_error / expected error tests. 
I have tried reading back xxx_trigger.timestamp_val in Labview but I can't seem to get back a value. Maybe I'm testing it wrongly because I'm new to daqmx. Maybe some reference on how to properly use wait_for_valid_timestamp will be helpful.

tests/component/_task_modules/test_triggers.py
```diff
@@ -1,24 +1,38 @@
...
-    """Gets AI voltage task."""
+    """Gets AI voltage field daq task."""
     task.ai_channels.add_ai_voltage_chan(sim_field_daq_device.ai_physical_chans[0].name)
     yield task
 
 
+@pytest.fixture()
+def ci_count_edges_task(task) -> nidaqmx.Task:
+    chan = task.ci_channels.add_ci_count_edges_chan(
+        "cdaqChassisTester/_ctr0",
```
zhindes, line 22: You should make a `sim_9185_device` fixture to `conftest.py` rather than hardcoding this name

tests/component/_task_modules/test_triggers.py
```diff
@@ -51,3 +67,59 @@ def test___arguments_provided___cfg_time_start_trig___no_errors(
...
+
+
+def test___start_trigger___wait_for_valid_timestamp___no_errors(
+    ai_voltage_field_daq_task: Task,
+):
+    ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
+    ai_voltage_field_daq_task.triggers.start_trigger.timestamp_enable = True
+    ai_voltage_field_daq_task.start()
+
+    ai_voltage_field_daq_task.wait_for_valid_timestamp(TimestampEvent.START_TRIGGER)
```
zhindes, line 79: FYI these tests are all failing


```
tests\component\_task_modules\test_triggers.py:79: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
generated\nidaqmx\task.py:1036: in wait_for_valid_timestamp
    self._interpreter.wait_for_valid_timestamp(self._handle, timestamp_event.value, timeout)
generated\nidaqmx\_grpc_interpreter.py:3339: in wait_for_valid_timestamp
    return convert_timestamp_to_time(response.timestamp)
generated\nidaqmx\_grpc_time.py:49: in convert_timestamp_to_time
    dt = ht_datetime.fromtimestamp(seconds, timezone.utc)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'hightime.datetime'>, t = -2082844800, tz = datetime.timezone.utc

    @classmethod
    def fromtimestamp(cls, t, tz=None):
        # NOTE: Does not support sub-microsecond values!
>       result = std_datetime.datetime.fromtimestamp(t, tz)
E       OSError: [Errno 22] Invalid argument
```

zhindes (changes requested):
What do you mean the timestamp_val doesn't return anything? This is a simulated device, so I would at least expect a 0 timestamp...

tests/component/_task_modules/test_triggers.py
DeborahOoi96: they are failing because the test returns AbsoluteTime 0, which isnt supported in the libtime (because my fix to support negative timestamps is not in yet)

DeborahOoi96:
> What do you mean the timestamp_val doesn't return anything? This is a simulated device, so I would at least expect a 0 timestamp...

@zhindes sorry I wasn't clear in the description. I meant it only returns AbsoluteTime 0. I have updated the comment

zhindes:
> > What do you mean the timestamp_val doesn't return anything? This is a simulated device, so I would at least expect a 0 timestamp...
> 
> @zhindes sorry I wasn't clear in the description. I meant it only returns AbsoluteTime 0. I have updated the comment

Ah ok. Well that's the best we can do with simulated device. That being said, I suspect simulated devices always return that regardless of whether or not we waited for a timestamp. So, yeah, not very useful. I'm fine with what you have.

tests/component/_task_modules/test_triggers.py
zhindes: Ah, ok :)

tests/component/_task_modules/test_triggers.py
```diff
@@ -1,24 +1,38 @@
...
+import nidaqmx
+from nidaqmx.constants import Edge, Timescale, TimestampEvent, TriggerType
+from nidaqmx.error_codes import DAQmxErrors
 from nidaqmx.task import Task
 
 
 @pytest.fixture()
 def ai_voltage_field_daq_task(task, sim_field_daq_device):
-    """Gets AI voltage task."""
+    """Gets AI voltage field daq task."""
```
bkeryan, line 14: Update this for https://github.com/ni/nidaqmx-python/pull/513

tests/component/_task_modules/test_triggers.py
```diff
@@ -1,24 +1,38 @@
...
+
+
 def test___default_arguments___cfg_time_start_trig___no_errors(
     ai_voltage_field_daq_task: Task,
 ):
     ai_voltage_field_daq_task.timing.cfg_samp_clk_timing(1000)
-    trigger_time = ht_datetime.now() + timedelta(seconds=10)
+    utc_dt = ht_datetime.now(timezone.utc)  # UTC time
+    dt_now = utc_dt.astimezone()  # local time
+    trigger_time = dt_now + ht_timedelta(seconds=10)
```
bkeryan, line 35: `trigger_time` is in local time. However, the interpreter class's [get_trig_attribute_timestamp](https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_library_interpreter.py#L3806) doesn't specify a timezone. So how is `when_value` coming back in local time?

bkeryan (changes requested):

tests/component/_task_modules/test_triggers.py
DeborahOoi96: I believe the function name should also be changed to ai_voltage_compact_daq_task since that PR was to replace FieldDAQ with cDAQ. I will create another PR for to address this comment.

tests/component/_task_modules/test_triggers.py
DeborahOoi96: Thanks for helping to create a sim_time_aware_9215_device fixture!

DeborahOoi96: Thanks for helping to add `sim_9185_device` in your PR!

DeborahOoi96: I still need to hardcode this to some degree because 9185's counter channel (_ctr0) cannot be accessed by the typical approach of `sim_9185_device.ci_physical_chans[0].name`. My guess is the _ctr0 is not a physical channel, so `sim_9185_device.ci_physical_chans.channel_names` returns [].

tests/component/_task_modules/test_triggers.py
DeborahOoi96: To briefly summarize the whole thing, 
trigger time gets calculated as AbsoluteTime in the interpreter class's [from_datetime](https://github.com/ni/nidaqmx-python/blob/0905e65642ee934d011fc594d52829fce21d3fcd/generated/nidaqmx/_library_interpreter.py#L403C32-L403C45) which does a simple subtraction following the formula in my libtime PR
`seconds_since_1904 = int((dt - AbsoluteTime._EPOCH_1904) / ht_timedelta(seconds=1))`.
This, in itself, returns the correct number of seconds _**taking into consideration timezone difference**_. I checked this on an online compiler
image

So, timezone is accounted for when setting `trigger_time`. 

Then when reading it back, as you said it goes to [get_trig_attribute_timestamp](https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_library_interpreter.py#L3806), which returns the value [**_in datetime_**](https://github.com/ni/nidaqmx-python/blob/0905e65642ee934d011fc594d52829fce21d3fcd/generated/nidaqmx/_library_interpreter.py#L3819C20-L3819C35). This to_datetime function will call `_convert_to_desired_timezone` when returning the values, as seen in my [lib_time PR](https://github.com/ni/nidaqmx-python/pull/504) here. 
image

This shows that we do use this timezone conversion at least in python.

tests/component/_task_modules/test_triggers.py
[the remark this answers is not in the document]
DeborahOoi96: Yes absolutely :) I have tested with the libtime changes in my local test machine and test results are already posted in the description of this PR.

tests/component/_task_modules/test_triggers.py
zhindes: Hah, woops, sorry for leaving that name confusing! I agree with Brad, no need to be precise here.

tests/component/_task_modules/test_triggers.py
zhindes: Yup, hardcoding the "_ctr0" part is fine. I've done stuff like that in other tests.

tests/component/_task_modules/test_triggers.py
```diff
@@ -14,11 +16,22 @@ def ai_voltage_field_daq_task(task, sim_time_aware_9215_device):
     yield task
 
 
+@pytest.fixture()
+def ci_count_edges_task(task, sim_9185_device) -> nidaqmx.Task:
+    device_name = sim_9185_device.name
+    chan = task.ci_channels.add_ci_count_edges_chan(device_name + "/_ctr0")
```
zhindes, line 22: I find fstrings to be cleaner

```suggestion
    chan = task.ci_channels.add_ci_count_edges_chan(f"{sim_9185_device.name}/_ctr0")
```

tests/component/_task_modules/test_triggers.py
```diff
@@ -14,11 +16,22 @@ def ai_voltage_field_daq_task(task, sim_time_aware_9215_device):
...
+    chan.ci_count_edges_term = "/" + device_name + "/te0/SampleClock"
```
zhindes, line 23:
```suggestion
    chan.ci_count_edges_term = f"/{sim_9185_device.name}/te0/SampleClock"
```

zhindes (approved):
and fix the field_daq task name

tests/component/_task_modules/test_triggers.py
DeborahOoi96: Addressed in this PR #520

bkeryan (approved):

Issue #483, opened 2024-02-01
add_ai_rosette_strain_gage_chan parameter rosette_meas_types has the wrong type


zhindes:
The current signature:

```python
def add_ai_rosette_strain_gage_chan(
            self, physical_channel, rosette_type, gage_orientation,
            rosette_meas_types, name_to_assign_to_channel="", min_val=-0.001,
            max_val=0.001,
            strain_config=StrainGageBridgeType.QUARTER_BRIDGE_I,
            voltage_excit_source=ExcitationSource.INTERNAL,
            voltage_excit_val=2.5, gage_factor=2.0,
            nominal_gage_resistance=350.0, poisson_ratio=0.3,
            lead_wire_resistance=0.0):
```

The docstring says:
```python
rosette_meas_types (List[int]): Specifies information about
   the rosette configuration and measurements.
```

The code matches the docstring and converts the measurement types to an int array:
```python
rosette_meas_types = numpy.int32(rosette_meas_types)
```

The actual type should be `List[nidaqmx.constants.StrainGageRosetteMeasurementType]`

PR #511, opened 2024-02-21
Fix parameter rosette_meas_types type in docstrings

WayneDroid:
- [ x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).

Added `'python_codegen_method': 'CustomCode'` to `CreateAIRosetteStrainGageChan` in order to update parameter `rosette_meas_types` type in the function.

This PR is to address issue #483.

image


generated/nidaqmx/_task_modules/ai_channel_collection.py
```diff
@@ -57,6 +56,79 @@ def _create_chan(self, physical_channel, name_to_assign_to_channel=''):
...
+                measurement.
+        Returns:
+            nidaqmx._task_modules.channels.ai_channel.AIChannel:
+
+            Indicates the newly created channel object.
+        """
+        if rosette_meas_types is None:
+            rosette_meas_types = []
+
+        rosette_meas_types = numpy.int32(rosette_meas_types)
```
zhindes, line 120: It's not just the docstring that is wrong. This is wrong, too. This should be converting the `.value` of the enumerations. The way to validate your change is to update [this test](https://github.com/ni/nidaqmx-python/blob/a7a3643911fd549cac198c0a2cc8dadd4a8ad8f4/tests/component/_task_modules/channels/test_ai_channel.py#L848) to not create an array of `.value`, just the enumerations themselves.

zhindes (changes requested):
This is either a gap in our support for arrays of enumerations or some sort of issue with this specific function definition in the metadata. I want to say we have array of enums somewhere else in the API, so I think maybe the metadata is bad?

zhindes:
> This is either a gap in our support for arrays of enumerations or some sort of issue with this specific function definition in the metadata. I want to say we have array of enums somewhere else in the API, so I think maybe the metadata is bad?

My point here is I don't think just no longer code-generating rosette is correct. I think there is a metadata problem.

generated/nidaqmx/_task_modules/ai_channel_collection.py
bkeryan: Note that the way this uses `numpy.int32(some_list)` to create a NumPy array is not well documented and `numpy.typing` doesn't allow this behavior. If we touch any code like this, let's update it to use `numpy.array(some_list, dtype=numpy.int32)`.

bkeryan:
> > This is either a gap in our support for arrays of enumerations or some sort of issue with this specific function definition in the metadata. I want to say we have array of enums somewhere else in the API, so I think maybe the metadata is bad?
> 
> My point here is I don't think just no longer code-generating rosette is correct. I think there is a metadata problem.

I think `rosette_meas_types = numpy.int32(rosette_meas_types)` is generated based on `get_instantiation_lines`:
https://github.com/ni/nidaqmx-python/blob/a7a3643911fd549cac198c0a2cc8dadd4a8ad8f4/src/codegen/utilities/function_helpers.py#L136

It has a special case for enum arrays, so yeah, I think the metadata for this parameter is wrong.

generated/nidaqmx/_task_modules/ai_channel_collection.py
WayneDroid: @zhindes  @bkeryan I updated functions.py to have `rosette_meas_types` parameter reflected with correct type, I'm not sure is this correct approach. If it's okay, I will continue to work on scrapigen to update the metadata accordingly.

src/codegen/templates/_task_modules/ai_channel_collection.py.mako
```diff
@@ -37,8 +37,8 @@ class AIChannelCollection(ChannelCollection):
             name_to_assign_to_channel (Optional[str]): Specifies a name to
                 assign to the virtual channel this method creates.
         Returns:
-            nidaqmx._task_modules.channels.ai_channel.AIChannel: 
-            
+            nidaqmx._task_modules.channels.ai_channel.AIChannel:
+
             Specifies the newly created AIChannel object.
```
bkeryan, line 42: Adding a blank line between the return type and return description looks wrong. I guess it's moving the 1st line to the 2nd paragraph. I know some of the other files in this repo do this too but that doesn't make it better than what was there before.

https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html 

BTW this method is private so it doesn't really matter how we document it.

bkeryan (approved):
Approved with suggestions (revert doc change or delete unnecessary docstring)

Sorry, the review was in pending for a couple days because I forgot to submit it.

src/codegen/templates/_task_modules/ai_channel_collection.py.mako
WayneDroid: The blank line between return type and description has been there in current file ([ai_channel_collection.py.mako](https://github.com/ni/nidaqmx-python/blob/205dc3bcd6f5c32c7d4f1ed46e66af30af65f927/src/codegen/templates/_task_modules/ai_channel_collection.py.mako#L41)). Based on the diff-view on generated ai_channel_collection.py, I believe it's my code editor removed whitespaces at line 41 on mako file when I saved the file after reverting the changes from first commit.
image

I find that other public methods having a blank line between return type and description, maybe I should keep the blank line to align with existing docstring format.

bkeryan: Oh, I see. Removing trailing whitespace without changing the formatting is fine.

maxxboehme (approved):

generated/nidaqmx/_task_modules/ai_channel_collection.py
```diff
@@ -1503,13 +1503,15 @@ def add_ai_rosette_strain_gage_chan(
...
+
             Indicates the newly created channel object.
         """
         if rosette_meas_types is None:
             rosette_meas_types = []
 
-        rosette_meas_types = numpy.int32([p.value for p in rosette_meas_types])
+        temp_vals = [p.value for p in rosette_meas_types]
+
+        rosette_meas_types = numpy.int32(temp_vals)
```
WayneDroid, line 1514: @bkeryan  @zhindes Help needed for the mypy check error. 

image

I tried type hint `List[StrainGageRosetteMeasurementType]` to `rosette_meas_types`, but I still get the similar error. Any suggestions? 
image
image

image

zhindes: I think the solution is to type `temp_vals` like this:


```python
temp_vals: List[int] = [p.value for p in rosette_meas_types]
```

zhindes (approved):

generated/nidaqmx/_task_modules/ai_channel_collection.py
bkeryan: Rerunning codegen on your branch does this:
```
-        temp_vals = [p.value for p in rosette_meas_types]
-
-        rosette_meas_types = numpy.int32(temp_vals)
+        rosette_meas_types = numpy.int32([p.value for p in rosette_meas_types])
```

Which tells me two things:
- I missed one of the codegen templates
- build.yml should check for dirty codegen *before* it runs linters :)

bkeryan: You need to update this: https://github.com/ni/nidaqmx-python/blob/master/src/codegen/utilities/function_helpers.py#L133

If `param.ctypes_data_type.startswith("numpy.")` then it should generate `numpy.array({1}, dtype={0})`, and likewise for the array version.

WayneDroid: @bkeryan @zhindes Thanks for the suggestions.

For some reasons, I can't use `numpy.array({1}, dtype={0})` for `rosette_meas_types` (enum list). As a generated code `rosette_meas_types = numpy.array(rosette_meas_types, dtype=numpy.int32)` causing pytest `Type Error` failure. I have to extract value from enum list with `numpy.array([p.value for p in {0}], dtype={1})`.

image

bkeryan: That's right. I should have said "and likewise for the enum-array version"

bkeryan (approved):

PR #512, opened 2024-02-21
Automate Driver Install Experience within nidaqmx Module

dharaniprakashkm:

1. Provides a command line python interface to install NI-DAQmx Software for Windows Platform.
2. Checks for existing installation of DAQmx by checking the registry entry, If DAQmx Software is not installed then it will download the latest installer from ni.com and will launch the installer in an interactive mode.
3. If DAQmx Software is already installed then it will compare the installed version with latest available version on the server and prompts for user confirmation on whether they want to upgrade to the latest version or not.
 

We need to install the NI-DAQmx Driver to have Python library functional. 
This script will improve the driver install experience for Windows Users.

Verified the script works fine on both Windows 11 and Windows Server 2016 OS.

1. Clean Machine
2. Lower version of NI-DAQmx driver  is preinstalled
3. Latest version of NI-DAQmx driver is preinstalled
4. Higher version (Dev builds for NI internal users) of NI-DAQmx driver is preinstalled

Verified script handles/returns proper error info in the following cases.
1. Broken download link 
2. Incorrect version.
3. Running  script without admin privileges


pyproject.toml
```diff
@@ -72,6 +72,9 @@ pytest-mock = ">=3.0"
 pykka = ">=3.0"
 tox = ">=3.24"
 
+[tool.poetry.scripts]
+install-daqmx-driver = 'nidaqmx._install_daqmx:install_daqmx_software'
```
maxxboehme, line 76: We should probably add a step [here ](https://github.com/ni/nidaqmx-python/blob/a7a3643911fd549cac198c0a2cc8dadd4a8ad8f4/.github/workflows/CI.yml#L14) that runs this script to validate that it continues to work.

installDAQmx.py
```diff
@@ -0,0 +1,81 @@
+import urllib.request
```
maxxboehme, line 1: Do we need this if we already have the version in pyproject.toml to call?

generated/nidaqmx/_install_daqmx.py
```diff
@@ -0,0 +1,154 @@
+import urllib.request
```
maxxboehme, line 1: This seems to be the same file as the one is src/handwritten.

bkeryan: The codegen process copies the handwritten files to generated/nidaqmx.

generated/nidaqmx/_install_daqmx.py
bkeryan, line 1: Is this ready for feedback? I noticed it's a draft PR, but I glanced at the code and had some comments I didn't want to wait to post.

generated/nidaqmx/_install_daqmx.py
```diff
@@ -0,0 +1,154 @@
...
+                        print(f"Skipping Upgrade.")
+                    
+                else:
+                    print("Installation aborted due to missing download URL or temporary file.")
+            else:
+                print(f"Installed NI-DAQmx version {daqmx_installed_version} is up-to-date. Skipping installation.")
+        else:
+            print("Administrator privileges required. Installation aborted.")
+    else:
+        print("Unsupported OS detected.")
```
bkeryan, line 150: None of these error cases return a failure exit code to the OS.

installDAQmx.py
dharaniprakashkm: We dont need this file, I had copied for testing initial. I have deleted it

generated/nidaqmx/_install_daqmx.py
dharaniprakashkm: I have removed in the latest changes.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Verifies the check for a user having Administrator privileges.
+    """
+    try:
+        return ctypes.windll.shell32.IsUserAnAdmin()
+    except Exception as e:
+        print(f"Error checking admin privileges: {e}")
+        return False
+def parse_version(version):
+    """
```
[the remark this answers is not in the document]
dharaniprakashkm: @bkeryan , I installed ni-python-styleguide and and ran block formatter for the file  src/handwritten/_install_daqmx.py.
python -m black -l 100 

Do you want me to run ni-python-styleguide lint ?

generated/nidaqmx/_install_daqmx.py
dharaniprakashkm: @bkeryan , I was preparing for it but I am fine with early feedback.

src/handwritten/_install_daqmx.py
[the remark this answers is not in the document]
mshafer-NI: > I installed ni-python-styleguide and and ran block formatter for the file src/handwritten/_install_daqmx.py.
python -m black -l 100

You should be able to just "poetry install" followed by "poetry run ni-python-styleguide fix" (handles black and import sorting)

> Do you want me to run ni-python-styleguide lint ?

Yes, you should make sure code passes the linter

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,154 @@
...
+            else:
+                print(f"Installed NI-DAQmx version {daqmx_installed_version} is up-to-date. Skipping installation.")
+        else:
+            print("Administrator privileges required. Installation aborted.")
+    else:
+        print("Unsupported OS detected.")
+
+
+if __name__ == "__main__":
+    install_daqmx_software()
```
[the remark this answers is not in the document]
bkeryan: Either `_main` or `_install_daqmx` is ok with me.

(I don't think suffixing the function name with `_software` adds anything, unless it's necessary to avoid ambiguity, and I don't think that's the case.)

pyproject.toml
bkeryan: That would install DAQmx. Should the script have a command line option that causes it to exit without installing DAQmx (like --version)?

src/handwritten/_install_daqmx.py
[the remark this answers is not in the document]
bkeryan: Yes, but again, src/handwritten is currently excluded: https://github.com/ni/nidaqmx-python/blob/master/pyproject.toml#L75

So `poetry run ni-python-styleguide lint` or `fix` will currently ignore this file.

Someone should remove that exclusion, reformat the handwritten code, and acknowledge existing violations, but that is outside the scope of this PR.

pyproject.toml
mshafer-NI: or --help

generated/nidaqmx/_install_daqmx.py
bkeryan: Whether to call `sys.exit` depends on which command line parser you are using. See https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/65873/Common-Modules-List

`click` is recommended for command line programs that can use 3rd party modules. If you use `click`'s decorators, then your `main` function will automatically `sys.exit` to report errors.

`argparse` is recommended for command line programs that cannot use 3rd party modules. If you use `argparse`, then you need to `sys.exit`. This is a common idiom:
```
if __name__ == '__main__':
    sys.exit(main())
```

If you do call `sys.exit`, -1 is not a good choice of exit code. Read https://docs.python.org/3/library/sys.html#sys.exit

Can/should nidaqmx-python depend on `click`? We should be careful what dependencies we add, but now that the `import_metadata` debacle is behind us, I think adding a dependency on `click` is fairly low risk.

Cc: @maxxboehme @mshafer-NI

mshafer-NI: `click` is not just commonly used, [it is our standard](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/35648/Command-Line-Interface-Parser)

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,154 @@
...
+        with winreg.OpenKeyEx(winreg.HKEY_LOCAL_MACHINE,
+                              r"SOFTWARE\National Instruments\NI-DAQmx\CurrentVersion",
+                              0, winreg.KEY_READ | winreg.KEY_WOW64_32KEY) as daqmx_reg_key:
+            product_name = winreg.QueryValueEx(daqmx_reg_key, "ProductName")[0]
+            product_version = winreg.QueryValueEx(daqmx_reg_key, "Version")[0]
+
+        if product_name == "NI-DAQmx":
+            return product_version
+    except Exception as e:
+        print(f"Info: No existing NI-DAQmx installation found")
```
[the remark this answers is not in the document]
bkeryan: To use logging:
```python
import logging
...
_logger = logging.getLogger(__name__)
...
_logger.info("No existing NI-DAQmx installation found.")
```
You can find examples of this in nidaqmx-python itself.

To configure logging, I would do this:
```python
@click.command
@click.option(
    "-v",
    "--verbose",
    "verbosity",
    count=True,
    help="Enable verbose logging. Repeat to increase verbosity.",
)
def main(verbosity: int) -> None:
    """..."""
    _configure_logging(verbosity)


def _configure_logging(verbosity: int) -> None:
    """Configure logging for this process."""
    if verbosity > 1:
        level = logging.DEBUG
    elif verbosity == 1:
        level = logging.INFO
    else:
        level = logging.WARNING
    logging.basicConfig(format="%(asctime)s %(levelname)s: %(message)s", level=level)
```
You can find more examples of this in the [MeasurementLink Python examples](https://github.com/ni/measurementlink-python/blob/main/examples/nidaqmx_analog_input/measurement.py).

Note that NI has a [click mixin](https://dev.azure.com/ni/DevCentral/_git/Tools?path=%2Fshared%2Fniclickutils%2Fniclickutils%2F_commands.py&version=GBmain&_a=contents) for this, which is more sophisticated, but it's not publicly available at this time.

Cc: @mshafer-NI

src/handwritten/_install_daqmx.py
bkeryan, line 1: Offline, you brought up the idea of invoking this through `nidaqmx install` instead of `install-nidaqmx-software`. To make this happen, in addition to renaming the entry point / script, you would also add a click command for "install".

This shows how to do that: https://click.palletsprojects.com/en/8.1.x/quickstart/#nesting-commands

You can find lots of examples of this on codesearch if you search for "@click\.group".

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
...
+    Compare two versions which are in the format: major.minor.update
+    """
+    iver_parts = _parse_version(installed_ver)
+    sver_parts = _parse_version(source_ver)
+
+    for iver_num, sver_num in zip(iver_parts + [0], sver_parts + [0]):
+        if iver_num < sver_num:
+            return -1
+        elif iver_num > sver_num:
+            return 1
```
bkeryan, line 44: Convert the versions to tuples and their comparison operators will automatically do lexical comparisons. The tuples don't have to have the same number of elements.

This is often used with [sys.version_info](https://docs.python.org/3/library/sys.html#sys.version_info):
- `if sys.version_info >= (3, 9):`
- `if sys.version_info < (3, 12, 2):`

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
...
+    try:
+        return ctypes.windll.shell32.IsUserAnAdmin()
+    except Exception as e:
+        print(f"Error checking admin privileges: {e}")
+        return False
+
+
+def _parse_version(version):
+    """
+    Split the version string into an array of integers.
```
bkeryan, line 28: This function returns a list, not an array. However, I think you should change it to return a tuple.

To do that, use `return tuple(int(part) for part in version.split("."))`

@mshafer-NI If there is a standard way that doesn't require additional packages, please let me know.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
+import urllib.request
+import json
+import tempfile
+import platform
+import os
+import subprocess
+import winreg
```
bkeryan, line 7: All platform-specific imports must be conditional:
```python
if sys.platform == "win32":
    import winreg
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
...
+            r"SOFTWARE\National Instruments\NI-DAQmx\CurrentVersion",
+            0,
+            winreg.KEY_READ | winreg.KEY_WOW64_32KEY,
+        ) as daqmx_reg_key:
+            product_name = winreg.QueryValueEx(daqmx_reg_key, "ProductName")[0]
+            product_version = winreg.QueryValueEx(daqmx_reg_key, "Version")[0]
+
+        if product_name == "NI-DAQmx":
+            return product_version
+    except Exception as e:
```
bkeryan, line 66: This exception should probably be more specific.
- `winreg` should raise `OSError`, you should catch that.
- `OSError` indicates the Windows error code. If you check the error code, you can distinguish between key-not-found and other errors. That way, you don't treat other errors as "DAQmx is not installed".

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
...
+    return False
+
+
+def _get_temp_file_name():
+    """
+    Generate a temporary file name.
+    """
+    try:
+        _, temp_file_path = tempfile.mkstemp()
+        return temp_file_path + "_daq-mx-online.exe"
```
bkeryan, line 78: `tempfile.mkstemp()` securely creates an actual temp file. Discarding the file handle and adding a suffix loses the security benefits of this. Also, you are responsible for deleting the temp file.

It may be better to use `tempfile.mkdtemp()` to create a temp subdirectory and download the installer into it. Then you don't have to rename the installer, and you aren't subverting the security benefits of the `tempfile` module. 

However, you will still leak a temp directory. The solution here is probably to invoke the installer in a way that waits for it to complete, and then delete the temp directory. A try/finally block or a context manager can ensure this cleanup happens in error cases.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
...
+    return None, None
+
+
+def _install_daqmx(download_url, temp_file):
+    """
+    Download and launch NI-DAQmx Driver installation in an interactive mode
+    """
+    try:
+        urllib.request.urlretrieve(download_url, temp_file)
+        subprocess.Popen([temp_file], shell=True)
```
bkeryan, line 111: `subprocess.Popen` is for when you want to interact with the child process's stdin, stdout, and/or stderr. If you aren't doing that, use a different `subprocess` function such as `subprocess.run`.

Also, this leaks a subprocess, which is why you got a `ResourceWarning` when you demoed this earlier. Using `subprocess.run` will wait for the command to complete, which should fix this leak.

bkeryan (changes requested):

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
+import urllib.request
+import json
+import tempfile
+import platform
```
bkeryan, line 4: Don't use `platform`. Use `sys.platform`.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
...
+def install_daqmx_software():
+    """
+    Get the platform name on which the Python script is being executed.
+    Check for the existing installaion of NI-DAQmx Driver.
+    If NI-DAQmx driver is already installed it will prompt for an upgrade if applicable.
+    Download and launch NI-DAQmx Driver installation in an interactive mode.
+    """
+
+    if sys.platform.startswith("win"):
+        if _is_admin():
```
bkeryan, line 139: Is this admin check required on Windows? Can the installer elevate to administrator when you run it? Test with a standard user account on a machine that has UAC enabled.

If the user is using a standard user account, running Python scripts as administrator may cause them problems down the line, such as creating cache files with administrator permissions that the standard user can't delete or overwrite.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
...
+                    print("Installation aborted due to missing download URL")
+            else:
+                print(
+                    f"Installed NI-DAQmx version {daqmx_installed_version} is up-to-date. Skipping installation."
+                )
+        else:
+            print("Administrator privileges required. Installation aborted.")
+            return
+    else:
+        print("Unsupported OS detected.")
```
bkeryan, line 181: This is deeply nested. It would be more readable if it was flatter and each error case threw an exception or returned early.

You can raise `click.ClickException`, `click.UsageError`, `click.BadArgument`, etc. to report errors to the user.
https://click.palletsprojects.com/en/8.1.x/exceptions/#which-exceptions-exist 

I thought I commented on this already, but I can't find it.

src/handwritten/_install_daqmx.py
dharaniprakashkm: @bkeryan , Installer can elevate to administrator and I have tested that.  We need administrator permissions to read windows registry, I will check with the product owner and/or @zhindes may be we can drop checking the existing installation of DAQmx software and we dont need the admin permissions.

bkeryan: > We need administrator permissions to read windows registry,

You should not need administrator permissions to read that registry key. Please try it again with your current code. Perhaps you were specifying `access=winreg.KEY_WOW64_32KEY` without `winreg.KEY_READ` at one point?

image

src/handwritten/_install_daqmx.py
zhindes: I prefer `nidaqmx install`, but I think that is a bit too generic. How about `nidaqmx installdriver`?

src/handwritten/_install_daqmx.py
zhindes: Related comment, for new code lets use [type hints](https://docs.python.org/3/library/typing.html).

src/handwritten/_install_daqmx.py
zhindes: Yep, on Windows I expect no admin privileges will be required. For Linux that is probably not the case, but we'll think about that later. Glad you've confirmed.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,185 @@
...
+                _META_DATA_URL
+            )
+            if daqmx_latest_version == None or daqmx_latest_url == None:
+                return -1
+            daqmx_installed_version = _check_daqmx_is_installed()
+            download_url, temp_file = daqmx_latest_url, _get_temp_file_name()
+            print(download_url)
+            print(temp_file)
+            if temp_file == None:
+               return -1
```
zhindes, line 150: Lots of cleanup to do here for "sad path" (opposite of [happy path](https://en.wikipedia.org/wiki/Happy_path)).

zhindes (changes requested):
Really great start. Sorry I missed the demo :(

src/handwritten/__main__.py
```diff
@@ -0,0 +1,36 @@
...
+
+@click.group("nidaqmx")
+@click.option(
+    "-v",
+    "--verbose",
+    "verbosity",
+    count=True,
+    help="Enable verbose logging. Repeat to increase verbosity."
+)
+
```
mshafer-NI, line 19: what's with this extra newline?

pyproject.toml
dharaniprakashkm: Added the --help option that will invoke the script but it will not install DAQmx driver

src/handwritten/_install_daqmx.py
dharaniprakashkm: I have refactored the code, @bkeryan , could you please check the updated code is fine?

src/handwritten/_install_daqmx.py
dharaniprakashkm: I have refactored the code, @zhindes could you please check once ?

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+def _get_daqmx_installed_version() -> str:
+    """
+    Check for existing installation of DAQmx by checking for Version information.
+    Version information is read from the registry key.
+    
+    Returns:
+        str: The version of the installed NI-DAQmx if found, else an empty string.
+
+    >>> _get_daqmx_installed_version()  # doctest: +SKIP
+    ''
```
mshafer-NI, line 100: Let's not put "skipped" tests in
```suggestion
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+        response = input(user_message + " (yes/no): ").strip().lower()
+        if response in ["yes", "y"]:
+            return True
+        elif response in ["no", "n"]:
+            return False
+        else:
+            print("Please enter 'yes' or 'no'.")
+
+
+def _create_user_question(latest_version: str, installed_version: str ) -> str:
```
mshafer-NI, line 256: This isn't used anywhere anymore, please remove this method

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+    Check for the existing installation of NI-DAQmx Driver.
+    If NI-DAQmx driver is already installed it will prompt for an upgrade if applicable.
+    Download and launch NI-DAQmx Driver installation in an interactive mode.
+    """
+    try:
+        if sys.platform.startswith("win"):
+            logging.info("Windows platform detected")
+            _install_daqmx_windows_driver()
+        else:
+            print(f"installdriver command is supported only on Windows")
```
mshafer-NI, line 328: Should this be an exception case?

src/handwritten/metadata.json
```diff
@@ -0,0 +1,28 @@
+{
```
mshafer-NI, line 1: consider normalizing the indentation in this file using 4 spaces.

```python
import json

import _install_daqmx

raw = _install_daqmx.METADATA_FILE.read_text()
data = json.loads(raw)
formatted = json.dumps(data, indent=" "*4)

_install_daqmx.METADATA_FILE.write_text(formatted)
```

.github/workflows/run_system_tests.yml
```diff
@@ -23,6 +23,8 @@ jobs:
         uses: actions/checkout@v3
       - name: Install dependencies
         run: poetry install
+      - name: check installdriver subcommand can be invoked
+        run: poetry run nidaqmx installdriver --help
```
bkeryan, line 27: The goal of `run_system_tests.yml` is to run Tox, which automatically iterates over multiple Python versions. Move this to `run_unit_tests.yml`, so it is run with each Python version.

pyproject.toml
```diff
@@ -82,7 +85,7 @@ extend_exclude = ".tox,docs,generated,src/codegen/metadata,src/codegen/templates
 [tool.pytest.ini_options]
 addopts = "--doctest-modules --strict-markers"
 filterwarnings = ["always::ImportWarning", "always::ResourceWarning"]
-testpaths = ["tests"]
+testpaths = [ "tests" , "generated/nidaqmx/_install_daqmx.py"]
```
bkeryan, line 88: Is this for doctests? If so, add all of generated/nidaqmx/, not just your new module.

src/handwritten/metadata.json
zhindes, line 1: Ensure this is actually bundled in the package. If not, you may need to add it to the [Poetry include config](https://python-poetry.org/docs/pyproject/#include-and-exclude).

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+    (23, 8, 0)
+    >>> _parse_version("24.0.0")
+    (24, 0, 0)
+    >>> _parse_version("invalid_version")
+    Traceback (most recent call last):
+    ...
+    click.exceptions.ClickException: Invalid version format found
+    """
+    try:
+        return tuple(int(part) for part in version.split("."))
```
zhindes, line 41: [namedtuple](https://docs.python.org/3.10/library/collections.html?highlight=namedtuple#collections.namedtuple) would be good here.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+    """
+    Split the version string into a tuple of integers representing major, minor, and update parts.
+
+    Args:
+        version (str): The version string in the format "major.minor.update".
+
+    Returns:
+        tuple: A tuple of integers representing the major, minor, and update parts of the version.
+
+    >>> _parse_version("23.8.0")
```
zhindes, line 31: does our pytest setup run these comment tests automatically?

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+    """
+    try:
+        logging.debug("Entering _compare_version")
+        latest_parts: Tuple[int, ...] = _parse_version(latest_version)
+        installed_parts: Tuple[int, ...] = _parse_version(installed_version)
+
+        if latest_parts > installed_parts:
+            return True
+        else:
+            print(f"Latest Version already installed")
```
zhindes, line 83: Maybe give a bit more information here...

```suggestion
            print(f"Installed NI-DAQmx version ({installed_version}) is equivalent or newer than the known version to install, {latest_parts}.")
```

src/handwritten/__main__.py
bkeryan, line 1: The new submodules are missing from `generated/nidaqmx`.

```
PS D:\dev\nidaqmx-python> poetry install --only main
Creating virtualenv nidaqmx in D:\dev\nidaqmx-python\.venv
Installing dependencies from lock file

Package operations: 4 installs, 0 updates, 0 removals

  - Installing packaging (23.2)
  - Installing deprecation (2.1.0)
  - Installing hightime (0.2.1)
  - Installing numpy (1.26.0)

Installing the current project: nidaqmx (0.9.1)
PS D:\dev\nidaqmx-python> poetry run nidaqmx installdriver --help
  File "<string>", line 1, in <module>
  File "C:\Users\Administrator\.pyenv\pyenv-win\versions\3.9.13\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'nidaqmx.__main__'
```

To fix this, run the codegen script described in https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md#getting-started and commit the changes to your branch.

The GitHub workflow runs the codegen script and fails if `generated/nidaqmx` is out-of-date. It does not automatically update `generated/nidaqmx` for you.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+
+    >>> _create_user_question("24.0.0", "23.8.0")
+    'Installed NI-DAQmx version is 23.8.0. Latest version available is 24.0.0. Do you want to upgrade?'
+    """
+    user_question = (
+                      "Installed NI-DAQmx version is "
+                        + installed_version
+                        + ". Latest version available is "
+                        + latest_version
+                        + ". Do you want to upgrade?"
```
zhindes, line 274: fstring is cleaner

src/handwritten/metadata.json
bkeryan, line 1: This file has an overly general filename. Consider renaming to something like `_installer_metadata.json`.

bkeryan (changes requested):
I didn't get to the code yet. I'll post another review when I do.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+        raise click.ClickException(f'Permission denied while trying to read the registry key.') from e
+    except OSError as e:
+        logging.info("Error: %s", e)
+        logging.debug(traceback.format_exc())
+        raise click.ClickException(f'An OS error occurred while trying to read the registry key.') from e
+
+
+# Creating a temp file that we then close and yield - this is used to allow subprocesses to access
+@contextlib.contextmanager
+def multi_access_temp_file(*, suffix=".exe", delete=True):
```
zhindes, line 132: url_open seems to be able to handle a separate open file better. see example code [here](https://stackoverflow.com/questions/55864160/how-to-make-urllib-request-append-to-an-existing-file). consider that, and not creating a temp file, but not deleting. Then you can remove this function and use a simple with statement in the calling function.

zhindes (changes requested):

src/handwritten/__main__.py
```diff
@@ -0,0 +1,34 @@
...
+import click
+
+from . import _install_daqmx
+
+from typing import Optional
+
+
+
+@click.group("nidaqmx")
```
bkeryan, line 11: What is this group name used for? I don't think I've seen anyone set a group name for a top-level group before.

If I change the group name, `click` still prints "nidaqmx.cmd":
```
PS D:\dev\nidaqmx-python> poetry run nidaqmx installdriver --help
Usage: nidaqmx.cmd installdriver [OPTIONS]

Options:
  --help  Show this message and exit.
```

I guess that's because the file in `.venv\Scripts` is a `.cmd` file.

`python -m nidaqmx` doesn't seem to use the group name either.

src/handwritten/__main__.py
bkeryan, line 1: Do we expect `python -m nidaqmx installdriver` to work too? It doesn't work on my machine. 

I think this usage requires:
```
if __name__ == "__main__":
    main()
```

mshafer-NI: Good catch -> Yes, it is the norm for users to expect that work as well, which is why we put this under `__main__.py` -> @dharaniprakashkm, please add the code block bkeryan provided at the bottom of the file (formatted of course)

src/handwritten/_install_daqmx.py
bkeryan: That's what `testpaths = [ "tests" , "generated/nidaqmx/_install_daqmx.py"]` in `pyproject.toml` is about.

src/handwritten/_install_daqmx.py
bkeryan: Honestly, I think `Tuple[int, ...]` is fine. I don't think we need to access/print field names, and it's not critical to limit to a max number of fields.

src/handwritten/_install_daqmx.py
bkeryan, line 83: Please don't capitalize common nouns. Only capitalize proper nouns. (or 1st word in the sentence, of course)

Common nouns
- Version
- Product Name
- Driver

Proper nouns
- Windows
- Python
- NI-DAQmx

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+    click.exceptions.ClickException: Invalid version format found
+    """
+    try:
+        return tuple(int(part) for part in version.split("."))
+    except ValueError  as e:
+        logging.info("Error: %s", e)
+        logging.debug(traceback.format_exc())
+        raise click.ClickException(f'Invalid version format found') from e
+
+def _is_latest_version_greater(latest_version : str, installed_version: str) -> int:
```
bkeryan, line 47:
```suggestion
def _is_latest_version_greater(latest_version : str, installed_version: str) -> bool:
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+
+def _parse_version(version: str) -> Tuple[int, ...]:
+    """
+    Split the version string into a tuple of integers representing major, minor, and update parts.
+
+    Args:
+        version (str): The version string in the format "major.minor.update".
+
+    Returns:
+        tuple: A tuple of integers representing the major, minor, and update parts of the version.
```
bkeryan, line 29: You don't have to write full documentation for private helper functions. I think too much boilerplate documentation makes the code less readable, not more.

IMO, """Split the version string into a tuple of integers.""" is more than enough.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+    >>> _parse_version("invalid_version")
+    Traceback (most recent call last):
+    ...
+    click.exceptions.ClickException: Invalid version format found
+    """
+    try:
+        return tuple(int(part) for part in version.split("."))
+    except ValueError  as e:
+        logging.info("Error: %s", e)
+        logging.debug(traceback.format_exc())
```
bkeryan, line 44: Don't use the root logger. Add this to the top:
```
_logger = logging.getLogger(__name__)
```

Don't invent new ways to log exceptions. Use `_logger.exception("Failed to parse version.")`. It will automatically find the current exception and log its contents + traceback.

If you want the same behavior at info level, use `_logger.info("Failed to parse version.", exc_info=True)`.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+    Latest Version already installed
+    False
+    >>> _is_latest_version_greater("23.8.1", "23.8.0")
+    True
+    >>> _is_latest_version_greater("24.0.0", "23.8.5")
+    True
+    >>> _is_latest_version_greater("invalid_version", "1.2.0")
+    Traceback (most recent call last):
+    ...
+    click.exceptions.ClickException: Invalid version format found
```
bkeryan, line 73: I would only use ClickException for errors caused by something the user did.

What would the user do to cause this error? You only get version numbers from the metadata.json and the registry. So I think encountering this error indicates that there is a bug in either NI-DAQmx or this Python module.

Maybe this function doesn't need a try/except at all.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+            product_name = winreg.QueryValueEx(daqmx_reg_key, "ProductName")[0]
+            product_version = winreg.QueryValueEx(daqmx_reg_key, "Version")[0]
+
+        if product_name == "NI-DAQmx":
+            logging.debug("Found registry entries for Product Name: %s and version %s", product_name , product_version )
+            return product_version
+    except FileNotFoundError:
+        logging.info("No existing NI-DAQmx installation found.")
+        logging.debug(traceback.format_exc())
+        return None
```
bkeryan, line 119: `None` is not a `str`. Either return `""` or change the type hint to `-> Optional[str]`.

src/handwritten/_install_daqmx.py
bkeryan, line 132:
```suggestion
def multi_access_temp_file(*, suffix: str = ".exe", delete: bool = True) -> Generator[str, None, None]:
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+import tempfile
+import traceback
+from typing import List, Tuple, Optional
+import urllib.request
+
+if sys.platform.startswith("win"):
+   import winreg
+
+MODULE_DIR = pathlib.Path(__file__).parent
+METADATA_FILE = MODULE_DIR / "metadata.json"
```
bkeryan, line 19: This makes assumptions about how this package is deployed.

https://docs.python.org/3/library/importlib.resources.html has APIs for accessing files embedded in packages.

src/handwritten/_install_daqmx.py
bkeryan: Also please run Black on this file, just once?

src/handwritten/metadata.json
```diff
@@ -0,0 +1,28 @@
...
+    "Windows": [
+        {
+
+                "Version": "23.8.0",
+                "Location": "https://download.ni.com/support/nipkg/products/ni-d/ni-daqmx/23.8/online/ni-daqmx_23.8_online.exe",
+                "supportedOS": [
+                    "Windows 11",
```
bkeryan, line 8: Windows 10???

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,332 @@
...
+    except ValueError  as e:
+        logging.info("Error: %s", e)
+        logging.debug(traceback.format_exc())
+        raise click.ClickException(f'Invalid version format found') from e
+
+
+def _get_daqmx_installed_version() -> str:
+    """
+    Check for existing installation of DAQmx by checking for Version information.
+    Version information is read from the registry key.
```
bkeryan, line 94: The first line of a multi-line docstring is a one-line summary, followed by a blank line.

See https://peps.python.org/pep-0257/

For most of these helpers, a one-line docstring is enough. For example, no one calling this function needs to know that version information is read from the registry key, and anyone reading this function understands this without the comment.

pyproject.toml
dharaniprakashkm: getting below error when I try to add generated/nidaqmx/

image

src/handwritten/_install_daqmx.py
mshafer-NI: (+ the 
```
[tool.pytest.ini_options]
addopts = "--doctest-modules ...
```
args to pytest)

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,309 @@
...
+    """
+    Check for existing installation of DAQmx by checking for Version information.
+
+    Returns:
+        str: The version of the installed NI-DAQmx if found, else an empty string.
+<<<<<<< Updated upstream
+
+=======
+    ''
+>>>>>>> Stashed changes
```
mshafer-NI, line 100: This is a merge conflict. Please remove these lines.

pyproject.toml
```diff
@@ -45,6 +45,7 @@ sphinx_rtd_theme = {version="^1.0", optional=true}
 grpcio = {version=">=1.49.0,<2.0", optional = true}
 protobuf = { version = "^4.21", optional = true }
 hightime = "^0.2.1"
+click = "^8.0.0"
```
bkeryan, line 48: It is not a given that click 9.0 will break compatibility with existing applications. Please use `>=` instead of `^` so that nidaqmx-python will not limit users to click 8.x when this happens.
```suggestion
click = ">=8.0.0"
```

pyproject.toml
```diff
@@ -85,7 +86,7 @@ extend_exclude = ".tox,docs,generated,src/codegen/metadata,src/codegen/templates
 [tool.pytest.ini_options]
 addopts = "--doctest-modules --strict-markers"
 filterwarnings = ["always::ImportWarning", "always::ResourceWarning"]
-testpaths = [ "tests" , "generated/nidaqmx/_install_daqmx.py"]
+testpaths = [ "tests" , "generated/nidaqmx"]
```
bkeryan, line 89: Please don't leave extra spaces.
```suggestion
testpaths = ["tests" , "generated/nidaqmx"]
```

src/handwritten/_install_daqmx.py
```diff
@@ -323,8 +302,8 @@ def installdriver() -> None:
             logging.info("Windows platform detected")
             _install_daqmx_windows_driver()
         else:
-            print(f"installdriver command is supported only on Windows")
+            raise click.ClickException(f'Error installdriver command is supported only on Windows OS.')
```
bkeryan, line 305: Don't print "Error" without a colon because that's confusing. Also, `ClickException` should already prefix the error message with "Error:".

src/handwritten/_install_daqmx.py
```diff
@@ -325,8 +302,8 @@ def installdriver() -> None:
...
```
bkeryan, line 305: In American English, "on Windows" sounds more natural than "on Windows OS".

src/handwritten/_install_daqmx.py
```diff
@@ -325,8 +302,8 @@ def installdriver() -> None:
...
     except Exception as e:
         logging.info("Error: %s", e)
         logging.debug(traceback.format_exc())
-        raise click.ClickException(f'Error occurred during installdriver') from e
+        raise click.ClickException(f'Error occurred during installdriver') from e
```
bkeryan, line 309: Does this print the error details when logging is disabled or does it hide the details? If it hides the details, this error message is not going to make users successful. They are going to have to rerun the install with `-v`, and this error message doesn't tell them how to do that.

src/handwritten/_install_daqmx.py
```diff
@@ -80,7 +79,7 @@ def _is_latest_version_greater(latest_version : str, installed_version: str) ->
         if latest_parts > installed_parts:
             return True
         else:
-            print(f"Latest Version already installed")
+            print(f"Installed NI-DAQmx version ({installed_version}) is equivalent or newer than the known version to install, {latest_parts}.")
```
bkeryan, line 82: This message is extra wordy in a way that highlights a potential problem.

What is the significance of "the known version to install"? You hardcoded this version in a file in the nidaqmx-python driver. That made this version number known to the nidaqmx-python install script.

Are you and/or the Venus team planning to update this hardcoded file for each new version of NI-DAQmx? Does that include patches and minor versions?

Does this update policy align with the one used for NI Update Service on Windows?

In the long run, should this metadata be on ni.com instead?

bkeryan (changes requested):

src/handwritten/__main__.py
```diff
@@ -0,0 +1,34 @@
+import logging
+
+import click
```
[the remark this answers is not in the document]
bkeryan: You have not done step 4 yet.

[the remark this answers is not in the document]
dharaniprakashkm: @bkeryan, I have completed all the steps and tested. Could you please check once?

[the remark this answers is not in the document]
bkeryan: image
You still have not done step 4.

I am expecting something like
```
      - name: Install main package dependencies
        run: |
          python -m pip install --upgrade pip
          poetry install --only main
      - name: check installdriver subcommand can be invoked
        run: poetry run nidaqmx installdriver --help
      - name: Install dev dependencies
        run: poetry install
      - name: Run unit tests
```

pyproject.toml
bkeryan: Yes, that doctest is formatted wrong.

src/handwritten/__main__.py
bkeryan: This is not fixed.

image

I am expecting to see changes to generated/nidaqmx/

dharaniprakashkm: @bkeryan, Are you expecting me to submit only the files that I have modified or all the files within the 'generated/nidaqmx' directory? Based on the previous review comment, my understanding is that I shouldn't submit these files.

image

src/handwritten/_install_daqmx.py
dharaniprakashkm: @bkeryan, in the previous review comment @zhindes recommended to replace the message "Latest Version already installed" with "
Installed NI-DAQmx version (23.8.0) is equivalent or newer than the known version to install, (23, 8, 0)".
@zhindes your thoughts on this ?

Are you and/or the Venus team planning to update this hardcoded file for each new version of NI-DAQmx? Does that include patches and minor versions?
Yes. 

Does this update policy align with the one used for NI Update Service on Windows?
No. My understanding is that we don't have to align it, I can double check and confirm.

In the long run, should this metadata be on ni.com instead?
No. @zhindes ?

src/handwritten/_install_daqmx.py
dharaniprakashkm: @bkeryan, If the version numbers are not updated correctly in metadata.json then we will encounter an error message.
Shall I still go ahead and remove try/except ?

src/handwritten/_install_daqmx.py
zhindes: > In the long run, should this metadata be on ni.com instead?

Should it? Yes. Will it? No in the short-term, probably no in the long-term. Working with IT on this has been a really really big pain. Eventually we gave up and have resorted to this. A better solution is to leverage the NIPM data that already can do something like this. There are two problems:
* They currently only have a .NET assembly available to use their API, and they don't want to open up the backend impl in a way that customers can see it.
* It only works for Windows NIPM, not Linux. Linux might have some well-defined ni.com format for where we can find the package metadata, but I'd rather not rely on ni.com being stable, sadly.

So for now, keeping this metadata here and updating it when its useful to do so (new hardware is added probably) is what we're gonna go with.


>  Installed NI-DAQmx version (23.8.0) is equivalent or newer than the known version to install, (23, 8, 0)".

I can't think of a better phrase than "known" version. I don't want to suggest the version we know about is the latest or anything in case we don't update it. Hmm.... Also, can you update that string to match the format of the first version?


```python
f"Installed NI-DAQmx version ({installed_version}) is equivalent or newer than the known version to install, {known_version[0]}.{known_version[1]}.{known_version[2}".

File: src/handwritten/_install_daqmx.py (hunk 44 above)
Comment (dharaniprakashkm) [reply src/handwritten/_install_daqmx.py]: @bkeryan 
Shall I rephrase the error message something like this
"An error occurred during the installation of the driver. Please rerun the install command with verbose(-v) logging enabled to debug.
@zhindes any suggestions how you want to improve this error message ?

File: src/handwritten/_install_daqmx.py (hunk 29 above)
Comment (dharaniprakashkm) [reply src/handwritten/_install_daqmx.py]: @zhindes, If we use url_open still we need to provide a filename as input and it needs to unique and we should also not leak the file. 
I had a offline discussion with @mshafer-NI he recommended this approach and we have tested and its working fine.
Do you still recommend to consider using url_open?
please let me know, I will look into that option

File: src/handwritten/_install_daqmx.py (hunk 44 above)
Comment (zhindes) [reply src/handwritten/_install_daqmx.py]: I think Brad's suggestion is that an error should be logged in a way that is hidden. Your default logging level is warning, so info/debug won't be shown. Line 307 should at least be a `logging.error`. I am guessing Python's normal exception handling with print the traceback for you (unless you catch this later and squelch it), so 308 is probably fine to leave as-is or remove. But you can try all this and make sure the behavior is good. Try to induce an error of an install in NIPM. Perhaps killing the NIPM process during one of the selection dialogs would do it?

File: src/handwritten/_install_daqmx.py (hunk 29 above)
Comment (zhindes) [reply src/handwritten/_install_daqmx.py]: I was suggesting something like this:

```python
with urllib.request.urlopen("ni.com/daqmx_installer") as response:
    with tempfile.NamedTemporaryFile(suffix=suffix, mode="ab") as temp_file:
       temp_file.write(response.read())
       yield temp_file
```

But Matt is way smarter than me. If he says this is a bad idea, I believe him :)

mshafer-NI: I'm not familiar with the difference between urllib.request.urlopen and urllib.request.urlretrieve

I do know that after writing to a temp file, you have to close it before trying to do more -> this is what the "multi_access_temp_file" is trying to provide, create a temp file, close it (so that it's not "open" and thread locked), then cleanup once exiting the context manager.

src/handwritten/__main__.py
dharaniprakashkm: @bkeryan, as discussed offline, I have submitted the files that I have touched.

src/handwritten/_install_daqmx.py
dharaniprakashkm: @bkeryan and @zhindes , I have tested an exception case by changing logging level and observed that current code is good.
We will document how to enable verbose logging along with the installation instructions [here](https://nidaqmx-python.readthedocs.io/en/latest/), so that users will know how to enable debugging when an error is encountered.

src/handwritten/_install_daqmx.py
zhindes: Oooh, closing it so it isn't locked makes sense. I think my suggestion probably breaks that. I'll drop.

pyproject.toml
dharaniprakashkm: @bkeryan , When I try to enable doctests for generated/nidaqmx/ I observed 5 failures and I may not be the right person to fix those issues. 
I have created a [work item](https://ni.visualstudio.com/DefaultCollection/DevCentral/_workitems/edit/2694479) to track and enable doctests in the next iteration and for now I have disabled doctests.

src/handwritten/__main__.py
bkeryan: ok

src/handwritten/_install_daqmx.py
bkeryan: IMO it's weird for a version check function that returns a bool to also print to stdout or terminate the script with an error. It would be less weird if you deleted this function and put this code in the calling function:

```
latest_parts = _parse_version(latest_version)
installed_parts = _parse_version(installed_version)
if installed_parts >= latest_parts:
    print(...)
    return
```

src/handwritten/metadata.json
bkeryan: ok

src/handwritten/_install_daqmx.py
bkeryan: ok

src/handwritten/_install_daqmx.py
bkeryan: ok

src/handwritten/_install_daqmx.py
bkeryan: ok

src/handwritten/_install_daqmx.py
bkeryan: ok

src/handwritten/_install_daqmx.py
bkeryan: ok

src/handwritten/_install_daqmx.py
bkeryan: ok, this will be resolved by AB#2694479

src/handwritten/_install_daqmx.py
bkeryan: > I can't think of a better phrase than "known" version.

"It is known." --Game of Thrones

@zhindes How about "Installed NI-DAQmx version (x.x.x) is up to date. (Expected y.y.y or newer.)"?

zhindes: Love it!

zhindes (approved):
Nothing else to add that Brad isn't already catching

generated/nidaqmx/_install_daqmx.py
```diff
@@ -283,30 +208,32 @@ def _confirm_and_upgrade_daqmx_driver(
...
             )
-            if is_upgrade:
-                _install_daqmx_driver(download_url)
+        if is_upgrade:
+            _install_daqmx_driver(download_url)
     except Exception as e:
         _logger.info("Failed to upgrade NI-DAQmx driver.", exc_info=True)
         raise click.ClickException(
-            f"An error occurred during upgrading the driver software for Windows"
+            f"An error occurred during upgrading the MI-DAQmx driver software for Windows"
```
bkeryan, line 229:
```suggestion
            f"An error occurred during upgrading the NI-DAQmx driver software for Windows"
```

generated/nidaqmx/_install_daqmx.py
```diff
@@ -283,30 +208,32 @@ def _confirm_and_upgrade_daqmx_driver(
...
         _logger.debug("Entering _confirm_and_upgrade_daqmx_driver")
-        is_upgrade_case = _is_latest_version_greater(latest_version, installed_version)
-        if is_upgrade_case:
-            is_upgrade = _ask_user_confirmation(
-                f"Installed NI-DAQmx version is {installed_version}. Latest version available is {latest_version}. Do you want to upgrade?"
+        latest_parts: Tuple[int, ...] = _parse_version(latest_version)
+        installed_parts: Tuple[int, ...] = _parse_version(installed_version)
+        if installed_parts >= latest_parts:
+            print(
+                f"Installed NI-DAQmx version ({installed_version}) is equivalent or newer than the known version to install, ({latest_version})."
```
bkeryan, line 218:
```suggestion
                f"Installed NI-DAQmx version ({installed_version}) is equivalent or newer than the known version to install ({latest_version})."
```

bkeryan (approved):
Approved with suggestions

src/handwritten/_install_daqmx.py
dharaniprakashkm: @bkeryan @zhindes : Is this fine ? 

image

maxxboehme (approved):

src/handwritten/_install_daqmx.py
bkeryan: Leave a space between the period at the end of a sentence and the beginning of the next sentence.

y.y.y is a placeholder for the actual version, not something you should print.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+    Confirm with the user and upgrade the NI-DAQmx driver if necessary.
+
+    """
+    _logger.debug("Entering _confirm_and_upgrade_daqmx_driver")
+    latest_parts: Tuple[int, ...] = _parse_version(latest_version)
+    installed_parts: Tuple[int, ...] = _parse_version(installed_version)
+    if installed_parts >= latest_parts:
+        print(
+            f"Installed NI-DAQmx version ({installed_version}) is up to date."
```
bkeryan, line 217:
```suggestion
            f"Installed NI-DAQmx version ({installed_version}) is up to date. (Expected {latest_version} or newer.)"
```

src/handwritten/_install_daqmx.py
bkeryan: I downloaded your branch so I can test the error handling. It's still not right.

If I change the metadata file to specify an invalid version like "asdf", here's what I get:
```
PS D:\dev\nidaqmx-python> poetry run nidaqmx installdriver
Error: An error occurred during the installation of the driver :Invalid version format found:invalid literal for int() with base 10: 'asdf'
```

- The spacing around the colons is wrong.
- The top-level try-except is catching ClickException and wrapping it with another ClickException.

Also, please be aware that string interpolation calls str() for you.

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+
+        if product_name == "NI-DAQmx":
+            _logger.info(
+                "Found registry entries for Product Name: %s and version %s",
+                product_name,
+                product_version,
+            )
+            return product_version
+    except FileNotFoundError:
+        _logger.info("No existing NI-DAQmx installation found.", exc_info=False)
```
bkeryan, line 71: exc_info=False is the default.
```suggestion
        _logger.info("No existing NI-DAQmx installation found.")
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+    >>> _parse_version("invalid_version")
+    Traceback (most recent call last):
+    ...
+    click.exceptions.ClickException: Invalid version format found
+    """
+    try:
+        return tuple(int(part) for part in version.split("."))
+    except ValueError as e:
+        _logger.info("Failed to parse version.", exc_info=True)
+        raise click.ClickException(f"Invalid version format found:{str(e)}") from e
```
bkeryan, line 44:
```suggestion
        _logger.info("Failed to parse version.", exc_info=True)
        raise click.ClickException(f"Invalid version number '{version}'.") from e
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+            f"An error occurred while trying to to create temporary file:{str(e)}"
+        ) from e
+
+    try:
+        yield temp_file.name
+    finally:
+        if delete:
+            try:
+                os.unlink(temp_file.name)
+                _logger.debug("Deleting temp file: %s", temp_file.name)
```
bkeryan, line 108: It is customary to log "Doing X" before doing X, not after. That way, if X hangs or raises an error, you know what it was trying to do.
```suggestion
                _logger.debug("Deleting temp file: %s", temp_file.name)
                os.unlink(temp_file.name)
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+        _logger.info("Failed to parse the json data.", exc_info=True)
+        raise click.ClickException(f"Failed to parse the json data:{str(e)}") from e
+
+    for metadata_entry in metadata:
+        location: Optional[str] = metadata_entry.get("Location")
+        version: Optional[str] = metadata_entry.get("Version")
+        _logger.debug("From metadata file found location %s and version %s.", location, version)
+        if location and version:
+            return location, version
+    raise click.ClickException(f"Unable to fetch driver details:{str(e)}")
```
bkeryan, line 144: This is not in a try/except block, so there is no `e`.
```suggestion
    raise click.ClickException(f"Unable to find driver version in metadata.")
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def _get_driver_details() -> Tuple[Optional[str], Optional[str]]:
+    """
+    Parse the JSON data and retrieve the download link and version information.
+
+    """
+    try:
+        with pkg_resources.open_text(__package__, METADATA_FILE) as json_file:
+            _logger.debug("Opening the metadatafile %s.", METADATA_FILE)
```
bkeryan, line 154:
```suggestion
            _logger.debug("Opening the metadata file %s.", METADATA_FILE)
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+    except PermissionError as e:
+        _logger.info("Failed to read the registry key.", exc_info=True)
+        raise click.ClickException(
+            f"Permission denied while trying to read the registry key"
+        ) from e
+    except OSError as e:
+        _logger.info("Failed to read the registry key.", exc_info=True)
+        raise click.ClickException(
+            f"An OS error occurred while trying to read the registry key"
+        ) from e
```
bkeryan, line 82: User-facing errors should use user-facing terminology.

```suggestion
    except PermissionError as e:
        _logger.info("Failed to read the registry key.", exc_info=True)
        raise click.ClickException(
            f"Permission denied while getting the installed NI-DAQmx version."
        ) from e
    except OSError as e:
        _logger.info("Failed to read the registry key.", exc_info=True)
        raise click.ClickException(
            f"An OS error occurred while getting the installed NI-DAQmx version.\nDetails: {e}"
        ) from e
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+    try:
+        yield temp_file.name
+    finally:
+        if delete:
+            try:
+                os.unlink(temp_file.name)
+                _logger.debug("Deleting temp file: %s", temp_file.name)
+            except ValueError as e:
+                _logger.info("Failed to delete temporary file.", exc_info=True)
+                raise click.ClickException(f"Failed to delete temporary file:{str(e)}") from e
```
bkeryan, line 111:
```suggestion
                _logger.info("Failed to delete temporary file.", exc_info=True)
                raise click.ClickException(f"Failed to delete temporary file '{temp_file.name}'.\nDetails: {e}") from e
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+    Traceback (most recent call last):
+    ...
+    click.exceptions.ClickException: Unable to fetch driver details.
+
+    """
+    try:
+        metadata = json.loads(json_data).get("Windows", [])
+    except json.JSONDecodeError as e:
+        _logger.info("Failed to parse the json data.", exc_info=True)
+        raise click.ClickException(f"Failed to parse the json data:{str(e)}") from e
```
bkeryan, line 136:
```suggestion
        _logger.info("Failed to parse the json data.", exc_info=True)
        raise click.ClickException(f"Failed to parse the driver metadata.\nDetails: {e}") from e
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+        with pkg_resources.open_text(__package__, METADATA_FILE) as json_file:
+            _logger.debug("Opening the metadatafile %s.", METADATA_FILE)
+            location, version = _load_data(json_file.read())
+        return location, version
+
+    except Exception as e:
+        _logger.info("Failed to parse version.", exc_info=True)
+        raise click.ClickException(
+            f"An error occurred while trying to fetch driver details:{str(e)}"
+        ) from e
```
bkeryan, line 162:
```suggestion
        _logger.info("Failed to get driver metadata.", exc_info=True)
        raise click.ClickException(
            f"An error occurred while getting the driver metadata.\nDetails: {e}"
        ) from e
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+        _logger.info("Failed to installed NI-DAQmx driver.", exc_info=True)
+        raise click.ClickException(
+            f"An error occurred while installing NI-DAQmx driver. Command returned non-zero exit status."
+        ) from e
+    except urllib.error.URLError as e:
+        _logger.info("Failed to download NI-DAQmx driver.", exc_info=True)
+        raise click.ClickException(f"Failed to download NI-DAQmx driver:{str(e)}") from e
+    except Exception as e:
+        _logger.info("Failed to download NI-DAQmx driver.", exc_info=True)
+        raise click.ClickException(f"Failed to install NI-DAQmx driver:{str(e)}") from e
```
bkeryan, line 187:
```suggestion
    except subprocess.CalledProcessError as e:
        _logger.info("Failed to installed NI-DAQmx driver.", exc_info=True)
        raise click.ClickException(
            f"An error occurred while installing the NI-DAQmx driver. Command returned non-zero exit status '{e.returncode}'."
        ) from e
    except urllib.error.URLError as e:
        _logger.info("Failed to download NI-DAQmx driver.", exc_info=True)
        raise click.ClickException(f"Failed to download the NI-DAQmx driver.\nDetails: {e}") from e
    except Exception as e:
        _logger.info("Failed to download NI-DAQmx driver.", exc_info=True)
        raise click.ClickException(f"Failed to install the NI-DAQmx driver.\nDetails: {e}") from e
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+        if sys.platform.startswith("win"):
+            _logger.info("Windows platform detected")
+            _install_daqmx_windows_driver()
+        else:
+            raise click.ClickException(f"installdriver command is supported only on Windows.")
+    except Exception as e:
+        _logger.info("Failed to download NI-DAQmx driver.", exc_info=True)
+        raise click.ClickException(
+            f"An error occurred during the installation of the driver :{str(e)}"
+        ) from e
```
bkeryan, line 255:
```suggestion
        else:
            raise click.ClickException(f"The 'installdriver' command is supported only on Windows.")
    except click.ClickException:
        raise
    except Exception as e:
        _logger.info("Failed to install driver.", exc_info=True)
        raise click.ClickException(
            f"An error occurred during the installation of the NI-DAQmx driver.\nDetails: {e}"
        ) from e
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+    Parse the JSON data and retrieve the download link and version information.
+
+    """
+    try:
+        with pkg_resources.open_text(__package__, METADATA_FILE) as json_file:
+            _logger.debug("Opening the metadatafile %s.", METADATA_FILE)
+            location, version = _load_data(json_file.read())
+        return location, version
+
+    except Exception as e:
```
bkeryan, line 158:
```suggestion

    except click.ClickException:
        raise
    except Exception as e:
```

bkeryan (changes requested):

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def _get_daqmx_installed_version() -> Optional[str]:
+    """
+    Check for existing installation of NI-DAQmx.
+
+    """
+    try:
+        _logger.debug("Reading the registry entries to get installed DAQmx version")
+        with winreg.OpenKeyEx(
```
bkeryan, line 54: All `winreg` access should be inside an `if sys.platform == "win32":` block. The else case can raise an exception.

mypy caught this:
```
generated/nidaqmx/_install_daqmx.py:54: error: Name "winreg" is not defined  [name-defined]
generated/nidaqmx/_install_daqmx.py:55: error: Name "winreg" is not defined  [name-defined]
generated/nidaqmx/_install_daqmx.py:58: error: Name "winreg" is not defined  [name-defined]
generated/nidaqmx/_install_daqmx.py:60: error: Name "winreg" is not defined  [name-defined]
generated/nidaqmx/_install_daqmx.py:61: error: Name "winreg" is not defined  [name-defined]
```

src/handwritten/_install_daqmx.py
bkeryan: mypy caught this:
```
generated/nidaqmx/_install_daqmx.py:[14](https://github.com/ni/nidaqmx-python/actions/runs/8461256431/job/23181579846?pr=512#step:8:15)4: error: Trying to read deleted variable "e"  [misc]
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+            product_version = winreg.QueryValueEx(daqmx_reg_key, "Version")[0]
+
+        if product_name == "NI-DAQmx":
+            _logger.info(
+                "Found registry entries for Product Name: %s and version %s",
+                product_name,
+                product_version,
+            )
+            return product_version
+    except FileNotFoundError:
```
bkeryan, line 70: This is missing a return statement in the case where product_name != "NI-DAQmx".

mypy caught this:
```
generated/nidaqmx/_install_daqmx.py:47: error: Missing return statement  [return]
```

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+    Install the NI-DAQmx driver on Windows.
+
+    """
+    installed_version = _get_daqmx_installed_version()
+    download_url, latest_version = _get_driver_details()
+    if installed_version:
+        _confirm_and_upgrade_daqmx_driver(latest_version, installed_version, download_url)
+    else:
+        _install_daqmx_driver(download_url)
```
bkeryan, line 237: If a function returns `Optional[str]`, you must check whether the return value is `None` before passing it to another function that expects `str`. Perhaps some of these functions should raise an exception instead of returning `None`.
 
mypy caught this:
```
generated/nidaqmx/_install_daqmx.py:235: error: Argument 1 to "_confirm_and_upgrade_daqmx_driver" has incompatible type "Optional[str]"; expected "str"  [arg-type]
generated/nidaqmx/_install_daqmx.py:235: error: Argument 3 to "_confirm_and_upgrade_daqmx_driver" has incompatible type "Optional[str]"; expected "str"  [arg-type]
generated/nidaqmx/_install_daqmx.py:237: error: Argument 1 to "_install_daqmx_driver" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```

bkeryan (changes requested):

src/handwritten/_install_daqmx.py
```diff
@@ -104,8 +108,8 @@ def _multi_access_temp_file(*, suffix: str = ".exe", delete: bool = True) -> Gen
     finally:
         if delete:
             try:
-                os.unlink(temp_file.name)
                 _logger.debug("Deleting temp file: %s", temp_file.name)
+                os.unlink(temp_file.name)
             except ValueError as e:
                 _logger.info("Failed to delete temporary file.", exc_info=True)
                 raise click.ClickException(f"Failed to delete temporary file:{str(e)}") from e
```
bkeryan, line 115:
```suggestion
                raise click.ClickException(f"Failed to delete temporary file '{temp_file.name}'.\nDetails: {e}") from e
```

bkeryan (approved):

src/handwritten/_install_daqmx.py
bkeryan: I still see multiple instances of `{str(e)}` that need updated.

generated/nidaqmx/_install_daqmx.py
```diff
@@ -0,0 +1,266 @@
...
+        raise click.ClickException(f"Invalid version number '{version}'.") from e
+
+
+def _get_daqmx_installed_version() -> Optional[str]:
+    """
+    Check for existing installation of NI-DAQmx.
+
+    """
+    if sys.platform != 'win32':
+        return None
```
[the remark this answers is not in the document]
bkeryan: See the example of platform-specific code in https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

It puts the platform-specific code in the "if", "elif", and "else" blocks.

Your version uses an early return, which makes the rest of the function unreachable. Unreachable code is a different warning. Also, it still warns about calling Windows code on Linux, which suggests that mypy's platform-specific code support may not accept an early return.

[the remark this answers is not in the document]
dharaniprakashkm: @bkeryan I was running it on Windows. I have tested now on Linux machine and I see that it is passing that error.
Here is an updated log:
~/NI-DAQmx/nidaqmx-python$ poetry run mypy generated/nidaqmx
generated/nidaqmx/__init__.py:10: error: Name "version" already defined (possibly by an import)  [no-redef]
Found 1 error in 1 file (checked 72 source files)

generated/nidaqmx/_install_daqmx.py
```diff
@@ -49,41 +49,42 @@ def _get_daqmx_installed_version() -> Optional[str]:
...
+                f"Permission denied while getting the installed NI-DAQmx version.\nDetails: {e}"
+            ) from e
+        except OSError as e:
+            _logger.info("Failed to read the registry key.", exc_info=True)
+            raise click.ClickException(
+                f"An OS error occurred while getting the installed NI-DAQmx version.\nDetails: {e}"
+            ) from e
+    else:
+        _logger.error("This function is only supported on Windows.")
         return None
```
bkeryan, line 87:
```suggestion
        raise NotImplementedError("This function is only supported on Windows.")
```

src/handwritten/_install_daqmx.py
bkeryan: You fixed them all, but the fixes are not in generated/nidaqmx yet.

generated/nidaqmx/_install_daqmx.py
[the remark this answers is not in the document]
bkeryan: I'm getting this failure in the trunk, so it's not something you caused. I posted this PR to fix it: #554 

Please do this to cherry-pick the fix into your branch:
```
git fetch origin
git cherry-pick 201ebf467d53cfd6608c4aaddaca77e05e27b72b
```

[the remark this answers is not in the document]
dharaniprakashkm: I have cherry picked the changes into my branch.
image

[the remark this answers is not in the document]
bkeryan: Oops, I forgot to modify the original file in src/handwritten. This is fixed in commit fe748fd41c4dd03d6b9bd2ac630597776c1d8b84, but we should probably just merge the 1st PR and then you should merge master into your branch.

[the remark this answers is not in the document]
bkeryan: #554 is in master now. Merging master into your branch should fix the build.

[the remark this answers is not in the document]
dharaniprakashkm: Thanks @bkeryan ! I have pulled the latest changes into my branch.
image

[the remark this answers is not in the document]
dharaniprakashkm: Build is passed now.

bkeryan (approved):

generated/nidaqmx/__main__.py
```diff
@@ -0,0 +1,38 @@
...
+        level = logging.DEBUG
+    elif verbosity == 1:
+        level = logging.INFO
+    else:
+        level = logging.WARNING
+    logging.basicConfig(format="%(asctime)s %(levelname)s: %(message)s", level=level)
+
+
+if __name__ == "__main__":
+    main()
```
mshafer-NI, line 38: no blank line at end of file tells me that `black` has not run on this file:
a: can you run black on this please?
b: why isn't the PR catching this?

generated/nidaqmx/_installer_metadata.json
```diff
@@ -0,0 +1,31 @@
+{
```
mshafer-NI, line 1: indentation in this file is still weird. Please run:
```bash
poetry run python -c "import json, pathlib; p=pathlib.Path("generated/nidaqmx/_installer_metadata.json"); 
 raw=p.read_text(); data=json.loads(raw); output=json.dumps(data, indent=' '*4'); p.write_text(output, encoding='UTF-8')"
```

to format it nicely

src/handwritten/__main__.py
```diff
@@ -0,0 +1,38 @@
...
+        level = logging.DEBUG
+    elif verbosity == 1:
+        level = logging.INFO
+    else:
+        level = logging.WARNING
+    logging.basicConfig(format="%(asctime)s %(levelname)s: %(message)s", level=level)
+
+
+if __name__ == "__main__":
+    main()
```
mshafer-NI, line 38: ditto:
> no blank line at end of file tells me that black has not run on this file:
> a: can you run black on this please?
> b: why isn't the PR catching this?

src/handwritten/_install_daqmx.py
```diff
@@ -0,0 +1,265 @@
...
+            _install_daqmx_windows_driver()
+        else:
+            raise click.ClickException(f"The 'installdriver' command is supported only on Windows.")
+    except click.ClickException:
+        raise
+    except Exception as e:
+        _logger.info("Failed to install driver.", exc_info=True)
+        raise click.ClickException(
+            f"An error occurred during the installation of the NI-DAQmx driver.\nDetails: {e}"
+        ) from e
```
mshafer-NI, line 265: ditto:
> no blank line at end of file tells me that black has not run on this file:
> a: can you run black on this please?
> b: why isn't the PR catching this?

src/handwritten/_installer_metadata.json
```diff
@@ -0,0 +1,31 @@
+{
```
mshafer-NI, line 1: See other comment about formatting this file

src/handwritten/__main__.py
dharaniprakashkm: I have added it manually.

generated/nidaqmx/__main__.py
bkeryan: Sorry, this was uncaptured tech debt. I just captured it in https://github.com/ni/nidaqmx-python/issues/556

mshafer-NI (approved):

Issue #384, opened 2023-05-26
Support internationalization


bkeryan:
The DAQmx C API (`nicaiu.dll`) supports internationalization but not localization:
- User-provided strings have ANSI/MBCS encoding.
- Localized strings (error messages) are in English. Extended error info may also contain user-provided strings.

nidaqmx-python currently assumes that user-provided strings are ASCII and error messages are UTF-8. User-provided strings that use non-ASCII characters produce encoding errors.

Note that there is also a UTF-8 version of the DAQmx C API (`nicai_utf8.dll`) but nidaqmx-python does not currently use it.

Steps to reproduce:
```
PS D:\dev\nidaqmx-python> poetry run python
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import nidaqmx.system
>>> d = nidaqmx.system.Device("Gerät")
>>> d
Device(name=Gerät)
>>> d.name
'Gerät'
>>> d.serial_num
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\dev\nidaqmx-python\generated\nidaqmx\system\device.py", line 991, in serial_num
    val = self._interpreter.get_device_attribute_uint32(self._name, 0x632)
  File "D:\dev\nidaqmx-python\generated\nidaqmx\_library_interpreter.py", line 2572, in get_device_attribute_uint32
    error_code = cfunc(
ctypes.ArgumentError: argument 1: <class 'UnicodeEncodeError'>: 'ascii' codec can't encode character '\xe4' in position 3: ordinal not in range(128)
```

robin-po:
I have the same error when the device name or task name contains latin characters.
I succeed to make the following code working with some modifications in _library_interpreter.py:
```
import nidaqmx.system

d = nidaqmx.system.Device("Gerät")
print(d.name)
print(d.serial_num)
```
I've modified my `get_device_attribute_uint32` to convert the `device_name` to iso-8859-1:
```
    def get_device_attribute_uint32(self, device_name, attribute):
        value = ctypes.c_uint32()

        cfunc = lib_importer.cdll.DAQmxGetDeviceAttribute
        if cfunc.argtypes is None:
            with cfunc.arglock:
                if cfunc.argtypes is None:
                    cfunc.argtypes = [
                        ctypes_byte_str, ctypes.c_int32]

[...]

Issue #20, opened 2017-08-18
DAQmx Python API returns garbage values for non-english strings


epage:
We should call UTF-8 entry points which are available in DAQmx as of DAQmx 17.1.

epage:
We could possibly fallback to old entry points if the new ones aren't available.

PR #518, opened 2024-02-26
Support Non-ASCII Characters for User-Provided String

charitylxy:

This PR aims to address the issue brought up in https://github.com/ni/nidaqmx-python/issues/384 and https://github.com/ni/nidaqmx-python/issues/384
- Currently, user-provided strings only accept ASCII characters.
- This change will allow non-ASCII characters for user-provided string

- Argument with String type are now encoded to UTF-8 to fix the issues:
   * https://github.com/ni/nidaqmx-python/issues/384 
   * https://github.com/ni/nidaqmx-python/issues/20

Manually tested on command line with non ASCII string argument.


charitylxy:
Not sure if it's appropriate to still load `nicaiu.dll` while encoding strings to utf-8 but seems like it's working on my end even without changing it to `nicai_utf8.dll`.

generated/nidaqmx/_lib.py
```diff
@@ -46,13 +46,13 @@ def _setter(self, val):
...
     """
-    Custom argtype that automatically converts unicode strings to ASCII
+    Custom argtype that automatically converts unicode strings to UTF-8
     strings in Python 3.
     """
     @classmethod
     def from_param(cls, param):
         if isinstance(param, str):
-            param = param.encode('ascii')
+            param = param.encode('utf-8')
```
bkeryan, line 55: `nicaiu.dll` does not use UTF-8. I am 100% sure this will generate [mojibake](https://en.wikipedia.org/wiki/Mojibake), unless you are using code points that are the same in Latin-1 and UTF-8.

Please see [nidaqmx_internal/errors.py](https://codesearch.natinst.com/search/text?q=nidaqmx_internal%2Ferrors.py&fold_case=auto&regex=true&context=true) and [test_cerrorinfo.py](https://codesearch.natinst.com/search/text?q=test_cerrorinfo.py&fold_case=auto&regex=true&context=true). Try changing `errors.py` to use UTF-8 and then run pytest on `test_cerrorinfo.py`. It won't work.

It also doesn't work for Chinese, Japanese, or Korean without changing the system locale.

bkeryan (changes requested):

generated/nidaqmx/_lib.py
charitylxy: @bkeryan I have updated to load `nicai_utf8.dll`. Do we want to add a simulated device with non ASCII device name so that it can be covered in some of the regression tests?

generated/nidaqmx/_lib.py
bkeryan, line 1: Problem: I don't think the UTF-8 version of the DAQmx C API gets installed on Linux.

```
$ ls -al /usr/lib/x86_64-linux-gnu/libnidaq*
lrwxrwxrwx 1 root root      15 Apr  2  2023 /usr/lib/x86_64-linux-gnu/libnidaqmx.so -> libnidaqmx.so.1
lrwxrwxrwx 1 root root      20 Apr  2  2023 /usr/lib/x86_64-linux-gnu/libnidaqmx.so.1 -> libnidaqmx.so.23.3.0
-rwxr-xr-x 1 root root 2762224 Apr  1  2023 /usr/lib/x86_64-linux-gnu/libnidaqmx.so.23.3.0
lrwxrwxrwx 1 root root      22 Apr  2  2023 /usr/lib/x86_64-linux-gnu/libnidaqmxStorage.so -> libnidaqmxStorage.so.1
lrwxrwxrwx 1 root root      27 Apr  2  2023 /usr/lib/x86_64-linux-gnu/libnidaqmxStorage.so.1 -> libnidaqmxStorage.so.23.3.0
-rwxr-xr-x 1 root root  197512 Apr  1  2023 /usr/lib/x86_64-linux-gnu/libnidaqmxStorage.so.23.3.0
```

This repo's PR workflow doesn't run the system tests on Linux yet, so it won't catch this. We recently added a Linux Python test event, which you can find on dashboard under Uncategorized >> DAQmx Python Tests. It's currently disabled due to AB#2665445.

generated/nidaqmx/_lib.py
bkeryan: > I have updated to load nicai_utf8.dll.

I think this is the right long-term solution, but in the short term we probably need to do one of the following:
- Fall back to MBCS if the UTF-8 DLL is not available
- Always use MBCS

Note that we can still enable some internationalization with MBCS, within the limits of the selected system locale. I'm not 100% sure the best way to query the encoding, but `locale.getlocale()[1]` looks like a good starting point. The other methods are either problematic (`locale.getpreferredencoding()`) or require a recent Python.

You could add an encoding property to `DaqLibImporter`. If the UTF-8 DLL is available, set the encoding property to `"utf-8"`, otherwise set it to `locale.getlocale()[1]`.

> Do we want to add a simulated device with non ASCII device name so that it can be covered in some of the regression tests?

This would make sense, but I would prefer for this to be in a separate PR once we have confirmed that the DLL change doesn't break the internal test suite.

We also need to update the internal tests that I just added.

bkeryan (changes requested):

generated/nidaqmx/_lib.py
bkeryan: More thoughts on this:

If we support both DLLs (via automatic fallback), we should test with both DLLs. That means we need some sort of feature toggle to override the automatic fallback.

If we support only UTF-8 on Windows and MBCS on Linux, we don't need to test with both DLLs on either OS. However, if there are Windows systems with only the MBCS DLL installed, this would cause a regression on those systems. I don't know if there are systems like this. The UTF-8 DLL was used for LV NXG, so it's not particularly new.

A feature toggle could use registry on Windows and INI on Linux like our driver feature toggles do, or it could potentially use something simpler like environment variables or a `.env` file. The MeasurementLink Python module is using `.env` for feature toggles via the `decouple` module.

I don't know whether NI drivers officially support internationalization or localization on Linux. If it's not supported, it would be better to keep using ASCII on Linux and only support internationalization on Windows.

Modern OSes have better support for UTF-8 locales than when NI introduced the UTF-8 version of this DLL. Win10 has opt-in support for UTF-8 ANSI code page but this is not enabled by default. Most Linux distros default to locales with UTF-8 encodings. I don't know if the MBCS DLL supports these configurations. Even if the string conversions work with UTF-8 in this case, localized strings for errors and such may be in the wrong encoding.

generated/nidaqmx/_lib.py
bkeryan: Oops, I didn't scroll down to see that the Linux code is separate and you didn't change it.

generated/nidaqmx/_lib.py
```diff
@@ -46,13 +47,13 @@ def _setter(self, val):
 
 class CtypesByteString:
     """
-    Custom argtype that automatically converts unicode strings to ASCII
+    Custom argtype that automatically converts unicode strings to UTF-8
```
bkeryan, line 50: That's not correct, it converts to the encoding used by the DAQmx C API DLL.

generated/nidaqmx/_lib.py
```diff
@@ -89,6 +90,10 @@ def __init__(self, library):
         self._library = library
         self._lib_lock = threading.Lock()
 
+    @property
+    def library(self):
+        return str(self._library).split(' ')[1].strip("',")
```
bkeryan, line 95: I think you should simplify/optimize this. This work is duplicated by most API calls.

`DaqLibImporter._import_lib` can save the encoding at the time that the library is loaded.
- It could save the encoding as `self._encoding` in the `DaqLibImporter` class. Or as `self.encoding`, and then you don't need to write a `@property` method.
- Or it could save the encoding in the `DaqFunctionImporter` by passing it as another constructor parameter. However, I see no reason why we would mix and match DLLs. Both `DaqFunctionImporter` objects are for the same DLL; they just have different calling conventions.

bkeryan (changes requested):

generated/nidaqmx/_lib.py
```diff
@@ -156,43 +153,48 @@ def cal_handle(self) -> type:
...
+            self._import_lib()
+        return self._encoding
 
     def _import_lib(self):
         """
         Determines the location of and loads the NI-DAQmx CAI DLL.
         """
         self._windll = None
         self._cdll = None
+        self.encoding = None
```
bkeryan, line 166: `self.encoding` is a read-only property. This looks like a typo for `self._encoding`.

[functools.cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) has a tricky way of "overwriting" a property with a simple value but I think it uses `__dict__` to do the "overwrite".

generated/nidaqmx/_lib.py
bkeryan: > The UTF-8 DLL was used for LV NXG, so it's not particularly new.

Seeing the comment on `TaskHandle` about NI-DAQmx 8.9 made me wonder, do we still support any DAQmx versions that don't have `nicai_utf8.dll`? I looked at the history in version control and it was added in DAQmx 16.4, which is newer than I thought. This change as it currently is would break compatibility with DAQmx <16.4.

charitylxy: @bkeryan meaning we should let it load `nicai_utf8.dll` and if causes an error (let's say having DAQmx <16.4), then we'll fallback to `nicaiu.dll`? and this only applies to Windows since we're not supporting UTF-8 in Linux version.

bkeryan: Yes. That would preserve compatibility with DAQmx versions from 2016 and older.

Also, if we support both DLLs, we should test both DLLs. It's easier to test the fallback case if there is some sort of override.

maxxboehme (approved):

generated/nidaqmx/_lib.py
```diff
@@ -165,8 +165,8 @@ def _import_lib(self):
                     windll = ctypes.windll.nicaiu
                     cdll = ctypes.cdll.nicaiu
```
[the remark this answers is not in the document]
maxxboehme: I am not against a PR removing IronPython support.

generated/nidaqmx/_lib.py
```diff
@@ -6,6 +6,9 @@
 import platform
 import sys
 import threading
+import locale
+import os
+from dotenv import load_dotenv
```
bkeryan, line 11: As far as I know, `nidaqmx` doesn't depend on `dotenv`. If it does, that's purely by accident.

If this worked, it's because this project's dev dependencies happened to depend on `dotenv`. This code may break when users install `nidaqmx` in their own projects.

generated/nidaqmx/_lib.py
```diff
@@ -148,39 +152,67 @@ def task_handle(self) -> type:
...
         cdll = None
-
-        if sys.platform.startswith('win') or sys.platform.startswith('cli'):
-            try:
-                if 'iron' in platform.python_implementation().lower():
+        encoding= None
+
+        # Feature Toggle to load nicaiu.dll or nicai_utf8.dll
+        # The Feature Toggle can be set in the .env file
+        load_dotenv()
```
bkeryan, line 176: This modifies the process environment, which is not a nice thing for libraries to do at load time. Either use `dotenv_values()` or switch to `decouple`.

generated/nidaqmx/_lib.py
bkeryan, line 11: For MeasurementLink Python feature toggles, we are using `decouple`, not `dotenv`. @mshafer-NI recommended `decouple`, and I haven't evaluated `dotenv`.
https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_featuretoggles.py
https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_configuration.py

generated/nidaqmx/_lib.py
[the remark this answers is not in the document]
bkeryan: By the time IronPython catches up with CPython 3.8, perhaps these IronPython workarounds will be unnecessary.

.env
bkeryan, line 1: Don't check `.env` into version control. Add `.env` to `.gitignore` and check `.env.sample` into version control. 

Reason: `.env` is for local configuration and may contain values (IP addresses, passwords) that are not suitable to check into version control.

src/handwritten/_lib.py
```diff
@@ -148,39 +152,67 @@ def task_handle(self) -> type:
...
+        # Feature Toggle to load nicaiu.dll or nicai_utf8.dll
+        # The Feature Toggle can be set in the .env file
+        load_dotenv()
+        nicaiu_dll_enabled = os.getenv('NICAIU_DLL_ENABLED') == 'true'
+        
+        if sys.platform.startswith('win') or sys.platform.startswith('cli'):     
+            try: 
+                if nicaiu_dll_enabled:
                     windll = ctypes.windll.nicaiu
                     cdll = ctypes.cdll.nicaiu
```
bkeryan, line 183: This is based on the IronPython code path. Do we want that?

src/handwritten/_lib.py
```diff
@@ -148,39 +152,67 @@ def task_handle(self) -> type:
...
-
-        if sys.platform.startswith('win') or sys.platform.startswith('cli'):
-            try:
-                if 'iron' in platform.python_implementation().lower():
+        encoding= None
+
+        # Feature Toggle to load nicaiu.dll or nicai_utf8.dll
+        # The Feature Toggle can be set in the .env file
+        load_dotenv()
+        nicaiu_dll_enabled = os.getenv('NICAIU_DLL_ENABLED') == 'true'
```
bkeryan, line 177: Let's prefix the variable with `NIDAQMX_PYTHON` to make its origin/purpose clear.

Runtime feature toggles and `.env` variables don't necessarily have to be Booleans. 

Suggestion:
- Rename to `NIDAQMX_PYTHON_C_LIBRARY`
- Valid options:
  - Windows: `nicaiu`, `nicai_utf8`
  - Linux: `nidaqmx`
- If the variable is set, load that DLL with no fallback.

bkeryan (changes requested):

src/handwritten/_lib.py
```diff
@@ -148,39 +152,67 @@ def task_handle(self) -> type:
...
         cdll = None
-
-        if sys.platform.startswith('win') or sys.platform.startswith('cli'):
-            try:
-                if 'iron' in platform.python_implementation().lower():
+        encoding= None
+
+        # Feature Toggle to load nicaiu.dll or nicai_utf8.dll
+        # The Feature Toggle can be set in the .env file
+        load_dotenv()
```
bkeryan, line 176: FYI, in MeasurementLink Python I had to do extra work to find the `.env` file in a reasonable location for TestStand code modules and PyInstaller EXEs: https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_dotenvpath.py

As long as this nidaqmx-python feature toggle is only used for testing, I don't think we need it to be compatible with these deployment scenarios.

src/handwritten/_lib.py
```diff
@@ -148,39 +151,69 @@ def task_handle(self) -> type:
...
-            try:
-                if 'iron' in platform.python_implementation().lower():
-                    windll = ctypes.windll.nicaiu
-                    cdll = ctypes.cdll.nicaiu
-                else:
-                    windll = ctypes.windll.LoadLibrary('nicaiu')
-                    cdll = ctypes.cdll.LoadLibrary('nicaiu')
-            except (OSError, WindowsError) as e:
-                raise DaqNotFoundError(_DAQ_NOT_FOUND_MESSAGE) from e
+        encoding= None
```
zhindes, line 171: spacing

src/handwritten/_lib.py
charitylxy: @bkeryan I have made the necessary changes, however I didn't set `nidaqmx` as an option since we will be loading `nidaqmx` regardless the variable is set or not when running on Linux.

src/handwritten/_lib.py
```diff
@@ -148,39 +151,69 @@ def task_handle(self) -> type:
...
+                    if nidaqmx_c_library=="nicaiu":
+                        windll = ctypes.windll.LoadLibrary('nicaiu')
+                        cdll = ctypes.cdll.LoadLibrary('nicaiu')
+                        encoding = locale.getlocale()[1]
+                    elif nidaqmx_c_library=="nicai_utf8":
+                        windll = ctypes.windll.LoadLibrary('nicai_utf8')
+                        cdll = ctypes.cdll.LoadLibrary('nicai_utf8')
+                        encoding = 'utf-8'  
+                except (OSError, WindowsError) as e:
+                    raise DaqNotFoundError(_DAQ_NOT_FOUND_MESSAGE) from e
```
zhindes, line 189: lots of repetition in these checks and also missing the else clause if someone actually sets this. maybe something like:
```python
if nidaqmx_c_library is not None:
   try:
      if nidaqmx_c_library=="nicaiu":
         ...
      else if nidaqmx_c_library=="nicai_utf8":
         ...
      else:
         DaqNotFoundError(_DAQ_NOT_FOUND_MESSAGE)
```

src/handwritten/_lib.py
```diff
@@ -148,39 +151,69 @@ def task_handle(self) -> type:
...
+        # The Feature Toggle can be set in the .env file
+        nidaqmx_c_library = config('NIDAQMX_C_LIBRARY', cast=str, default=None)
+
+        if sys.platform.startswith('win') or sys.platform.startswith('cli'):   
+            if nidaqmx_c_library=="nicaiu" or nidaqmx_c_library=="nicai_utf8":
+                try: 
+                    if nidaqmx_c_library=="nicaiu":
+                        windll = ctypes.windll.LoadLibrary('nicaiu')
+                        cdll = ctypes.cdll.LoadLibrary('nicaiu')
+                        encoding = locale.getlocale()[1]
```
zhindes, line 183: You repeat this clause a few times... maybe make this a short internal function?

```python
def _load_lib(libname: str):
   windll = ctypes.windll.LoadLibrary(libname)
   cdll = ctypes.cdll.LoadLibrary(libname)
   return windll, cdll

...

windll, cdll = _load_lib('nicaiu')
encoding = locale.getlocale()[1]
```

zhindes (changes requested):

charitylxy:
Adding python-decouple dependency in https://github.com/ni/nidaqmx-python/pull/524

generated/nidaqmx/_lib.py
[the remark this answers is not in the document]
charitylxy: @bkeryan I have removed IronPython for simplicity, let me know if you prefer it being done in a separate PR.

zhindes (approved):

src/handwritten/_lib.py
```diff
@@ -148,39 +151,72 @@ def task_handle(self) -> type:
...
+            if nidaqmx_c_library is not None:
+                try: 
+                    if nidaqmx_c_library=="nicaiu":
+                        windll, cdll = _load_lib("nicaiu")
+                        encoding = locale.getlocale()[1]
+                    elif nidaqmx_c_library=="nicai_utf8":
+                        windll, cdll = _load_lib("nicai_utf8")
+                        encoding = 'utf-8'  
+                    else:
+                        raise DaqNotFoundError(_DAQ_NOT_FOUND_MESSAGE)
```
bkeryan, line 187: Setting the env var to an invalid value shouldn't return a DaqNotFoundError.

```suggestion
                        raise ValueError(f"Unsupported NIDAQMX_C_LIBRARY value: {nidaqmx_c_library}")
```

src/handwritten/_lib.py
```diff
@@ -148,39 +151,72 @@ def task_handle(self) -> type:
...
-            except (OSError, WindowsError) as e:
-                raise DaqNotFoundError(_DAQ_NOT_FOUND_MESSAGE) from e
+        encoding = None
+
+        # Feature Toggle to load nicaiu.dll or nicai_utf8.dll
+        # The Feature Toggle can be set in the .env file
+        nidaqmx_c_library = config('NIDAQMX_C_LIBRARY', cast=str, default=None)
+
+        if sys.platform.startswith('win') or sys.platform.startswith('cli'):   
+            if nidaqmx_c_library is not None:
```
bkeryan, line 178: If we're not going to look at the env var on Linux, then let's only query it on Windows.
```suggestion
        if sys.platform.startswith('win') or sys.platform.startswith('cli'):   
            # Feature Toggle to load nicaiu.dll or nicai_utf8.dll
            # The Feature Toggle can be set in the .env file
            nidaqmx_c_library = config('NIDAQMX_C_LIBRARY', cast=str, default=None)

            if nidaqmx_c_library is not None:
```

src/handwritten/_lib.py
```diff
@@ -148,39 +151,72 @@ def task_handle(self) -> type:
...
-                    cdll = ctypes.cdll.LoadLibrary('nicaiu')
-            except (OSError, WindowsError) as e:
-                raise DaqNotFoundError(_DAQ_NOT_FOUND_MESSAGE) from e
+        encoding = None
+
+        # Feature Toggle to load nicaiu.dll or nicai_utf8.dll
+        # The Feature Toggle can be set in the .env file
+        nidaqmx_c_library = config('NIDAQMX_C_LIBRARY', cast=str, default=None)
+
+        if sys.platform.startswith('win') or sys.platform.startswith('cli'):
```
bkeryan, line 177: cli == IronPython
```suggestion
        if sys.platform.startswith('win'):   
```

.env
bkeryan: This is not fixed

.env
```diff
@@ -0,0 +1 @@
+NIDAQMX_C_LIBRARY=None
```
bkeryan, line 1:
```suggestion
# This is a sample nidaqmx-python configuration file.
#
# To use it:
# - Copy this file to your application's directory or one of its parent directories
#   (such as the root of your Git repository).
# - Rename it to `.env`.
# - Uncomment and edit the options you want to change.
# - Restart any affected applications or services.

# By default, nidaqmx-python on Windows uses nicai_utf8, the UTF-8 version of the NI-DAQmx C library.
# If that is not available, it falls back to nicaiu, the MBCS (multibyte character set) version. You can override
# this behavior by uncommenting the following option:

# NIDAQMX_C_LIBRARY=nicaiu
```

.env
bkeryan, line 1: Rename this file to `.env.sample`

You may need to temporarily remove it from `.gitignore` before you can do this.

generated/nidaqmx/_lib.py
```diff
@@ -148,39 +151,72 @@ def task_handle(self) -> type:
...
             raise DaqNotSupportedError(_DAQ_NOT_SUPPORTED_MESSAGE.format(sys.platform))
 
         self._windll = DaqFunctionImporter(windll)
         self._cdll = DaqFunctionImporter(cdll)
+        self._encoding = encoding
+
+        def _load_lib(libname: str):
+            windll = ctypes.windll.LoadLibrary(libname)
+            cdll = ctypes.cdll.LoadLibrary(libname)
+            return windll, cdll
```
bkeryan, line 219: This is only used inside the `if sys.platform.startswith('win')` block, so please move it there.

bkeryan (changes requested):
FYI, this PR looks good but the rename to `.env.sample` slipped through the cracks. We should also document it as if it is a sample. Hopefully this will be the last round of changes.

src/handwritten/_lib.py
bkeryan: ok, that's fine

bkeryan (approved):

PR #521, opened 2024-02-26
Add tzlocal dependency in poetry.lock

DeborahOoi96:

Trying to add tzlocal dependency in daqmx-python because it is needed in #504

Testing on PR build


pyproject.toml
bkeryan, line 1: https://stackoverflow.com/questions/76583100/is-pytz-deprecated-now-or-in-the-future-in-python

Do you really need `pytz`?

Python 3.9 adds [zoneinfo](https://docs.python.org/3/library/zoneinfo.html).

For Python 3.8, we can use [backports.zoneinfo](https://pypi.org/project/backports.zoneinfo/). It looks like tzlocal will pull it in automatically: https://github.com/regebro/tzlocal/blob/master/pyproject.toml

DeborahOoi96: I have removed pytz and used zoneinfo as per your comment in the libtime PR. Reverting this change.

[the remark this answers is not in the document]
DeborahOoi96: Added support for tzlocal and updated description

pyproject.toml
```diff
@@ -45,6 +45,7 @@ sphinx_rtd_theme = {version=">=1.0", optional=true}
 grpcio = {version=">=1.49.0,<2.0", optional = true}
 protobuf = { version = "^4.21", optional = true }
 hightime = "^0.2.1"
+tzlocal = "^5.2"
```
bkeryan, line 48: `^5.2` means `>=5.2,<6.0`. If you want to use `nidaqmx` with `tzlocal` 5.1, too bad, you're not allowed. Likewise, if you want to use `nidaqmx` with `tzlocal` 6.0 when it is released, you can't.

So I would normally recommend specifying a looser constraint like `>=5.0`.

Looking at `tzlocal`'s version history to see how far back we can go, I see it broke compatibility in 3.0, reverted it in 4.0, and broke it again in 5.0.
https://github.com/regebro/tzlocal?tab=readme-ov-file#api-change

So I guess it's safe to assume that a `tzlocal` 6.0 would break compatibility, so we should use `^5.0`

poetry.lock
```diff
@@ -932,766 +1681,6 @@ docs = ["Sphinx", "sphinx_rtd_theme"]
 grpc = ["grpcio", "protobuf"]
 
 [metadata]
-lock-version = "1.1"
+lock-version = "2.0"
```
bkeryan, line 1684: I guess this is ok since the repo's GitHub workflows use Poetry 1.4, which supports the new lock format.

bkeryan (approved):
Approved with suggestion: set minimum version to 5.0 not 5.2

pyproject.toml
DeborahOoi96: Reverted my tzlocal 5.2 change, updated it to be ^5.0 and reran poetry.lock.

PR #523, opened 2024-03-01
nidaqmx: Fix mypy errors

bkeryan:

Enable running mypy on the actual `nidaqmx` package without errors. For details, see commits.

Update PR build workflow to run mypy on `generated/nidaqmx`.

Even without type hints, running mypy on the `nidaqmx` package can catch bugs, such as passing the wrong arguments to a function.

Running mypy on the `nidaqmx` package is a prerequisite for adding type hints.

> **Note**
>
> Adding type hints will force us to stop reassigning to parameters, like `val = numpy.array(val, dtype=numpy.float64)`. These currently pass because they are untyped and thus are assumed to be `typing.Any`.

Ran `poetry run pytest -v`
Ran `poetry run mypy generated/nidaqmx` and `poetry run mypy tests`


generated/nidaqmx/task.py
```diff
@@ -759,6 +677,102 @@ def read(self, number_of_samples_per_channel=NUM_SAMPLES_UNSET,
...
+                ]
+            else:
+                # n channel, n samples
+                return [
+                    [
+                        PowerMeasurement(voltage=v, current=i)
+                        for v, i in zip(voltages[channel_index], currents[channel_index])
+                    ]
+                    for channel_index in range(number_of_channels)
+                ]
```
bkeryan, line 765: Shouldn't this limit to `[:samples_read]`?

zhindes (approved):
lots of great improvements!

generated/nidaqmx/task.py
bkeryan: Good news: I added a test for this. https://github.com/ni/nidaqmx-python/pull/530

Bad news: I found two other issues besides this one.
- #527 
- #528 
- #529 
@zhindes @hongloonwong

Issue #461, opened 2023-12-11
Make user-visible types public to support type hints


bkeryan:
The task subobject types (AIChannel, Timing, etc.) are defined in internal modules contained in the private `nidaqmx._task_modules` package. When users write helper functions that accept channel objects or other task subobjects, they must refer to these internal modules. We should make these modules public and/or add public aliases for their types.

@StSav012 brought this up in https://github.com/ni/nidaqmx-python/pull/458

bkeryan:
I think we should turn `task.py` into a package, merge it with `_task_modules`, make most of the submodules private, and add aliases to reexport types from the subpackages.

Something like this:
```
nidaqmx/
  task/
    __init__.py - defines Task; reexports ExportSignals, InStream, OutStream, Timing, Triggers
    channels/
      __init__.py - reexports Channel, AIChannel, AOChannel, etc.
      _channel.py - defines Channel
      _ai_channel.py - defines AIChannel
      ...
    collections
      __init__.py - reexports ChannelCollection, AIChannelCollection, AOChannelCollection, etc.
      _channel_collection.py - defines ChannelCollection
      _ai_channel_collection.py - defines AIChannelCollection
      ...
    _export_signals.py - defines ExportSignals
    _in_stream.py - defines InStream
    _out_stream.py - defines OutStream
    _timing.py - defines Timing
    _triggers.py - defines Triggers
    triggering/
      __init_.py - reexports triggering classes
      _arm_start_trigger.py - defines ArmStartTrigger
      ...
```

The public types would be:
```
nidaqmx.task.Task
nidaqmx.task.ExportSignals
nidaqmx.task.InStream
nidaqmx.task.OutStream
nidaqmx.task.Timing
nidaqmx.task.Triggers
nidaqmx.task.channels.Channel, AIChannel, etc.
nidaqmx.task.collections.ChannelCollection, AIChannelCollection, etc.
nidaqmx.task.triggering.ArmStartTrigger, HandshakeTrigger, etc.
```

charitylxy:
Completed refactoring work for task and its submodules, documentation is updated accordingly. Closing this issue.

bkeryan:
Some of the types in nidaqmx.system are not public yet. Also, the docs refer to nonexistent modules like nidaqmx.system.collections:
https://nidaqmx-python.readthedocs.io/en/latest/collections.html

(This is really nidaqmx.system._collections, which is private.)

charitylxy:
[...]

Issue #46, opened 2017-12-14
stream_writers.write_one_sample : auto_start arg ?


MaxP92:
Hello,

In [read the docs](http://nidaqmx-python.readthedocs.io/en/latest/stream_writers.html) about stream_writers, it says 'write_one_sample' has the optional 'auto_start' argument, which is false

Also the def description in stream_writers.py indicates it

MaxP92:
Actually the whole stream_writers.write_one_sample is bad : it requires a numpy array even if one element has to be written

Also with 2 channels AO, it raises an error 'No. of dimensions of NumPy Array required: 1' when passing a (2,1) data array, which is yet the correct numpy shape for 2 channels
(with a (1,1) array, it raises the error DAQmxErrors.WRITE_NUM_CHANS_MISMATCH)

brianlee1200:
Hi Max,

For the `auto_start` parameter, it does not say the default value is `False`. The docs specify:

> If you do not specify a value for this parameter, NI-DAQmx determines its value based on the type of write method used. If you use a one sample write method, its value is True; conversely, if you use a many sample write method, its value is False.

Please elaborate on your second problem, or post an example code. The `write_one_sample` method for a multi channel writer will always require a NumPy array, even if there is only one channel in the task. Have you tried using a single channel writer?

There are some examples/unit tests here for your reference:
https://github.com/ni/nidaqmx-python/blob/master/nidaqmx/tests/test_stream_analog_readers_writers.py
https://github.com/ni/nidaqmx-python/blob/master/nidaqmx/tests/test_stream_counter_readers_writers.py
https://github.com/ni/nidaqmx-python/blob/master/nidaqmx/tests/test_stream_digital_readers_writers.py

Please let us know if you still have any issues.

neilvana:
Is this still an issue?  Can I close this?

MaxP92:
no, 'write_one_sample' has not the optional 'auto_start' argument

PR #534, opened 2024-03-07
Refactor task.py and internal module (in_stream.py) 

charitylxy:

This PR  is to address issue brought up in https://github.com/ni/nidaqmx-python/issues/461
- Currently, task subobject types are defined in internal modules contained in private `nidaqmx._task_modules` package.
- We want to turn `task.py` into a package and merge it with `_task_modules`

_**Note to Reviewers:** This is the initial stage of implementation to ensure the approach is correct before we continue moving other task subobjects to  `task` package._
In this PR to address https://github.com/ni/nidaqmx-python/issues/46, following changes have been made: 
-  Turned `task.py` into a package.
- Made `in_stream.py`  into private submodule `_in_stream.py` and moved to `task` package.
- Added type hints for `in_stream` and update documentation.

Ran `poetry run pytest` and all tests passed.


generated/nidaqmx/task/task.py
bkeryan, line 1: This makes the `nidaqmx.task.task` submodule public. Rename it to `_task.py` so that `nidaqmx.task._task` is private.

generated/nidaqmx/task/__init__.py
bkeryan, line 1: Set `__all__` to make it crystal-clear to both humans and automated tools that the re-exported classes are public. (At least the ones that don't start with underscores.)

src/codegen/templates/task/task.py.mako
bkeryan, line 1: Why turn task.py into a mako template? I think it's fine to keep it hand-written. None of the helpers you are importing are used anywhere.

src/codegen/templates/task/task.py.mako
```diff
@@ -242,9 +251,9 @@ def export_signals(self):
         return self._export_signals
 
     @property
-    def in_stream(self):
+    def in_stream(self) -> InStream: 
         """
-        :class:`nidaqmx._task_modules.in_stream.InStream`: Gets the read
+        :class:`nidaqmx.task._in_stream.InStream`: Gets the read
```
bkeryan, line 256: I don't like how this codebase has been using private types in docstrings up to now.

`_in_stream` is currently a private submodule, so docstrings should use the alias: `:class:``nidaqmx.task.InStream``` (ignore markdown mess) in docstrings. This might not work until you set `__all__`.

If you make `nidaqmx.task.in_stream` public, then it's ok to use it in docstrings, but that makes the docstrings more verbose.

docs/in_stream.rst
```diff
@@ -1,6 +1,6 @@
 nidaqmx.task.in_stream
 ======================
 
-.. automodule:: nidaqmx._task_modules.in_stream
+.. automodule:: nidaqmx.task._in_stream
```
bkeryan, line 4: This is confusing.

The file says `nidaqmx.task.in_stream`, but that module does not exist. (It never did, so this was not a new problem.)

The actual help says `nidaqmx.task._in_stream.InStream`, which is correct, but it is a private submodule.

I think we should do one of the following:
- Make `nidaqmx.task.in_stream` public to match the docs
- Use [:canonical:](https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-option-py-function-canonical) to make the docs match the public alias.

Even if you make `in_stream` public, we still need to solve this for `Task`, and `:canonical:` may help.

I can't find the `Task` class on https://nidaqmx-python--534.org.readthedocs.build/en/534/task.html

src/codegen/templates/task/task.py.mako
bkeryan, line 1: FYI, it is also possible to put the actual code in `task/__init__.py`. But I think aliases are fine if we make the docs use them instead of the private names.

bkeryan (changes requested):

generated/nidaqmx/task/task.py
charitylxy: @bkeryan Just to clarify, meaning `nidaqmx.task.task` is supposed to be a private submodule, while the remaining submodules like `nidaqmx.task._in_stream`, `nidaqmx.task._export_signals` etc. will be public? And what about the files in `channels`, `triggering` and `collection`, should they be private or public?

bkeryan: I'm following [PEP8's guideline](https://peps.python.org/pep-0008/) that a leading underscore is a weak "internal use" indicator.

`nidaqmx.task.task` doesn't have a leading underscore, so it is public.
`nidaqmx.task._task` has a leading underscore, so it is private.

I don't like `nidaqmx.task.task` because it is redundant (and `nidaqmx.task.task.Task` is even worse). 

Whether the other submodules should be public or private is an entirely separate question and I wrote a proposed layout in https://github.com/ni/nidaqmx-python/issues/461#issuecomment-1850854948

docs/in_stream.rst
charitylxy: @bkeryan I've removed `in_stream.rst` since setting `__all__ = ['Task', 'InStream']` will allow `InStream` to be available as `nidaqmx.task.InStream`. Documentation looks something like this:

image

charitylxy:
Hi @bkeryan, wondering if you have the time to take a look at this PR? I've made the changes you requested.

bkeryan (approved):

PR #535, opened 2024-03-07
Update cfg analog multi edge trig APIs

WayneDroid:

Updated `CfgAnlgMultiEdgeRefTrig` and `CfgAnlgMultiEdgeStartTrig` metadata and generated the APIs.

Enable `CfgAnlgMultiEdgeRefTrig` and `CfgAnlgMultiEdgeStartTrig` APIs for the 1.0 release.

Add test cases:
- [x] `cfg_anlg_edge_start_trig`
- [x] `cfg_anlg_edge_ref_trig`
- [x] `CfgAnlgMultiEdgeRefTrig`
- [x] `CfgAnlgMultiEdgeStartTrig`

image
image
image
image

### DIFF
--- a/generated/nidaqmx/_base_interpreter.py
+++ b/generated/nidaqmx/_base_interpreter.py
@@ -65,8 +65,8 @@ class BaseInterpreter(abc.ABC):
 
     @abc.abstractmethod
     def cfg_anlg_multi_edge_ref_trig(
-            self, task, trigger_sources, trigger_slope_array,
-            trigger_level_array, pretrigger_samples):
+            self, task, trigger_sources, pretrigger_samples,
+            trigger_slope_array, trigger_level_array):
         raise NotImplementedError
 
     @abc.abstractmethod
--- a/generated/nidaqmx/_task_modules/triggering/start_trigger.py
+++ b/generated/nidaqmx/_task_modules/triggering/start_trigger.py
@@ -1016,8 +1016,8 @@ class StartTrigger:
         if trigger_level_array is None:
             trigger_level_array = []
 
-        trigger_slope_array = numpy.int32([p.value for p in trigger_slope_array])
-        trigger_level_array = numpy.float64(trigger_level_array)
+        trigger_slope_array = numpy.array([p.value for p in trigger_slope_array], dtype=numpy.int32)
+        trigger_level_array = numpy.array(trigger_level_array, dtype=numpy.float64)
 
 
         self._interpreter.cfg_anlg_multi_edge_start_trig(
--- a/generated/nidaqmx/_task_modules/triggering/start_trigger.py
+++ b/generated/nidaqmx/_task_modules/triggering/start_trigger.py
@@ -987,6 +987,43 @@ class StartTrigger:
         self._interpreter.cfg_anlg_edge_start_trig(
             self._handle, trigger_source, trigger_slope.value, trigger_level)
 
+    def cfg_anlg_multi_edge_start_trig(
+            self, trigger_sources, trigger_slope_array,
+            trigger_level_array=None):
+        """
+        Configures the task to start acquiring or generating samples
+        when any of the configured analog signals cross the respective
+        levels you specified. Multi-edge triggering treats the
+        configured triggers as if a logical OR is applied.
+
+        Args:
+            trigger_sources (str): Is the name of a virtual channel or
+                terminal where there is an analog signal to use as the
+                source of the trigger.
+            trigger_slope_array (List[nidaqmx.constants.Slope]): 
+                Specifies on which slope of the signal to start
+                acquiring or generating samples when the signal crosses
+                **level**.
+            trigger_level_array (Optional[List[float]]): Specifies at
+                what threshold to start acquiring or generating samples.
+                Specify this value in the units of the measurement or
+                generation. Use **slope** to specify on which slope to
+                trigger at this threshold.
+        """
+        if trigger_slope_array is None:
+            trigger_slope_array = []
+
+        if trigger_level_array is None:
+            trigger_level_array = []
+
+        trigger_slope_array = numpy.int32([p.value for p in trigger_slope_array])
+        trigger_level_array = numpy.float64(trigger_level_array)
+
+
+        self._interpreter.cfg_anlg_multi_edge_start_trig(
+            self._handle, trigger_sources, trigger_slope_array,
+            trigger_level_array)
+
     def cfg_anlg_window_start_trig(
             self, window_top, window_bottom, trigger_source="",
             trigger_when=WindowTriggerCondition1.ENTERING_WINDOW):
--- a/tests/component/_task_modules/test_triggers.py
+++ b/tests/component/_task_modules/test_triggers.py
@@ -156,19 +156,20 @@ def test___start_trigger___cfg_anlg_edge_start_trig___no_errors(
     trig_level: float,
 ):
     device_name = sim_6363_ai_voltage_task.devices[0].name
+
     sim_6363_ai_voltage_task.timing.cfg_samp_clk_timing(1000)
     sim_6363_ai_voltage_task.triggers.start_trigger.cfg_anlg_edge_start_trig(
         trigger_source=trig_src, trigger_slope=trig_slope, trigger_level=trig_level
     )
-
-    sim_6363_ai_voltage_task.start()
 
     assert (
         sim_6363_ai_voltage_task.triggers.start_trigger.anlg_edge_src
         == f"/{device_name}/{trig_src}"
     )
     assert sim_6363_ai_voltage_task.triggers.start_trigger.anlg_edge_slope == trig_slope
-    assert round(sim_6363_ai_voltage_task.triggers.start_trigger.anlg_edge_lvl, 1) == trig_level
+    assert sim_6363_ai_voltage_task.triggers.start_trigger.anlg_edge_lvl == pytest.approx(
+        trig_level, abs=0.001
+    )
 
 
 @pytest.mark.parametrize(
@@ -195,8 +196,6 @@ def test___start_trigger___cfg_anlg_multi_edge_start_trig___no_errors(
         trigger_level_array=trig_levels,
     )
 
-    sim_9775_ai_voltage_multi_edge_task.start()
-
     assert (
         sim_9775_ai_voltage_multi_edge_task.triggers.start_trigger.anlg_multi_edge_srcs
         == flatten_trigger_sources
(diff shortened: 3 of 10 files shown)


zhindes (approved):
are tests coming in this PR or another one?

WayneDroid:
> are tests coming in this PR or another one?

tests will be added in this PR.

generated/nidaqmx/_base_interpreter.py
```diff
@@ -65,8 +65,8 @@ def cfg_anlg_edge_start_trig(
 
     @abc.abstractmethod
     def cfg_anlg_multi_edge_ref_trig(
-            self, task, trigger_sources, trigger_slope_array,
-            trigger_level_array, pretrigger_samples):
+            self, task, trigger_sources, pretrigger_samples,
+            trigger_slope_array, trigger_level_array):
```
bkeryan, line 69: It's kind of weird that the position of pretrigger_samples in the parameter list changed, but I think it's consistent between all 3 interpreter classes and the caller.

@zhindes Do the changes you proposed affect this too?

zhindes: I think is because of what I found, yes. The C interface is this:

```C
 int32 __CFUNC     DAQmxCfgAnlgMultiEdgeRefTrig   (TaskHandle taskHandle, const char triggerSources[], const int32 triggerSlopeArray[], const float64 triggerLevelArray[], uInt32 pretriggerSamples, uInt32 arraySize); 
```

Notice that the arraySize doesn't follow the "cluster" of arrays. It looks like all of this Python codegen actually handles that correctly (the code _library_interpreter.py looks correct).

```Python
        cfunc = lib_importer.windll.DAQmxCfgAnlgMultiEdgeRefTrig
        if cfunc.argtypes is None:
            with cfunc.arglock:
                if cfunc.argtypes is None:
                    cfunc.argtypes = [
                        lib_importer.task_handle, ctypes_byte_str,
                        wrapped_ndpointer(dtype=numpy.int32, flags=('C')),
                        wrapped_ndpointer(dtype=numpy.float64, flags=('C')),
                        ctypes.c_uint32, ctypes.c_uint]

        error_code = cfunc(
            task, trigger_sources, trigger_slope_array, trigger_level_array,
            pretrigger_samples, len(trigger_level_array))
```

However, I think it caused this oddity in the codegen of the interpreter signatures. I will be adding `DAQmxCfgAlgMultiEdgeRefTrigEx` which is more sane. I think if we used that, we could make this a bit less confusing, but it doesn't look like we need to. I'll bubble my changes up through scrapigen and here just to make sure I didn't break anything unexpectedly.

generated/nidaqmx/_task_modules/triggering/start_trigger.py
```diff
@@ -987,6 +987,43 @@ def cfg_anlg_edge_start_trig(
...
+
+        if trigger_level_array is None:
+            trigger_level_array = []
+
+        trigger_slope_array = numpy.array([p.value for p in trigger_slope_array], dtype=numpy.int32)
+        trigger_level_array = numpy.array(trigger_level_array, dtype=numpy.float64)
+
+
+        self._interpreter.cfg_anlg_multi_edge_start_trig(
+            self._handle, trigger_sources, trigger_slope_array,
```
WayneDroid, line 1024: @bkeryan @zhindes 
I need advice on the `trigger_sources`. In the test [test___start_trigger___cfg_anlg_multi_edge_start_trig___no_errors](https://github.com/ni/nidaqmx-python/pull/535/files#diff-05b936155dab8db8b3bc0dfce153a40c0021599197bbba8487a4b0ec5e4d6872R154), giving a list of str for  trigger_sources will get type error. By referring [set_analog_power_up_states_with_output_type](https://github.com/ni/nidaqmx-python/blob/0f65354421c5de8799eef8e1343db6f16af95524/generated/nidaqmx/system/system.py#L405), I have to provide `flatten_trigger_sources`.

Should we set `'python_codegen_method': 'custom'` to `CfgAnlgMultiEdgeStartTrig` metadata and then define the function in mako file like `set_analog_power_up_states_with_output_type`? So that we can process `trigger_sources` before passing it to interpreter.

zhindes: An alternative implementation for all of this would be to create a cluster of `{ trigger_source: string, trigger_slope: enum, trigger_level: float}` and accept an array of those, just like the power up states. However, this is a very niche API only supported by one device. We don't need to go above and beyond for this. Forcing the user to call `utils.flatten_channel_string` is fine by me. If, in the future, we really cared, we could add a new Python API that does that, but I guarantee you we won't do that :)

zhindes: Note: @bkeryan just noticed that we have precedence for exactly what you have implemented in .NET.
* [StartTrigger.ConfigureAnalogMultiEdgeTrigger](https://www.ni.com/docs/en-US/bundle/ni-daqmx-net-framework-45-api-ref/page/ninetdaqmxfx45ref/html/m_nationalinstruments_daqmx_starttrigger_configureanalogmultiedgetrigger.htm)
* [ReferenceTrigger.ConfigureAnalogMultiEdgeTrigger](https://www.ni.com/docs/en-US/bundle/ni-daqmx-net-framework-45-api-ref/page/ninetdaqmxfx45ref/html/m_nationalinstruments_daqmx_referencetrigger_configureanalogmultiedgetrigger.htm)

Those methods require the user to flatten the sources.

WayneDroid:
> are tests coming in this PR or another one?

test cases are added.

zhindes (changes requested):

tests/component/_task_modules/test_triggers.py
```diff
@@ -121,3 +137,66 @@ def test___timestamp_not_enabled___wait_for_valid_timestamp___throw_error(
...
+def test___start_trigger___cfg_anlg_edge_start_trig___no_errors(
+    sim_6363_ai_voltage_task: Task,
+):
+    sim_6363_ai_voltage_task.timing.cfg_samp_clk_timing(1000)
+    sim_6363_ai_voltage_task.triggers.start_trigger.cfg_anlg_edge_start_trig(
+        trigger_source="APFI0", trigger_slope=Slope.RISING, trigger_level=2.0
+    )
+
+    sim_6363_ai_voltage_task.start()
+    sim_6363_ai_voltage_task.read()
```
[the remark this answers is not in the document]
zhindes: disregard this, I don't think we need to read at all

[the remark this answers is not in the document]
WayneDroid: Thanks for the inputs. I'll make the changes in next commit.

tests/component/_task_modules/test_triggers.py
```diff
@@ -24,6 +25,21 @@ def ci_count_edges_task(task, sim_9185_device) -> nidaqmx.Task:
     return task
 
 
+@pytest.fixture()
+def sim_6363_ai_voltage_task(task, sim_6363_device):
+    """Gets AI voltage task."""
+    task.ai_channels.add_ai_voltage_chan(sim_6363_device.ai_physical_chans[0].name)
+    yield task
```
bkeryan, line 32: Use `yield` when the fixture needs to do cleanup. In this case, the `task` fixture already does the cleanup, so use `return task`.

A yield fixtures would have a return type hint of `-> Generator[Task, None, None]`. For return, do `-> Task`.

tests/component/_task_modules/test_triggers.py
```diff
@@ -139,64 +140,154 @@ def test___timestamp_not_enabled___wait_for_valid_timestamp___throw_error(
...
+    trig_level: float,
 ):
+    device_name = sim_6363_ai_voltage_task.devices[0].name
     sim_6363_ai_voltage_task.timing.cfg_samp_clk_timing(1000)
     sim_6363_ai_voltage_task.triggers.start_trigger.cfg_anlg_edge_start_trig(
-        trigger_source="APFI0", trigger_slope=Slope.RISING, trigger_level=2.0
+        trigger_source=trig_src, trigger_slope=trig_slope, trigger_level=trig_level
     )
 
     sim_6363_ai_voltage_task.start()
```
zhindes, line 164: You don't need this

tests/component/_task_modules/test_triggers.py
```diff
@@ -139,64 +140,154 @@ def test___timestamp_not_enabled___wait_for_valid_timestamp___throw_error(
...
-    sim_6363_ai_voltage_task.read()
-
 
+    assert (
+        sim_6363_ai_voltage_task.triggers.start_trigger.anlg_edge_src
+        == f"/{device_name}/{trig_src}"
+    )
+    assert sim_6363_ai_voltage_task.triggers.start_trigger.anlg_edge_slope == trig_slope
+    assert round(sim_6363_ai_voltage_task.triggers.start_trigger.anlg_edge_lvl, 1) == trig_level
```
zhindes, line 171: Try `pytest.approx`. The default behavior is pretty strict, so you may want to pass `abs=0.1` or something to make it more granular

```suggestion
    assert sim_6363_ai_voltage_task.triggers.start_trigger.anlg_edge_lvl == pytest.approx(trig_level)
```

tests/component/_task_modules/test_triggers.py
```diff
@@ -139,64 +140,154 @@ def test___timestamp_not_enabled___wait_for_valid_timestamp___throw_error(
...
+        == flatten_trigger_sources
+    )
+    assert (
+        sim_9775_ai_voltage_multi_edge_task.triggers.start_trigger.anlg_multi_edge_slopes
+        == trig_slopes
+    )
+    assert [
+        round(v, 1)
+        for v in sim_9775_ai_voltage_multi_edge_task.triggers.start_trigger.anlg_multi_edge_lvls
+    ] == trig_levels
```
zhindes, line 211: approx works for arrays, too!

```suggestion
    assert sim_9775_ai_voltage_multi_edge_task.triggers.start_trigger.anlg_multi_edge_lvls == pytest.approx(trig_levels)
```

zhindes (changes requested):
my comments apply to all the new methods

tests/component/_task_modules/test_triggers.py
WayneDroid: Noted, it's `.start()` is removed for new cases.

tests/component/_task_modules/test_triggers.py
WayneDroid: Thanks a lot, it makes the test case simpler.

tests/component/_task_modules/test_triggers.py
WayneDroid: Test case is updated as per suggestion.

zhindes (approved):

WayneDroid:
@zhindes can this PR go in or do I need to wait for the actual fix on the metadata?

zhindes:
> @zhindes can this PR go in or do I need to wait for the actual fix on the metadata?

It can go in, that's gonna be a slow roll.

Issue #26, opened 2017-08-18
Make TDMS file paths use pathlib.Path


epage:
Good
- pathlib is nice to work with

Bad
- Didn't exist or was provisional until Python 3.6
  - The [pypi version](https://pypi.python.org/pypi/pathlib/) might be compatible enough as a backport?

bkeryan:
I think the minimum fix is to update these API methods to accept either a `str` or a `pathlib.PurePath`:
- `InStream.logging_file_path` setter
- `InStream.start_new_file()`
- `PhysicalChannel.configure_teds()`

I think this would be a nice-to-have, but it would break compatibility. If we don't do it in 1.0, we might not get another chance:
- `InStream.logging_file_path` getter returns `pathlib.Path`

PR #539, opened 2024-03-14
Refactor Task Internal Modules (ExportSignal, OutStream, Timing)

charitylxy:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).

### What does this Pull Request accomplish?
This PR aims to address issue brought up in https://github.com/ni/nidaqmx-python/issues/461:
- Currently, task subobject types are defined in internal modules contained in private `nidaqmx._task_modules` package.
- We want to turn `task.py` into a package and merge it with the internal modules contained in ` nidaqmx._task_modules`

Refactoring work will be split into 5 PRs:
- [x] Refactor `_task.py` and `_in_stream.py` -  https://github.com/ni/nidaqmx-python/pull/534
- [x] **Refactor `_export_signal.py`, `_out_stream.py`, `_timing.py` - Implemented in this PR**
- [ ] Refactor collections: `_channel_collection.py`, `_ai_channel_collection.py`, `_ao_channel_collection.py`, `_ci_channel_collection.py`, `_co_channel_collection.py`, `_di_channel_collection.py`, `_do_channel_collection.py`
- [ ] Refactor channels: `_channel.py`, `_ai_channel.py`, `_ao_channel.py`, `_ci_channel.py`, `_co_channel.py`, `_di_channel.py`, `_do_channel.py` 
- [ ] Refactor triggering: `_triggers.py`, `_arm_start_trigger.py`, `_handshake_trigger.py`, `_pause_trigger.py`, `_reference_trigger.py`, `_start_trigger.py`

### Why should this Pull Request be merged?
In this PR to address https://github.com/ni/nidaqmx-python/issues/461, following changes have been made: 
- Made `export_signal.py`, `out_stream.py` and `timing.py`  into private submodule and moved them to `task` package.
- Added type hints for `export_signal`, `out_stream.` and `timing.`.
- Documentation updated.

### What testing has been done?
Ran `poetry run pytest` and all tests passed.

PR #542, opened 2024-03-15
Add pathlib.PurePath type for file path

WayneDroid:

This PR is added file path handling for parameter and return type with existing metadata.
- added type hint `pathlib.PurePath` for `file_path`.
- added  `-> pathlib.PurePath` for file path return type.
- added and updated test cases for impacted functions.

This PR is to address #26.

Run test cases on impacted functions.

image
image
image
image
image
image


generated/nidaqmx/system/physical_channel.py
```diff
@@ -619,15 +620,18 @@ def configure_teds(self, file_path=""):
...
-                this function attempts to find a TEDS sensor connected
-                to the physical channel.
+            file_path (Optional[pathlib.PurePath]): Is the path to a
+                Virtual TEDS data sheet that you want to associate with
+                the physical channel. If you do not specify anything for
+                this input, this function attempts to find a TEDS sensor
+                connected to the physical channel.
         """
+        if not file_path.name:
+            file_path = "" # type: ignore
```
bkeryan, line 630: Is this about how `str(pathlib.PurePath(""))` evaluates to `"."`? That suggests that the default should be `None` rather than `pathlib.PurePath("")`.

Use `Optional[T]` in the type hint to indicate that the parameter is allowed to be `None`. 

Note that this is completely orthogonal to whether the parameter uses default args. `def f(x: int = 3)` uses default args but does not allow `None`. `def f(x: Optional[int])` allows `None` but does not use default args.

For `Union[X, Y]`, you can use either `Optional[Union[X, Y]]` or `Union[X, Y, None]`. They should be equivalent.

generated/nidaqmx/_task_modules/in_stream.py
```diff
@@ -330,22 +331,23 @@ def input_onbrd_buf_size(self):
...
+            task is running, this takes effect on the next sample
+            interval (if Logging.SampsPerFile has been set) or when
+            DAQmx Start New File is called. New file paths can be
+            specified by ending with "\\" or "/". Files created after
+            specifying a new file path retain the same name and
+            numbering sequence.
         """
 
         val = self._interpreter.get_read_attribute_string(self._handle, 0x2ec4)
         return val
```
bkeryan, line 347: The interpreter returns a `str`, so this layer is responsible for converting to `pathlib.Path`, just like when we return an enum.

generated/nidaqmx/_task_modules/in_stream.py
```diff
@@ -330,22 +331,23 @@ def input_onbrd_buf_size(self):
...
 
     @property
-    def logging_file_path(self):
+    def logging_file_path(self) -> pathlib.PurePath:
```
bkeryan, line 334: My suggestion was to return `pathlib.Path` here: 
> InStream.logging_file_path getter returns pathlib.Path

The difference is that `pathlib.Path` has methods that support filesystem access and `pathlib.PurePath` omits these methods.

The most versatile approach is to accept `PurePath`, but return `Path`.

src/codegen/templates/function_template.py.mako
```diff
@@ -52,12 +54,17 @@
...
+## Script default values for parameters that are lists or file path,
+## since default values in Python functions should not be mutable.
     %for input_param in func.parameters:
         %if input_param.is_list:
         if ${input_param.parameter_name} is None:
             ${input_param.parameter_name} = ${get_list_default_value(func, input_param)}
 
+        %elif is_path_type(input_param):
+        if not ${input_param.parameter_name}.name:
+            ${input_param.parameter_name} = "" # type: ignore
```
bkeryan, line 66: I think you should rewrite this to not need `# type: ignore`. For example, if the parameter type is `Union[str, pathlib.PurePath, None]`, then you can do this without `# type: ignore`:

```
if parameter is None:
    parameter = ""
```

When you **do** need `# type: ignore`, use `# type: ignore[category]` to document what you are ignoring.

bkeryan (changes requested):

zhindes (changes requested):
brad's comments

generated/nidaqmx/system/physical_channel.py
```diff
@@ -609,7 +610,7 @@ def clear_teds(self):
         self._interpreter.clear_teds(
             self._name)
 
-    def configure_teds(self, file_path=""):
+    def configure_teds(self, file_path: pathlib.PurePath=pathlib.PurePath("")):
```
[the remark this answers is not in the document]
WayneDroid: Thanks for clarification, it's changed to `Optional[Union[str, pathlib.PurePath]]=None`

src/codegen/templates/_task_modules/in_stream.py.mako
```diff
@@ -94,13 +95,13 @@ ${property_template.script_property(attribute)}\
             return num_samps_per_chan
 
     def configure_logging(
-            self, file_path, logging_mode=LoggingMode.LOG_AND_READ,
+            self, file_path: pathlib.PurePath, logging_mode=LoggingMode.LOG_AND_READ,
```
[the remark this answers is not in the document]
WayneDroid: This function will accept file_path either `str` or `pathlib.PurePath`, and then file_path is casted to `str` when passing to interpreter.

generated/nidaqmx/_task_modules/in_stream.py
WayneDroid: Noted, it's changed according to suggestion.

generated/nidaqmx/_task_modules/in_stream.py
WayneDroid: My mistake, mistook your suggestion `pathlib.Path` to `pathlib.PurePath`. The return type is corrected as per suggestion.

src/codegen/templates/function_template.py.mako
WayneDroid: `# type: ignore` is removed in new commit.

src/codegen/templates/task/_in_stream.py.mako
```diff
@@ -94,13 +96,13 @@ ${property_template.script_property(attribute)}\
...
     def configure_logging(
-            self, file_path, logging_mode=LoggingMode.LOG_AND_READ,
+            self, file_path: Union[str, pathlib.PurePath], logging_mode=LoggingMode.LOG_AND_READ,
             group_name="", operation=LoggingOperation.OPEN_OR_CREATE):
         """
         Configures TDMS file logging for the task.
 
         Args:
-            file_path (str): Specifies the path to the TDMS file to
+            file_path (Union[str, pathlib.PurePath]): Specifies the path to the TDMS file to
```
bkeryan, line 105: FYI, parameters that have type hints don't need to specify the type in the docstring: https://github.com/ni/nidaqmx-python/pull/539#discussion_r1525339107

src/codegen/templates/property_setter_template.py.mako
```diff
@@ -32,7 +36,10 @@
         if attribute.python_class_name == "Watchdog":
             function_call_args.append("\"\"")
         function_call_args.append(hex(attribute.id))
-        function_call_args.append('val')
+        if attribute.name in ATTRIBUTES_WITH_FILE_PATH_PARAMETER:
+            function_call_args.append('str(val)')
+        else:
```
bkeryan, line 41: Move this special case to lines 12-27 where the other special cases are.

src/codegen/utilities/attribute_helpers.py
```diff
@@ -233,6 +233,10 @@
     "PhysicalChannel": "physical_chan",
 }
 
+ATTRIBUTES_RETURN_FILE_PATH_TYPE = ("logging_file_path",)
+
+ATTRIBUTES_WITH_FILE_PATH_PARAMETER = ("logging_file_path",)
```
bkeryan, line 238: Is there a good reason for these to be separate? Since you are updating both the getter and setter for the one file path property, I think you should merge these.

src/codegen/templates/property_getter_template.py.mako
```diff
@@ -1,13 +1,21 @@
 <%def name="script_property_getter(attribute)">\
 <%
         from codegen.utilities.text_wrappers import docstring_wrap
-        from codegen.utilities.attribute_helpers import get_generic_attribute_function_name, get_generic_attribute_function_type
+        from codegen.utilities.attribute_helpers import get_generic_attribute_function_name, get_generic_attribute_function_type, ATTRIBUTES_RETURN_FILE_PATH_TYPE
     %>\
+    %if attribute.name in ATTRIBUTES_RETURN_FILE_PATH_TYPE:
```
bkeryan, line 6: FYI, I was going to suggest that you wait until https://github.com/ni/nidaqmx-python/issues/209 so we don't end up adding special cases, but you know what? It's fine. The code for returning the value has a ton of special cases anyway. In the worst case, the code declaring the property would have the same number of special cases, but it will probably have less. 😅

src/codegen/templates/task/_in_stream.py.mako
```diff
@@ -11,12 +11,14 @@ from typing import Type
...
 import numpy.typing
 import deprecation
+import pathlib
 
 from nidaqmx._task_modules.channels.channel import Channel
 from nidaqmx.utils import unflatten_channel_string
 from nidaqmx.constants import (
     ${', '.join([c for c in enums_used]) | wrap(4, 4)})
 
+from typing import Union
```
bkeryan, line 21: Add this to line 9. These imports are sorted into 3 sections: standard, 3rd party, nidaqmx.

tests/component/_task_modules/test_in_stream_read_properties.py
```diff
@@ -126,7 +128,7 @@ def test___ai_task___reset_bool_property___returns_default_value(ai_task: Task):
     raises=DaqError,
 )
 def test___ai_task___get_string_property___returns_default_value(ai_task: Task):
-    assert ai_task.in_stream.logging_file_path == ""
+    assert ai_task.in_stream.logging_file_path == pathlib.Path()
```
bkeryan, line 131: Sigh, I'm not sure we want this to return a path that is equivalent to "." (current directory).

bkeryan (changes requested):

src/codegen/templates/task/_in_stream.py.mako
WayneDroid: Noted. `Union` is moved up to line 9.

src/codegen/utilities/attribute_helpers.py
WayneDroid: I thought there could be some attributes have either return or accept file path type. It's fixed.

src/codegen/templates/task/_in_stream.py.mako
WayneDroid: It's updated as per suggestion. I also updated [function_template.py.mako ](https://github.com/ni/nidaqmx-python/pull/542/files#diff-de3716142179fdfddaef0c1a49af4f98923119b0ed3264e7a17d5da7fe276ed9R40) to get the same effect for functions in physical_channel.py.

tests/component/_task_modules/test_in_stream_read_properties.py
WayneDroid: @mshafer-NI do you have any thoughts on returning "." path as default value? It seems like [pathlib](https://docs.python.org/3/library/pathlib.html#methods-and-properties) does not have property to be set as "not a path".

mshafer-NI: I normally use `None` for "not a path" -> in this case, it probably makes sense for the default to be `pathlib.Path.cwd()`

bkeryan: > I normally use None for "not a path"

Ok, so the property setter should accept `Optional[Union[str, pathlib.PurePath]]` and the property getter should return `Optional[pathlib.Path]`?

> -> in this case, it probably makes sense for the default to be pathlib.Path.cwd()

Why? This property is a file path, not a directory path.

bkeryan: Would it be better to keep returning `str` from the property getter, but accept `Union[str, pathlib.PurePath]` in the property setter?

mshafer-NI: > Why? This property is a file path, not a directory path.

Oh, I missed that, yeah, probably return None

> Would it be better to keep returning str from the property getter, but accept Union[str, pathlib.PurePath] in the property setter?

Our guidance is to use pathlib (because it's nicer) unless you are in a _very_ heavily executed code path where swithing to using string operations can save a measurable amount of time.

WayneDroid: Thanks for the inputs, I will change it to return None if the value is empty string; else return file path.

src/codegen/templates/property_getter_template.py.mako
```diff
@@ -1,13 +1,21 @@
 <%def name="script_property_getter(attribute)">\
 <%
         from codegen.utilities.text_wrappers import docstring_wrap
-        from codegen.utilities.attribute_helpers import get_generic_attribute_function_name, get_generic_attribute_function_type
+        from codegen.utilities.attribute_helpers import get_generic_attribute_function_name, get_generic_attribute_function_type, ATTRIBUTE_WITH_FILE_PATH_TYPE
     %>\
+    %if attribute.name in ATTRIBUTE_WITH_FILE_PATH_TYPE:
+    @property
+    def ${attribute.name}(self) -> Union[pathlib.Path, None]:
```
bkeryan, line 8:
```suggestion
    def ${attribute.name}(self) -> Optional[pathlib.Path]:
```

src/codegen/templates/property_setter_template.py.mako
```diff
@@ -1,9 +1,13 @@
 <%def name="script_property_setter(attribute)">\
 <%
-        from codegen.utilities.attribute_helpers import get_generic_attribute_function_name, get_generic_attribute_function_type
+        from codegen.utilities.attribute_helpers import get_generic_attribute_function_name, get_generic_attribute_function_type, ATTRIBUTE_WITH_FILE_PATH_TYPE
     %>\
     @${attribute.name}.setter
+    %if attribute.name in ATTRIBUTE_WITH_FILE_PATH_TYPE:
+    def ${attribute.name}(self, val: Union[str, pathlib.PurePath]):
```
bkeryan, line 7: If the property can return None, then it should accept None.
```suggestion
    def ${attribute.name}(self, val: Optional[Union[str, pathlib.PurePath]]):
```

src/codegen/templates/property_setter_template.py.mako
```diff
@@ -20,6 +24,8 @@
         val = flatten_channel_string(val)
     %elif attribute.is_list:
         val = numpy.array(val, dtype=${attribute.ctypes_data_type})
+    %elif attribute.name in ATTRIBUTE_WITH_FILE_PATH_TYPE:
+        val = str(val)
```
bkeryan, line 28:
```suggestion
        if val is None:
            val = ""
        val = str(val)
```

bkeryan (approved):
Approved with suggestions

src/codegen/templates/property_getter_template.py.mako
WayneDroid: Thanks for the suggestion, it's updated.

src/codegen/templates/property_setter_template.py.mako
WayneDroid: Noted, it's updated.

src/codegen/templates/property_setter_template.py.mako
WayneDroid: Noted, it's updated.

zhindes (approved):

bkeryan (approved):

PR #550, opened 2024-03-22
Improve error reporting in some cases

texasaggie97:

If for some reason the extended error message is blank, the `DaqError` ends up not showing any output when it is printed. This can happen if the DLL doesn't properly load plugins. If message is blank, try to put at least something in to the message that is then given to the parent Error class.

Changed `DaqReadError` and `DaqWriteError` to let `DaqError` handle setting the message

Improves error reporting

Local testing

### DIFF
--- a/generated/nidaqmx/errors.py
+++ b/generated/nidaqmx/errors.py
@@ -54,12 +54,12 @@ class DaqError(Error):
 
         # If message is empty, we try to put at least some information in it
         if not message and self._error_type != DAQmxErrors.UNKNOWN:
-            message = f'{self._error_type.name} occurred. Error code: {self._error_code}.'
-        elif not message:
-            message = f'Error code: {self._error_code}.'
+            message = f'Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n\nError {self._error_type.name}'
 
         if task_name:
             message = f'{message}\n\nTask Name: {task_name}'
+            
+        message = f'{message}\n\nStatus Code: {self._error_code}'
 
         super().__init__(message)
 
--- a/generated/nidaqmx/errors.py
+++ b/generated/nidaqmx/errors.py
@@ -65,7 +65,6 @@ class DaqError(Error):
 
         super().__init__(message)
 
-
     @property
     def error_code(self):
         """
--- a/generated/nidaqmx/errors.py
+++ b/generated/nidaqmx/errors.py
@@ -54,7 +54,7 @@ class DaqError(Error):
 
         # If message is empty, we try to put at least some information in it
         if not message:
-            message = f'Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n\nStatus Code: {self._error_code}'
+            message = f'Description could not be found for the status code.\n\nStatus Code: {self._error_code}'
 
         if task_name:
             message = f'{message}\n\nTask Name: {task_name}'
--- a/generated/nidaqmx/errors.py
+++ b/generated/nidaqmx/errors.py
@@ -53,15 +53,18 @@ class DaqError(Error):
             self._error_type = DAQmxErrors.UNKNOWN
 
         # If message is empty, we try to put at least some information in it
-        if not message and self._error_type != DAQmxErrors.UNKNOWN:
-            message = f'Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n\nError {self._error_type.name}'
+        if not message:
+            message = f'Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n\nStatus Code: {self._error_code}'
 
         if task_name:
             message = f'{message}\n\nTask Name: {task_name}'
             
-        message = f'{message}\n\nStatus Code: {self._error_code}'
+        # We do not know where the error description came from, so we add the status code if it is not already in the message
+        if str(self._error_code) not in message:
+            message = f'{message}\n\nStatus Code: {self._error_code}'
 
         super().__init__(message)
+
 
     @property
     def error_code(self):
--- a/generated/nidaqmx/errors.py
+++ b/generated/nidaqmx/errors.py
@@ -45,17 +45,23 @@ class DaqError(Error):
             message (string): Specifies the error message.
             error_code (int): Specifies the NI-DAQmx error code.
         """
-        if task_name:
-            message = f'{message}\n\nTask Name: {task_name}'
-
-        super().__init__(message)
-
         self._error_code = int(error_code)
 
         try:
             self._error_type = DAQmxErrors(self._error_code)
         except ValueError:
             self._error_type = DAQmxErrors.UNKNOWN
+
+        # If message is empty, we try to put at least some information in it
+        if not message and self._error_type != DAQmxErrors.UNKNOWN:
+            message = f'{self._error_type.name} occurred. Error code: {self._error_code}.'
+        elif not message:
+            message = f'Error code: {self._error_code}.'
+
+        if task_name:
+            message = f'{message}\n\nTask Name: {task_name}'
+
+        super().__init__(message)
 
     @property
     def error_code(self):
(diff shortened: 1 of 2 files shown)


src/handwritten/errors.py
```diff
@@ -45,18 +45,24 @@ def __init__(self, message, error_code, task_name=''):
...
         try:
             self._error_type = DAQmxErrors(self._error_code)
         except ValueError:
             self._error_type = DAQmxErrors.UNKNOWN
 
+        # If message is empty, we try to put at least some information in it
+        if not message and self._error_type != DAQmxErrors.UNKNOWN:
+            message = f'{self._error_type.name} occurred. Error code: {self._error_code}.'
+        elif not message:
+            message = f'Error code: {self._error_code}.'
```
bkeryan, line 59: Let's match the C API's terminology and formatting: "Status Code: -200220" (no period)

```
PS D:\dev\nidaqmx-python> poetry run python .\examples\ai_multi_task_pxie_ref_clk.py
  File "D:\dev\nidaqmx-python\examples\ai_multi_task_pxie_ref_clk.py", line 12, in <module>
    master_task.ai_channels.add_ai_voltage_chan("/PXI1Slot3/ai0")
  File "D:\dev\nidaqmx-python\generated\nidaqmx\task\collections\_ai_channel_collection.py", line 2224, in add_ai_voltage_chan
    self._interpreter.create_ai_voltage_chan(
  File "D:\dev\nidaqmx-python\generated\nidaqmx\_library_interpreter.py", line 1279, in create_ai_voltage_chan
  File "D:\dev\nidaqmx-python\generated\nidaqmx\_library_interpreter.py", line 6381, in check_for_error
    raise DaqError(extended_error_info, error_code)
nidaqmx.errors.DaqError: Device identifier is invalid.

Device Specified: PXI1Slot3
Suggested Device(s): Dev1

Task Name: _unnamedTask<0>

Status Code: -200220
```

src/handwritten/errors.py
```diff
@@ -45,18 +45,24 @@ def __init__(self, message, error_code, task_name=''):
...
+
+        if task_name:
+            message = f'{message}\n\nTask Name: {task_name}'
```
bkeryan, line 62: FYI: In case you're wondering why this doesn't log the task name twice, it's because the interpreters don't specify a task name when throwing a DaqError. Only errors that originate in nidaqmx-python use this parameter.

src/handwritten/errors.py
bkeryan, line 59: "Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n"?

bkeryan (changes requested):

generated/nidaqmx/errors.py
```diff
@@ -117,19 +114,10 @@ def __init__(self, message, error_code, samps_per_chan_written, task_name=''):
             error_code (int): Specifies the NI-DAQmx error code.
             samps_per_chan_written (int): Specifies the number of samples written.
         """
-        if task_name:
-            message = f'{message}\n\nTask Name: {task_name}'
-
         super().__init__(message, error_code, task_name)
```
bkeryan, line 117: FYI, this probably logged the task name twice.

src/handwritten/errors.py
zhindes, line 62: I think you should invert this to better match .NET

image

pseudocode:

```python
message = f"{message}\n\nTask Name: {task_name}\n\nStatus: {self._error_type.name} ({self._error_code})"
```

zhindes (changes requested):
woops, did a single comment before. see my comment about inverting the message order

src/handwritten/errors.py
zhindes: I like putting the python name, so what about:

`Status Code: -200220 (INVALID_DEVICE_ID)`

texasaggie97: Changed

texasaggie97: Changed

src/handwritten/errors.py
texasaggie97: Changed

generated/nidaqmx/errors.py
```diff
@@ -45,18 +45,24 @@ def __init__(self, message, error_code, task_name=''):
...
             self._error_type = DAQmxErrors.UNKNOWN
 
+        # If message is empty, we try to put at least some information in it
+        if not message and self._error_type != DAQmxErrors.UNKNOWN:
+            message = f'Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n\nError {self._error_type.name}'
+
+        if task_name:
+            message = f'{message}\n\nTask Name: {task_name}'
+            
+        message = f'{message}\n\nStatus Code: {self._error_code}'
```
bkeryan, line 62: The C API already adds "Status Code:". This code should only add it when the message didn't come from the C API.

generated/nidaqmx/errors.py
```diff
@@ -45,18 +45,24 @@ def __init__(self, message, error_code, task_name=''):
...
         self._error_code = int(error_code)
 
         try:
             self._error_type = DAQmxErrors(self._error_code)
         except ValueError:
             self._error_type = DAQmxErrors.UNKNOWN
 
+        # If message is empty, we try to put at least some information in it
+        if not message and self._error_type != DAQmxErrors.UNKNOWN:
+            message = f'Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n\nError {self._error_type.name}'
```
bkeryan, line 57: If the error is not in the enum, this will omit the text and log "\n\nTask Name: ..."

I think you should simplify this and ignore the enum. Users should very rarely encounter this situation, and if they do, the status code should be sufficient to debug the problem.

```suggestion
        # If message is empty, we try to put at least some information in it
        if not message:
            message = f'Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n\nStatus Code: {self._error_code}'
```

bkeryan (changes requested):

generated/nidaqmx/errors.py
texasaggie97: Changed.

src/handwritten/errors.py
```diff
@@ -45,18 +45,26 @@ def __init__(self, message, error_code, task_name=''):
...
         self._error_code = int(error_code)
 
         try:
             self._error_type = DAQmxErrors(self._error_code)
         except ValueError:
             self._error_type = DAQmxErrors.UNKNOWN
 
+        # If message is empty, we try to put at least some information in it
+        if not message:
+            message = f'Explanation could not be found for the requested status code.\n\nVerify that the requested status code is correct.\n\nStatus Code: {self._error_code}'
```
zhindes, line 57: The user getting the error can't verify anything, I don't think...

```suggestion
            message = f'Explanation could not be found for the status code.\n\nStatus Code: {self._error_code}'
```

zhindes (approved):
minor nit on the string when the description can't be found.

src/handwritten/errors.py
texasaggie97: That string came from @bkeryan. Is this change acceptable?

bkeryan: @zhindes It's not a new error message but your change is ok with me.

```
      /* hex: 0xFFFCF1D5 */
      /* Explanation could not be found for the requested status code.<LF/><LF/>Verify that the requested status code is correct. */
      kErrorExplanationNotFound = -200235,
```

bkeryan (approved):

src/handwritten/errors.py
bkeryan: I think "Verify that the requested status code is correct." probably was intended to mean "Verify that you didn't pass DAQmxGetErrorString a bogus status code like 0xCCCCCCCC" but in this context, that text is not as appropriate.

maxxboehme (approved):

PR #552, opened 2024-04-01
Accept either comma delimited list or flattened string for multi edge trigger source

texasaggie97:

* Accept either comma delimited list or flattened string for multi edge trigger source
* Add varying number of samples to ensure channels are not mixed up
* Set non-zero status to verify done event is called

* Ensure proper passing of tests

* Add tests pass both with and without redirection


tests/acceptance/test_multi_threading.py
```diff
@@ -187,7 +187,7 @@ def test___shared_interpreter___run_multiple_acquisitions_with_events___callback
             )
             for i, device in enumerate(multi_threading_test_devices)
         ]
-        samples_per_chan = [1000 for _ in tasks]
```
zhindes, line 190: either split this into 2 PRs or update the title to be slightly more generic about what you're doing. I prefer splitting them, personally.

tests/acceptance/test_multi_threading.py
```diff
@@ -187,7 +187,7 @@ def test___shared_interpreter___run_multiple_acquisitions_with_events___callback
...
+        samples_per_chan = [(100 * (i + 8)) for i in range(len(tasks))]
```
zhindes, line 190:
```suggestion
        # Set a unique value per-task to ensure events are mixed up
        samples_per_chan = [(100 * (i + 8)) for i in range(len(tasks))]
```

tests/acceptance/test_multi_threading.py
```diff
@@ -218,7 +218,7 @@ def test___shared_interpreter___unregister_events_during_other_acquisitions_with
...
```
zhindes, line 221:
```suggestion
        # Set a unique value per-task to ensure events are mixed up
        samples_per_chan = [(100 * (i + 8)) for i in range(len(tasks))]
```

zhindes (changes requested):

tests/acceptance/test_multi_threading.py
bkeryan, line 190: `test___shared_interpreter___run_multiple_acquisitions_with_events___callbacks_invoked` doesn't describe what you're testing. Since you are adding new test coverage, please create a new test function that describes what you're testing. As I suggested in chat, "Given shared interpreter & varying sample interval, when I run multiple acquisitions with every N samples events, then requested sample interval passed to callback"

`test___shared_interpreter_varying_sample_interval___run_multiple_acquisitions_with_every_n_samples_events___requested_sample_interval_passed_to_callback`

tests/acceptance/test_multi_threading.py
bkeryan, line 221: I don't think it's necessary to add the new test coverage to the unregister-during-other-acquisitions test.

bkeryan (changes requested):

tests/_event_utils.py
bkeryan, line 1: The first two files still have a whitespace change. I think you deleted the newline at the end of the file.

bkeryan (approved):
Approved with suggestions

PR #558, opened 2024-04-03
Add tests to verify closing task in event callback does not crash or error

texasaggie97:

DAQmx allows stopping and closing the task in an event callback. Test that is does not crash and behaves properly.

Improve test coverage

* Local tests pass
* PR

### DIFF
--- a/tests/component/test_task_events.py
+++ b/tests/component/test_task_events.py
@@ -1,3 +1,4 @@
+import threading
 import time
 import traceback
 from logging import LogRecord
@@ -60,10 +61,15 @@ def test___done_event_registered___call_clear_in_callback___stop_close_in_callba
     init_kwargs: dict,
 ) -> None:
 
+    # We have to define and set this here because mypy will error if it is
+    # defined below with everything else: https://github.com/python/mypy/issues/7057
+    side_effect_semaphore = threading.Semaphore(value=0)
+
     # side_effect for callback
     def _clear_task():
         task.stop()
         task.close()
+        side_effect_semaphore.release()
 
     try:
         # We need to create our own task here because we need to call stop() and close() on it.
@@ -83,9 +89,7 @@ def test___done_event_registered___call_clear_in_callback___stop_close_in_callba
 
     task.start()
     event_observer.wait_for_events(1)
-    # The side_effect is invoked after the semaphore is released, so we need to wait a bit for it
-    # to actually be called and stop and close the task
-    time.sleep(0.1)
+    acquired = side_effect_semaphore.acquire(timeout=1.0)
     # Ensure we get the expected exception and warning
     with pytest.raises(nidaqmx.DaqError) as exc_info:
         task.stop()
@@ -95,6 +99,7 @@ def test___done_event_registered___call_clear_in_callback___stop_close_in_callba
     ) as warnings_record:
         task.close()
 
+    assert acquired
     assert len(event_observer.events) == 1
     assert event_observer.events[0].status == 0
     assert exc_info.value.error_code == DAQmxErrors.INVALID_TASK
@@ -106,13 +111,19 @@ def test___every_n_samples_event_registered___call_clear_in_callback___stop_clos
     init_kwargs: dict,
 ) -> None:
 
+    # We have to define and set this here because mypy will error if it is
+    # defined below with everything else: https://github.com/python/mypy/issues/7057
+    callback_call_number = 0
+    side_effect_semaphore = threading.Semaphore(value=0)
+
     # side_effect for callback
     def _clear_task():
         nonlocal callback_call_number
-        if callback_call_number == 3:
+        callback_call_number += 1
+        if callback_call_number == 4:
             task.stop()
             task.close()
-        callback_call_number += 1
+            side_effect_semaphore.release()
 
     try:
         # We need to create our own task here because we need to call stop() and close() on it.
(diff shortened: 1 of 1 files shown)


tests/component/test_task_events.py
```diff
@@ -55,6 +55,104 @@ def test___done_event_registered___run_finite_acquisition___callback_invoked_onc
...
+    task.register_done_event(event_observer.handle_done_event)
+    task.timing.cfg_samp_clk_timing(
+        rate=10000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=1000
+    )
+
+    task.start()
+    event_observer.wait_for_events(1)
+    # The side_effect is invoked after the semaphore is released, so we need to wait a bit for it
+    # to actually be called and stop and close the task
+    time.sleep(0.1)
```
zhindes, line 88: can you create a local semaphore, release it in `_clear_task` and wait on it here with a timeout (and up to 1s)? if the timeout elapses, something is probably wrong we can just assert that `acquire` returned true.

tests/component/test_task_events.py
```diff
@@ -55,6 +55,104 @@ def test___done_event_registered___run_finite_acquisition___callback_invoked_onc
...
+        task.stop()
+    with pytest.warns(
+        DaqResourceWarning,
+        match="Attempted to close NI-DAQmx task",
+    ) as warnings_record:
+        task.close()
+
+    assert len(event_observer.events) == 1
+    assert event_observer.events[0].status == 0
+    assert exc_info.value.error_code == DAQmxErrors.INVALID_TASK
```
zhindes, line 100: we chatted about this, redirection has different behavior, which isn't good. for now accept either error, but lets reference the bug you file here, because we'll want to fix that before we ship

zhindes (dismissed):
ditto my comments on the other test

tests/component/test_task_events.py
texasaggie97: We decided to allow the tests to fail in the internal test suite until internal bug #2700979 is fixed.

zhindes (approved):

tests/component/test_task_events.py
```diff
@@ -55,6 +56,118 @@ def test___done_event_registered___run_finite_acquisition___callback_invoked_onc
     assert all(e.status == 0 for e in event_observer.events)
 
 
+@pytest.mark.grpc_xfail(
+    reason="#559: Can't close tasks in an event callback when using gRPC", raises=AssertionError
+)
+def test___done_event_registered___call_clear_in_callback___stop_close_in_callback_with_success_status(
```
bkeryan, line 62: "call_clear_in_callback" describes the underlying C API call. The "act" section of this test does "call_stop_close_in_callback"

"stop_close_in_callback_with_success_status" sort of describes the result but I think "task_closed_with_success_status" matches the code better.

tests/component/test_task_events.py
```diff
@@ -55,6 +56,118 @@ def test___done_event_registered___run_finite_acquisition___callback_invoked_onc
...
+    event_observer.wait_for_events(1)
+    acquired = side_effect_semaphore.acquire(timeout=1.0)
+    # Ensure we get the expected exception and warning
+    with pytest.raises(nidaqmx.DaqError) as exc_info:
+        task.stop()
+    with pytest.warns(
+        DaqResourceWarning,
+        match="Attempted to close NI-DAQmx task",
+    ) as warnings_record:
+        task.close()
```
bkeryan, line 103: This looks like it's asserting that the task was already closed, which belongs in "assert" not "act"

tests/component/test_task_events.py
```diff
@@ -55,6 +56,118 @@ def test___done_event_registered___run_finite_acquisition___callback_invoked_onc
...
+        if task is not None:
+            task.close()
+        raise
+    event_observer = DoneEventObserver(side_effect=_clear_task)
+    task.register_done_event(event_observer.handle_done_event)
+    task.timing.cfg_samp_clk_timing(
+        rate=10000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=1000
+    )
+
+    task.start()
```
bkeryan, line 93: This try/except block covers some operations but not others. It's ok to put the entire test in the try block if we can't use a test fixture for cleanup.

tests/component/test_task_events.py
```diff
@@ -55,6 +56,118 @@ def test___done_event_registered___run_finite_acquisition___callback_invoked_onc
...
+    event_observer.wait_for_events(1)
+    acquired = side_effect_semaphore.acquire(timeout=1.0)
```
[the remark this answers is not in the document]
texasaggie97: Handled in #564

tests/component/test_task_events.py
texasaggie97: Handled in #564

tests/component/test_task_events.py
texasaggie97: Handled in #564

tests/component/test_task_events.py
texasaggie97: This is on purpose. We are not capturing the exception to prevent it from being raised. We only capture the DAQmx exception so we can close the task if needed, but we still raise the exception. We don't need to catch anything else, but all exceptions are errors. pytest will capture it and report it as an error.

bkeryan: Why is it important to avoid leaking the underlying C API task when `add_ai_voltage_chan` throws a DaqError but not when `register_done_event`, `start`, or `event_observer.wait_for_events` throws an exception? I think both are important, because leaking a task can affect other tests (especially if it has resources reserved, like in the case of `event_observer.wait_for_events`).

I am suggesting that you expand this try/except to cover the entire test. Also, change DaqError to Exception so it covers timeout errors etc.

```
try:
    # arrange

   # act

   # assert
except Exception:
   # cleanup
   raise
```

bkeryan: Normally the `task` or `generate_task` test fixture does this cleanup for us. I understand why you didn't use those fixtures here: double-closing the task would emit a DaqResourceWarning.

PR #560, opened 2024-04-05
Update hightime library

WayneDroid:

- updated hightime version to `2.2`
- replaced `total_seconds` with `precision_total_seconds` for hightime datetime
- updated test cases involved `femtosecond` and `yoctosecond` use case

Update hightime library to have better precision on total seconds.

`poetry run pytest` - pass

image


tests/unit/test_grpc_time.py
```diff
@@ -302,7 +305,10 @@ def test___datetime_before_1970_with_femtoseconds___convert_to_timestamp___is_re
     to_ts = grpc_time.convert_time_to_timestamp(from_dt)
     roundtrip_dt = grpc_time.convert_timestamp_to_time(to_ts, tzinfo=timezone.utc)
 
-    assert to_ts.seconds == JAN_01_1850_TIMESTAMP_1970_EPOCH
+    if femtosecond:
+        assert to_ts.seconds == JAN_01_1850_TIMESTAMP_1970_EPOCH + 1
```
zhindes, line 309: this is because the seconds are negative, right? since we have non-zero subseconds, the whole seconds is off by one? that makes sense, but can you add a comment to all these test cases explaining that a bit. it took me a second (hoho!) to figure it out

zhindes (approved):
add some comments is all! FYI brad is out, so feel free to complete when you add those

tests/unit/test_grpc_time.py
WayneDroid: Comments are added. :)

PR #570, opened 2024-04-23
Add and update AO examples

WayneDroid:

- renamed exisiting AO examples
- added finite and continuous examples

Add AO examples for release 1.0.

run `poetry run pytest tests/ -q` - pass

image


examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,54 @@
...
+        print("Starting task. Press Ctrl+C to stop.")
+        time.sleep(1.0)
+
+        while True:
+            data = get_sine_wave_data(cycle)
+            task.write(data)
+            cycle = cycle + 1
+            if is_first_run:
+                is_first_run = False
+                task.start()
```
WayneDroid, line 50: @zhindes @bkeryan 
I'm not sure am I doing it right for this example. Also, should I use `stream_writers`?

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,32 @@
...
+    duration = 1
+
+    # generate the time array
+    t = np.arange(0, duration, 1 / sampling_rate)
+    # Generate the sine wave
+    data = amplitude * np.sin(2 * np.pi * frequency * t)
+
+    task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+    task.timing.cfg_samp_clk_timing(
+        1000, sample_mode=AcquisitionType.CONTINUOUS, samps_per_chan=100
```
zhindes, line 25: should 1000 be sampling_rate?

```suggestion
        sampling_rate, sample_mode=AcquisitionType.CONTINUOUS, samps_per_chan=100
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_every_n_samples_event.py
```diff
@@ -0,0 +1,40 @@
...
+    data = amplitude * np.sin(2 * np.pi * frequency * t)
+
+    def callback(task_handle, every_n_samples_event_type, number_of_samples, callback_data):
+        """Callback function for written data."""
+        print(f"transferred {number_of_samples} samples event invoked.")
+        return 0
+
+    task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+    task.timing.cfg_samp_clk_timing(
+        1000, sample_mode=AcquisitionType.CONTINUOUS, samps_per_chan=100
```
zhindes, line 32:
```suggestion
        sampling_rate, sample_mode=AcquisitionType.CONTINUOUS, samps_per_chan=100
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,54 @@
...
+    data = amplitude[selection] * np.sin(2 * np.pi * frequency[selection] * t)
+    return data
+
+
+with nidaqmx.Task() as task:
+    is_first_run = True
+    task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+    task.out_stream.regen_mode = RegenerationMode.DONT_ALLOW_REGENERATION
+    task.timing.cfg_samp_clk_timing(
+        1000, sample_mode=AcquisitionType.CONTINUOUS, samps_per_chan=100
```
zhindes, line 36:
```suggestion
        sampling_rate, sample_mode=AcquisitionType.CONTINUOUS, samps_per_chan=100
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,54 @@
...
+
+import nidaqmx
+from nidaqmx.constants import AcquisitionType, RegenerationMode
+
+
+def get_sine_wave_data(cycle):
+    """Generate a sine wave data."""
+    frequency = (10, 50)
+    amplitude = (1, 3)
+    sampling_rate = (100, 500)
```
zhindes, line 20: you can change sampling_rate in the middle of a task, but you're not doing that

```suggestion
    sampling_rate = (100, 100)
```

examples/analog_out/gen_voltage_sample.py
```diff
@@ -0,0 +1,13 @@
...
+(Sample) to an Analog Output Channel.
+"""
+
+import nidaqmx
+
+with nidaqmx.Task() as task:
+    task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+
+    print(f"Generate {task.write(1.1)} voltage sample.")
+    task.stop()
```
zhindes, line 13: unnecessary since you didn't explicitly call start

examples/analog_out/gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,17 @@
...
+voltage samples to an Analog Output Channel using an internal
+sample clock.
+"""
+
+import nidaqmx
+from nidaqmx.constants import AcquisitionType
+
+with nidaqmx.Task() as task:
+    task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+    task.timing.cfg_samp_clk_timing(1000, sample_mode=AcquisitionType.FINITE, samps_per_chan=1000)
```
zhindes, line 13: since we're doing so few samples, I'd slow this down

```suggestion
    task.timing.cfg_samp_clk_timing(10.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=1000)
```

examples/analog_out/gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,17 @@
...
+
+    print(f"Generate {task.write([1.1, 2.2, 3.3, 4.4, 5.5])} voltage samples.")
```
zhindes, line 15: not a big fan of an f-string having such a massive side-effect. can you make this two statements? Also, I would be explicit about auto-start optional variable.

```suggestion
    number_of_samples_written = task.write([1.1, 2.2, 3.3, 4.4, 5.5], auto_start=True)
    print(f"Generate {number_of_samples_written} voltage samples.")
```

zhindes (changes requested):

examples/analog_out/gen_voltage_wfm_int_clk.py
bkeryan: I agree. Printing to the console is presentation logic and removing it should not affect the business logic. Please remove all calls from `task.write` from f-strings.

examples/analog_out/gen_voltage_sample.py
bkeryan, line 1: The ANSI C example is named `VoltUpdate.c`. I would go with `voltage_update.py` or `volt_update.py`.

examples/analog_out/gen_voltage_wfm_int_clk.py
bkeryan: Alternately, we can use math to generate more samples. From `MultVoltUpdates-IntClk.c`:
```
	for(;i<4000;i++)
		data[i] = 5.0*(double)i/4000.0;
```
From `ContGen-IntClk.c` (but I see, you added a sine wave to the continuous example already):
```
	for(;i<1000;i++)
		data[i] = 9.95*sin((double)i*2.0*PI/1000.0);
```

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+    task.write(data)
+    task.start()
+
+    input("Running task. Press Enter to stop.\n")
```
bkeryan, line 30: The ANSI C example's "Generating voltage continuously" is a better description of what it's doing.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,32 @@
...
+waveform using an internal sample clock.
+"""
+
+import numpy as np
+
+import nidaqmx
+from nidaqmx.constants import AcquisitionType
+
+with nidaqmx.Task() as task:
+    frequency = 10
```
bkeryan, line 13: FYI, the ANSI C examples use 9.95 because some older devices like E Series and some S Series can't output full-scale voltage without coercing and returning a warning. (This is because the max voltage is 1 LSB smaller than the min voltage: 32767 vs. -32768, etc.)

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
bkeryan, line 25: samps_per_chan is unnecessary here. Calling task.write() will set the buffer size, and total samples to generate is not relevant for continuous.

examples/analog_out/cont_gen_voltage_wfm_int_clk_every_n_samples_event.py
```diff
@@ -0,0 +1,40 @@
...
+    duration = 1
+
+    # generate the time array
+    t = np.arange(0, duration, 1 / sampling_rate)
+    # Generate the sine wave
+    data = amplitude * np.sin(2 * np.pi * frequency * t)
+
+    def callback(task_handle, every_n_samples_event_type, number_of_samples, callback_data):
+        """Callback function for written data."""
+        print(f"transferred {number_of_samples} samples event invoked.")
```
bkeryan, line 27: This is a run-on of two separate sentences:
- transferred N samples
- every N samples event invoked

Choose one or the other. (I like "Transferred N samples" better")

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
bkeryan: You don't need to change the signal parameters to demonstrate non-regen. You just need to output a rate that does not evenly divide the sample clock rate.

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,54 @@
+"""Example of analog output voltage generation.
+
+This example demonstrates how to output a continuous periodic
+waveform with new data using an internal sample clock while
+the task is running.
```
bkeryan, line 5: The LV example help explains this better and provides a default rate that demonstrates non-regen.

> This example demonstrates how to continuously generate an analog output waveform by providing new data to the output buffer as the task is running.
> 
> This example is useful if you want to generate a non-repeating waveform, make updates on-the-fly, or generate a frequency that is not an even divide-down of your sample clock. In this example, the default frequency value is 17 to demonstrate that non-regenerative output can be used to create a signal with a frequency that is not an even divide-down of your sample clock.

bkeryan (changes requested):

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,54 @@
...
+    amplitude = (1, 3)
+    sampling_rate = (100, 500)
+    duration = 1
+    selection = cycle % 2
+
+    # generate the time array
+    t = np.arange(0, duration, 1 / sampling_rate[selection])
+    # Generate the sine wave
+    data = amplitude[selection] * np.sin(2 * np.pi * frequency[selection] * t)
+    return data
```
bkeryan, line 28: This does not preserve phase from call to call, so it will generate a glitch in the sine wave for some frequencies.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,32 @@
...
+    amplitude = 1
+    sampling_rate = 100
+    duration = 1
+
+    # generate the time array
+    t = np.arange(0, duration, 1 / sampling_rate)
+    # Generate the sine wave
+    data = amplitude * np.sin(2 * np.pi * frequency * t)
```
bkeryan, line 21: [style] Instead of declaring a ton of global variables, how about writing a function that takes those variables as parameters?

for example,
```
def create_sine_wave(frequency: float, amplitude: float, sampling_rate: float, duration: float) -> numpy.typing.NDArray[numpy.double]:
   ...
```

examples/analog_out/gen_voltage_sample.py
WayneDroid: Noted, `stop` is removed.

examples/analog_out/gen_voltage_wfm_int_clk.py
WayneDroid: Okay, I take @bkeryan 's opinion, referring `MultVoltUpdates-IntClk.c` to generate 1000 samples.

examples/analog_out/gen_voltage_wfm_int_clk.py
WayneDroid: Sure, it's updated as per suggestion.

examples/analog_out/gen_voltage_sample.py
WayneDroid: The file has been renamed.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
WayneDroid: Okay, updated all continuous examples.

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
WayneDroid: Thanks for writing up the description. The examples is updated accordingly and tested on real device.

image

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
WayneDroid: Created helper to for the function.

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
WayneDroid: Yeah, it's removed. The example will demo how to generate a freq signal is not even divide-down by sample clock as per suggestion.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
WayneDroid: Noted. If that's the case, should I update the frequence value to 9.95 to support these older devices?

examples/analog_out/analog_out_helper.py
```diff
@@ -0,0 +1,24 @@
...
+    return amplitude * np.sin(2 * np.pi * frequency * t)
+
+
+def create_voltage_sample(voltage: float, number_of_samples: int) -> float:
+    """Generate a voltage sample."""
+    data = []
+    for i in range(number_of_samples):
+        data.append(voltage * i / number_of_samples)
+
+    return data
```
bkeryan, line 24: Use a list comprehension. Does this one even need to be a function?

```suggestion
    return [voltage * i / number_of_samples for i in range(number_of_samples)]
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_every_n_samples_event.py
bkeryan: This was a comment about the text. I wasn't suggesting to remove the f-string. Sorry if that wasn't clear.

examples/analog_out/cont_gen_voltage_wfm_int_clk_every_n_samples_event.py
```diff
@@ -6,35 +6,29 @@
...
-    data = amplitude * np.sin(2 * np.pi * frequency * t)
+    sampling_rate = 1000.0
+    data = create_sine_wave(
+        frequency=10.0, amplitude=1.0, sampling_rate=sampling_rate, duration=1.0
+    )
 
     def callback(task_handle, every_n_samples_event_type, number_of_samples, callback_data):
         """Callback function for written data."""
-        print(f"transferred {number_of_samples} samples event invoked.")
+        print("Transferred N samples")
```
bkeryan, line 22: This scrolls off the screen, right? How about printing the total and using `end="\r"` like you did for AI?

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -1,50 +1,37 @@
...
-        print("Starting task. Press Ctrl+C to stop.")
-        time.sleep(1.0)
-
+        print("Generating voltage continuously. Press Ctrl+C to stop.")
         while True:
-            data = get_sine_wave_data(cycle)
+            data = create_sine_wave(
+                frequency=17.0, amplitude=1.0, sampling_rate=sampling_rate, duration=1.0
+            )
             task.write(data)
```
bkeryan, line 34: This doesn't demonstrate a non-repeating waveform. Each loop iteration generates the same waveform chunk, and the number of samples in the chunk is based on the sample rate.

Look at the corresponding LabVIEW example. 
image

It doesn't generate the same sine wave data for each loop iteration. Instead, each iteration generates a new sine wave chunk that lines up with the previous sine wave chunk. The chunks do not necessarily contain an integral number of sine wave cycles. This is how you generate a continuous sine wave at a frequency that is not related to your sample rate and buffer size.

To demonstrate this, the `create_sine_wave` function should also take the number of samples to generate and the starting phase, and return the ending phase to use for the next call.

bkeryan (changes requested):

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
bkeryan: [numpy.linspace](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html#numpy.linspace) might be helpful. Or maybe it's easier to build a list with range() and use the `math` module.

bkeryan: Two examples to clarify:

1. The frequency you are generating is so slow, you can only generate one quadrant of the sine wave at a time.
- 1st iteration generates time range [0, 0.5*pi)
- 2nd iteration generates time range [0.5*pi, 1.0*pi)
- 3rd iteration generates time range [1.0*pi, 1.5*pi)
- 4th iteration generates time range [1.5*pi, 2*pi)
- 5th iteration generates time range [2*pi, 2.5*pi)

For each chunk, the phase input is the time offset at the start of the time range, and the phase output is the time offset at the end of the time range. Note that the time ranges are partially open intervals; we don't want to repeat the last sample of chunk N as the 1st sample of chunk N+1.

The sine function is cyclical and its implementation in computers gets less accurate as you get farther from the Y-axis, so it would make sense to keep the phase within 1 cycle of the Y-axis using the `%` operator or something. In this example, the 4th iteration would return an ending phase of 0 instead of 2*pi. 

Note that for many frequencies, the phase may never go back to exactly 0 at the end of a buffer. Or it may take a large number of iterations to do so. That's expected. In these situations, generating repeated data would introduce frequency error, and generating non-repeating data corrects that frequency error.

2. The frequency you are generating may translate to multiple sine wave cycles, then a fractional one. For example, one full cycle plus one quadrant.
- 1st iteration generates time range [0, 2.5*pi)
- 2nd iteration generates time range [2.5*pi, 5.0*pi)
- 3rd iteration generates time range [5.0*pi, 7.5*pi)
- 4th iteration generates time range [7.5*pi, 10.0*pi)

Keeping the phase within 2*pi by modding:
- 1st iteration phase in: 0, generate [0, 2.5*pi), phase out: 2.5*pi % 2*pi -> 0.5*pi
- 2nd iteration phase in: 0.5*pi, generate [0.5*pi, 3.0*pi), phase out: 3.0*pi % 2*pi -> 1.0*pi
- 3rd iteration phase in: 1.0*pi, generate [1.0*pi, 3.5*pi), phase out: 3.5*pi % 2*pi -> 1.5*pi
- 4th iteration phase in: 1.5*pi, generate [1.5*pi, 4.0*pi), phase out: 4.0*pi % 2*pi -> 0.0

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,41 @@
...
+
+from analog_out_helper import create_sine_wave
+
+import nidaqmx
+from nidaqmx.constants import AcquisitionType, RegenerationMode
+
+
+with nidaqmx.Task() as task:
+    is_first_run = True
+    sampling_rate = 1000.0
```
bkeryan, line 23: The DAQ device may not be able to output at exactly this sample rate. When that happens, it coerces the sample clock rate property to the closest rate that it can output. You should read back the sample rate in order to avoid introducing frequency error. 

You might think "any DAQ device should be able to generate exactly 1000.0, right?" However, devices with delta-sigma converters may not, depending on their oversample clocks and divisors. Also, the user is free to change the sampling rate.

examples/analog_out/analog_out_helper.py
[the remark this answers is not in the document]
WayneDroid: Noted, the helper file is removed.

examples/analog_out/analog_out_helper.py
WayneDroid: The function is not needed, your suggested list comprehension is used in the example file.

examples/analog_out/cont_gen_voltage_wfm_int_clk_every_n_samples_event.py
WayneDroid: True, the printing is updated as AI example.

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
WayneDroid: '''
The frequency you are generating may translate to multiple sine wave cycles, then a fractional one. For example, one full cycle plus one quadrant.
1st iteration generates time range [0, 2.5*pi)
2nd iteration generates time range [2.5*pi, 5.0*pi)
3rd iteration generates time range [5.0*pi, 7.5*pi)
4th iteration generates time range [7.5*pi, 10.0*pi)
Keeping the phase within 2*pi by modding:

1st iteration phase in: 0, generate [0, 2.5*pi), phase out: 2.5*pi % 2*pi -> 0.5*pi
2nd iteration phase in: 0.5*pi, generate [0.5*pi, 3.0*pi), phase out: 3.0*pi % 2*pi -> 1.0*pi
3rd iteration phase in: 1.0*pi, generate [1.0*pi, 3.5*pi), phase out: 3.5*pi % 2*pi -> 1.5*pi
4th iteration phase in: 1.5*pi, generate [1.5*pi, 4.0*pi), phase out: 4.0*pi % 2*pi -> 0.0
'''

I made changes to the comment above, hopefully, I get the points correctly. 
I tested with physical device, so far, I can see the waveform on the scope.

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
WayneDroid: Noted, added `actual_sampling_rate = task.timing.samp_clk_rate` to all impacted example files.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,45 @@
...
+
+def main():
+    """Continuously generates a sine wave."""
+    with nidaqmx.Task() as task:
+        sampling_rate = 1000.0
+        task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+        task.timing.cfg_samp_clk_timing(sampling_rate, sample_mode=AcquisitionType.CONTINUOUS)
+
+        actual_sampling_rate = task.timing.samp_clk_rate
+        print(f"Actual sampling rate: {actual_sampling_rate} Hz")
```
bkeryan, line 31: Units should be S/s (samples/sec).

Use format 'f' or 'g' to eliminate excess precision. In this case, 'g' is fine because we don't need to display trailing zeros.

```suggestion
        print(f"Actual sampling rate: {actual_sampling_rate:g} S/s")
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_every_n_samples_event.py
```diff
@@ -0,0 +1,59 @@
...
+            print(f"Transferred data: {number_of_samples} samples. Total {total_write}.", end="\r")
+
+            return 0
+
+        task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+        task.timing.cfg_samp_clk_timing(sampling_rate, sample_mode=AcquisitionType.CONTINUOUS)
+        task.register_every_n_samples_transferred_from_buffer_event(1000, callback)
+
+        actual_sampling_rate = task.timing.samp_clk_rate
+        print(f"Actual sampling rate: {actual_sampling_rate} Hz")
```
bkeryan, line 44:
```suggestion
        print(f"Actual sampling rate: {actual_sampling_rate:g} S/s")
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,72 @@
...
+    """Continuously generates non-repeating waveform."""
+    with nidaqmx.Task() as task:
+        is_first_run = True
+        sampling_rate = 1000.0
+        task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+        task.out_stream.regen_mode = RegenerationMode.DONT_ALLOW_REGENERATION
+        task.timing.cfg_samp_clk_timing(sampling_rate, sample_mode=AcquisitionType.CONTINUOUS)
+
+        actual_sampling_rate = task.timing.samp_clk_rate
+        print(f"Actual sampling rate: {actual_sampling_rate} Hz")
```
bkeryan, line 42:
```suggestion
        print(f"Actual sampling rate: {actual_sampling_rate:g} S/s")
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,72 @@
...
+
+import nidaqmx
+from nidaqmx.constants import AcquisitionType, RegenerationMode
+
+
+def create_sine_wave(
+    frequency: float, amplitude: float, sampling_rate: float, phase_in: float, phase_out: float
+) -> numpy.typing.NDArray[numpy.double]:
+    """Generate a sine wave."""
+    number_of_samples = int(sampling_rate * (phase_out - phase_in) * np.pi)
```
bkeryan, line 26: This highlights the fact that the way these example handle number_of_samples is backwards, compared to the AO examples for LV, C, CVI, etc.

In the AO examples for the other ADEs, number_of_samples is a parameter provided by the user. Here, you are not letting the user specify a number of samples. Instead, you are calculating the number of samples based on one period of the signal you are generating.

This approach is functional for low signal frequencies, but it has significant performance problems for high signal frequencies.

For example, if you are generating a 10000 Hz signal using a 1 MS/s sample clock, then each period is 10 samples and you must call DAQmxWrite 10000 times per second. This is unsustainable.

It is better to accept number_of_samples as a parameter and use np.linspace instead of np.arange. This applies to all of the examples that generate sine waves, but it has the biggest impact on the non-regen example.

Perhaps something like this? (untested)
```
duration_time = number_of_samples / sampling_rate
duration_radians = duration_time * 2 * np.pi
phase_out = (phase_in + duration_radians) % (2 * np.pi)
x = np.linspace(phase_in, phase_in + duration_radians, number_of_samples)
return amplitude * np.sin(frequency * x)
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,72 @@
...
+
+import numpy as np
+import numpy.typing
+
+import nidaqmx
+from nidaqmx.constants import AcquisitionType, RegenerationMode
+
+
+def create_sine_wave(
+    frequency: float, amplitude: float, sampling_rate: float, phase_in: float, phase_out: float
```
bkeryan, line 23: phase_out should be a return value (using a tuple), not an input parameter. Callers should not have to calculate the phase.

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,72 @@
...
+        task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+        task.out_stream.regen_mode = RegenerationMode.DONT_ALLOW_REGENERATION
+        task.timing.cfg_samp_clk_timing(sampling_rate, sample_mode=AcquisitionType.CONTINUOUS)
+
+        actual_sampling_rate = task.timing.samp_clk_rate
+        print(f"Actual sampling rate: {actual_sampling_rate} Hz")
+
+        try:
+            full_cycle = 2.0
+            full_cycle_and_one_quadrant = full_cycle + 0.5
```
bkeryan, line 46: "full cycle and one quadrant" was an example to illustrate one possible situation. You should not hard code the example to implement this particular situation.

create_sine_wave should calculate and return the correct phase_out based on the other parameters. The user should not have to calculate phase_out.

examples/analog_out/voltage_update.py
```diff
@@ -0,0 +1,12 @@
...
+This example demonstrates how to output a single Voltage Update
+(Sample) to an Analog Output Channel.
+"""
+
+import nidaqmx
+
+with nidaqmx.Task() as task:
+    task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+
+    print(f"Generate {task.write(1.1)} voltage sample.")
```
bkeryan, line 12: Don't call methods with side effects inside f-strings.

https://github.com/ni/nidaqmx-python/pull/570#discussion_r1576228003

examples/analog_out/gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,23 @@
...
+    data = []
+    total_samples = 1000
+    task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
+    task.timing.cfg_samp_clk_timing(
+        1000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=total_samples
+    )
+
+    data = [5.0 * i / total_samples for i in range(total_samples)]
+    number_of_samples_written = task.write(data, auto_start=True)
+    print(f"Generate {number_of_samples_written} voltage samples.")
```
bkeryan, line 21:
```suggestion
    print(f"Generating {number_of_samples_written} voltage samples.")
```

examples/analog_out/voltage_update.py
bkeryan, line 12: Change "generate" to either "generating" or "generated" depending on whether it's before or after the call to task.write.

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,72 @@
...
+) -> numpy.typing.NDArray[numpy.double]:
+    """Generate a sine wave."""
+    number_of_samples = int(sampling_rate * (phase_out - phase_in) * np.pi)
+    t = np.linspace(phase_in * np.pi, phase_out * np.pi, number_of_samples)
+
+    return amplitude * np.sin(2 * np.pi * frequency * t)
+
+
+def main():
+    """Continuously generates non-repeating waveform."""
```
bkeryan, line 33: Docstring summaries are imperative.

Did I link PEP 257 yet? https://peps.python.org/pep-0257/

bkeryan (changes requested):

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
bkeryan: Oops, duration_radians is wrong.

Dimensional analysis:
- duration_time is in seconds (S / (S/s) = s)
- frequency is in Hz = s/cycle
- duration in cycles is duration_time (s) / frequency (s/cycle) = cycles
- duration in radians is duration in cycles * 2pi radians/cycle

so I think `duration_radians = (duration_time / frequency) * 2 * np.pi`

examples/analog_out/voltage_update.py
WayneDroid: Noted, it's updated.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
WayneDroid: all 3 cont examples are updated as per suggestion.

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
WayneDroid: Thank you for the explanation and the suggestion. I run the example with the latest changes on physical module (PXIe-6363), the scope shows sine wave signals are generated for 1 second duration and then flat line for 10 seconds (sometimes longer) until another sine wave signals are generated.
I am unfamiliar with non-regeneration example, I'm not sure is the outcome from this example is correct or not. Somehow, I don't get same sine wave signal from LabVIEW example (Voltage (non-regeneration) - Continuous Output.vi).

Generated signal from python example:
image

Generated signal from LabVIEW example:
image

examples/analog_out/cont_gen_voltage_wfm_int_clk_non_regen.py
```diff
@@ -0,0 +1,104 @@
...
+                    sampling_rate=actual_sampling_rate,
+                    phase_in=phase_in,
+                    number_of_samples=number_of_samples,
+                )
+                task.write(data)
+                if is_first_run:
+                    is_first_run = False
+                    task.start()
+
+                phase_in = phase_out
```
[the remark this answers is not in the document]
WayneDroid: @bkeryan really appreciate your help on fixing the sine wave generation function. All continuous sine wave examples are updated.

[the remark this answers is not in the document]
bkeryan: Thank you for your patience. Now they behave like the corresponding LabVIEW examples.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,72 @@
...
+from nidaqmx.constants import AcquisitionType
+
+
+def generate_sine_wave(
+    frequency: float,
+    amplitude: float,
+    sampling_rate: float,
+    phase_in: float,
+    number_of_samples: int,
+) -> Tuple[numpy.typing.NDArray[numpy.double], float]:
```
bkeryan, line 22: Since only one example cares about phase, please make phase_in optional:
- Swap number_of_samples and phase_in parameters
- Specify a default argument for phase_in
- Do not use typing.Optional, that means "can be None" and is not needed to default to 0.0
- Update all copies of this function

``` 
def generate_sine_wave(
    frequency: float,
    amplitude: float,
    sampling_rate: float,
    number_of_samples: int,
    phase_in: float = 0.0,
) -> Tuple[numpy.typing.NDArray[numpy.double], float]:
```

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
```diff
@@ -0,0 +1,72 @@
...
+        actual_sampling_rate = task.timing.samp_clk_rate
+        print(f"Actual sampling rate: {actual_sampling_rate:g} S/s")
+
+        data, phase = generate_sine_wave(
+            frequency=10.0,
+            amplitude=1.0,
+            sampling_rate=actual_sampling_rate,
+            phase_in=phase,
+            number_of_samples=number_of_samples,
+        )
```
bkeryan, line 62: This example doesn't care about phase.

```suggestion
        task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
        task.timing.cfg_samp_clk_timing(sampling_rate, sample_mode=AcquisitionType.CONTINUOUS)

        actual_sampling_rate = task.timing.samp_clk_rate
        print(f"Actual sampling rate: {actual_sampling_rate:g} S/s")

        data, _ = generate_sine_wave(
            frequency=10.0,
            amplitude=1.0,
            sampling_rate=actual_sampling_rate,
            number_of_samples=number_of_samples,
        )
```

examples/analog_out/cont_gen_voltage_wfm_int_clk_every_n_samples_event.py
```diff
@@ -0,0 +1,86 @@
...
+        actual_sampling_rate = task.timing.samp_clk_rate
+        print(f"Actual sampling rate: {actual_sampling_rate:g} S/s")
+
+        data, phase = generate_sine_wave(
+            frequency=10.0,
+            amplitude=1.0,
+            sampling_rate=actual_sampling_rate,
+            phase_in=phase,
+            number_of_samples=number_of_samples,
+        )
```
bkeryan, line 75: This example doesn't care about phase.

```suggestion
 
        def callback(task_handle, every_n_samples_event_type, number_of_samples, callback_data):
            """Callback function for Transferred N samples."""
            nonlocal total_write
            total_write += number_of_samples
            print(f"Transferred data: {number_of_samples} samples. Total {total_write}.", end="\r")

            return 0

        task.ao_channels.add_ao_voltage_chan("Dev1/ao0")
        task.timing.cfg_samp_clk_timing(sampling_rate, sample_mode=AcquisitionType.CONTINUOUS)
        task.register_every_n_samples_transferred_from_buffer_event(1000, callback)

        actual_sampling_rate = task.timing.samp_clk_rate
        print(f"Actual sampling rate: {actual_sampling_rate:g} S/s")

        data, _ = generate_sine_wave(
            frequency=10.0,
            amplitude=1.0,
            sampling_rate=actual_sampling_rate,
            number_of_samples=number_of_samples,
        )
```

bkeryan (approved):
Approved with suggestions

examples/analog_out/cont_gen_voltage_wfm_int_clk_every_n_samples_event.py
WayneDroid: The example is updated as per suggestion.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
WayneDroid: `phase_in` parameter is set as optional for all examples.

examples/analog_out/cont_gen_voltage_wfm_int_clk.py
```diff
@@ -17,17 +17,17 @@ def generate_sine_wave(
...
 ) -> Tuple[numpy.typing.NDArray[numpy.double], float]:
     """Generates a sine wave with a specified phase.
 
     Args:
         frequency (float): Specifies the frequency of the sine wave.
         amplitude (float): Specifies the amplitude of the sine wave.
         sampling_rate (float): Specifies the sampling rate of the sine wave.
-        phase_in (float): Specifies the phase of the sine wave in radians.
         number_of_samples (int): Specifies the number of samples to generate.
+        phase_in (Optional[float]): Specifies the phase of the sine wave in radians.
```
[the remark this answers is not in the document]
WayneDroid: Types have been removed from docstring, I did not get your comment right in the first place, sorry about that.

zhindes (approved):

PR #594, opened 2024-06-12
Add examples to demonstrate TDMS logging and data plotting

WayneDroid:

- add nptdms and matplotlib packages
- upgrade `poetry.lock`
- add examples for logging data in TDMS file and plotting data

Adding TDMS and matplotlib examples to align the example demonstration in readme in another PR.

run `poetry run pytest tests/ -q` - pass

image

### DIFF
--- a/examples/analog_in/voltage_acq_int_clk_plot_data.py
+++ b/examples/analog_in/voltage_acq_int_clk_plot_data.py
@@ -1,8 +1,8 @@
 ﻿"""Example of generating visualizations for acquired data.
 
 This example demonstrates how to plot the acquired data.
-To run this example, matplotlib module is required to be installed.
-Run 'pip install matplotlib' to install matplotlib module.
+This example requires the matplotlib module.
+Run 'pip install matplotlib' to install the matplotlib module.
 """
 
 import matplotlib.pyplot as plot
--- a/examples/analog_in/voltage_acq_int_clk_plot_data.py
+++ b/examples/analog_in/voltage_acq_int_clk_plot_data.py
@@ -3,7 +3,7 @@
 This example demonstrates how to plot the acquired data.
 """
 
-import matplotlib.pyplot as plt
+import matplotlib.pyplot as plot
 
 import nidaqmx
 from nidaqmx.constants import READ_ALL_AVAILABLE, AcquisitionType
@@ -14,7 +14,7 @@ with nidaqmx.Task() as task:
 
     data = task.read(READ_ALL_AVAILABLE)
 
-    plt.plot(data)
-    plt.ylabel("Amplitude")
-    plt.title("Waveform")
-    plt.show()
+    plot.plot(data)
+    plot.ylabel("Amplitude")
+    plot.title("Waveform")
+    plot.show()
--- a/examples/analog_in/voltage_acq_int_clk_plot_data.py
+++ b/examples/analog_in/voltage_acq_int_clk_plot_data.py
@@ -1,6 +1,8 @@
 ﻿"""Example of generating visualizations for acquired data.
 
 This example demonstrates how to plot the acquired data.
+This example requires matplotlib module to be installed.
+Run 'pip install matplotlib' to install matplotlib module.
 """
 
 import matplotlib.pyplot as plot
--- a/examples/analog_in/voltage_acq_int_clk_plot_data.py
+++ b/examples/analog_in/voltage_acq_int_clk_plot_data.py
@@ -1,7 +1,7 @@
 ﻿"""Example of generating visualizations for acquired data.
 
 This example demonstrates how to plot the acquired data.
-This example requires matplotlib module to be installed.
+To run this example, matplotlib module is required to be installed.
 Run 'pip install matplotlib' to install matplotlib module.
 """
 
--- a/examples/analog_in/voltage_acq_int_clk_plot_data.py
+++ b/examples/analog_in/voltage_acq_int_clk_plot_data.py
@@ -1,0 +1,20 @@
+﻿"""Example of generating visualizations for acquired data.
+
+This example demonstrates how to plot the acquired data.
+"""
+
+import matplotlib.pyplot as plt
+
+import nidaqmx
+from nidaqmx.constants import READ_ALL_AVAILABLE, AcquisitionType
+
+with nidaqmx.Task() as task:
+    task.ai_channels.add_ai_voltage_chan("Dev1/ai0")
+    task.timing.cfg_samp_clk_timing(1000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=50)
+
+    data = task.read(READ_ALL_AVAILABLE)
+
+    plt.plot(data)
+    plt.ylabel("Amplitude")
+    plt.title("Waveform")
+    plt.show()
--- a/examples/analog_in/voltage_acq_int_clk_tdms_logging.py
+++ b/examples/analog_in/voltage_acq_int_clk_tdms_logging.py
@@ -1,0 +1,38 @@
+﻿"""Example of logging acquired data to TDMS file and read back.
+
+This example demonstrates how to log the acquired data to a TDMS file
+and then read the data from the file.
+"""
+
+import os
+
+from nptdms import TdmsFile
+
+import nidaqmx
+from nidaqmx.constants import (
+    READ_ALL_AVAILABLE,
+    AcquisitionType,
+    LoggingMode,
+    LoggingOperation,
+)
+
+with nidaqmx.Task() as task:
+    task.ai_channels.add_ai_voltage_chan("Dev1/ai0")
+    task.timing.cfg_samp_clk_timing(1000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=10)
+    task._in_stream.configure_logging(
+        "TestData.tdms", LoggingMode.LOG_AND_READ, operation=LoggingOperation.CREATE_OR_REPLACE
+    )
+
+    task.read(READ_ALL_AVAILABLE)
+
+with TdmsFile.open("TestData.tdms") as tdms_file:
+    for group in tdms_file.groups():
+        for channel in group.channels():
+            data = channel[:]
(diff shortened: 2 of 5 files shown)


examples/analog_in/voltage_acq_int_clk_plot_data.py
```diff
@@ -0,0 +1,20 @@
+﻿"""Example of generating visualizations for acquired data.
+
+This example demonstrates how to plot the acquired data.
```
zhindes, line 3: I would put a blurb mentioning that this requires pip installing matplotlib.

examples/analog_in/voltage_acq_int_clk_plot_data.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+
+import matplotlib.pyplot as plt
```
zhindes, line 6: Is `plt` a common shorthand in matplotlib? If not, I prefer `plot`

examples/analog_in/voltage_acq_int_clk_tdms_logging.py
```diff
@@ -0,0 +1,38 @@
+﻿"""Example of logging acquired data to TDMS file and read back.
+
+This example demonstrates how to log the acquired data to a TDMS file
+and then read the data from the file.
```
zhindes, line 4: Similar to above, mention what lib to pip install

pyproject.toml
```diff
@@ -49,6 +49,8 @@ toml = {version=">=0.10.2", optional=true}
 tzlocal = "^5.0"
 python-decouple = ">=3.8"
 click = ">=8.0.0"
+nptdms = "^1.9.0"
+matplotlib = {version=">=3.9.0", python = ">=3.9"}
```
zhindes, line 53: Oh, I see. Yeah, we shouldn't add dependencies here for examples. Instead just let the example document what a user should install.

examples/analog_in/voltage_acq_int_clk_plot_data.py
zhindes: Ooh, better yet, you could try/catch the import and print a message if it fails!

tests/acceptance/test_examples.py
```diff
@@ -33,6 +33,8 @@ def test___shipping_example___run___no_errors(example_path: Path, system):
         pytest.skip("Example times out if there is no trigger signal.")
     if re.search(r"\binput\(|\bKeyboardInterrupt\b", example_source):
         pytest.skip("Example waits for keyboard input.")
+    if re.search(r"\bmatplotlib\b", example_source):
```
zhindes, line 36: need to add a tdms blob here, since I don't want to add those deps

zhindes (changes requested):

pyproject.toml
bkeryan: You can add a `[tool.poetry.group.example.dependencies]` group for 3rd party dependencies that are only needed by example programs. This will install these dependencies when you run `poetry install` but not when you run install `pip install nidaqmx` or `poetry install --only main`.

pyproject.toml
```diff
@@ -49,6 +49,8 @@ toml = {version=">=0.10.2", optional=true}
 tzlocal = "^5.0"
 python-decouple = ">=3.8"
 click = ">=8.0.0"
+nptdms = "^1.9.0"
```
bkeryan, line 52: Use `>=` unless you have reason to believe 2.0 will break compatibility

bkeryan (approved):
Approved with suggestions:
- Add dependencies to a separate Poetry group

pyproject.toml
bkeryan: BTW the difference is that `[tool.poetry.dependencies]` defines what packages `nidaqmx` depends on, whereas `[tool.poetry.group.*.dependencies]` defines additional dependency groups that are only used when developing nidaqmx-python.

pyproject.toml
bkeryan: Using `^` means "I don't want to automatically update to 2.0".

It is unfortunate that `poetry add` uses `^` by default.

examples/analog_in/voltage_acq_int_clk_plot_data.py
WayneDroid: We will add matplotlib and nptdms packages as dependencies for examples, do we still need to mention the `pip install matplotlib` in this example? And the try/catch for importing the library, I think maybe all these things are not needed anymore, right?

zhindes: Yeah, I think the try/catch is a good idea. I expect people will just copy/paste the examples and try to run them.

bkeryan: I think a comment is sufficient. If you add a try/catch, people are likely to copy/paste it into their code.

zhindes: That's fair. A comment is fine by me.

WayneDroid: It seems like matplotlib is not supported on python 32-bit, the CI win-10-py32 shows 

```
PoetryException

  Failed to install C:/Users/ec2-user/AppData/Local/pypoetry/Cache/artifacts/18/be/b2/c68f3e7f22c04f84a1e658deab0e526a2f71c8a221cb585f3616afd348/matplotlib-3.9.0.tar.gz
```

I get the similar error after I switch to python 32-bit on my test machine when I execute poetry install.

```
- Installing matplotlib (3.9.0): Failed

  ChefBuildError

  Backend subprocess exited when trying to invoke build_wheel
  
  + meson setup C:\Users\jocheah\AppData\Local\Temp\tmp_fxi70na\matplotlib-3.9.0 C:\Users\jocheah\AppData\Local\Temp\tmp_fxi70na\matplotlib-3.9.0\.mesonpy-_r8xcjka -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md --native-file=C:\Users\jocheah\AppData\Local\Temp\tmp_fxi70na\matplotlib-3.9.0\.mesonpy-_r8xcjka\meson-python-native-file.ini
  The Meson build system
  Version: 1.4.1
  Source dir: C:\Users\jocheah\AppData\Local\Temp\tmp_fxi70na\matplotlib-3.9.0
  Build dir: C:\Users\jocheah\AppData\Local\Temp\tmp_fxi70na\matplotlib-3.9.0\.mesonpy-_r8xcjka
  Build type: native build
  Program python3 found: YES
  Project name: matplotlib
  Project version: 3.9.0
  Activating VS 17.8.4
  C compiler for the host machine: cl (msvc 19.38.33134 "Microsoft (R) C/C++ Optimizing Compiler Version 19.38.33134 for x64")
  C linker for the host machine: link link 14.38.33134.0
  C++ compiler for the host machine: cl (msvc 19.38.33134 "Microsoft (R) C/C++ Optimizing Compiler Version 19.38.33134 for x64")
  C++ linker for the host machine: link link 14.38.33134.0
  Host machine cpu family: x86_64
  Host machine cpu: x86_64
  Program python found: YES (C:\Users\jocheah\AppData\Local\Temp\tmp17r72o4i\.venv\Scripts\python.exe)
  Need python for x86_64, but found x86
  Run-time dependency python found: NO (tried sysconfig)
```

I can't find a way to specify the matplotlib package is only installed for python 64-bit in pyproject.toml file. I think I have to remove matplotlib from pyproject.toml file, and then skip the example from pytest. In the docstring, remind users to pip install matplotlib in order to run this example.

bkeryan: I found the matplotlib commit that stopped building Windows 32-bit wheels: https://github.com/matplotlib/matplotlib/commit/da5aaeb1e2d6f98bbbb2416ed102c9e9bc361da7

I asked @mshafer-NI and we couldn't come up with a way to specify this either, since https://peps.python.org/pep-0508/#environment-markers doesn't distinguish between 32-bit and 64-bit Windows.

How about making the `example` group optional? You would run `poetry install --with example` to install it. It probably makes sense to rename it to `examples`.
```python
[tool.poetry.group.examples]
# The "example" group is optional because matplotlib does not distribute wheels for 32-bit Windows.
optional = true

[tool.poetry.group.examples.dependencies]
matplotlib = {version=">=3.9.0", python = ">=3.9"}
nptdms = ">=1.9.0"
```

examples/analog_in/voltage_acq_int_clk_tdms_logging.py
```diff
@@ -0,0 +1,40 @@
...
+To run this example, nptdms module is required to be installed.
+Run 'pip install nptdms' to install nptdms module.
```
bkeryan, line 6:
```suggestion
This example requires the nptdms module.
Run 'pip install nptdms' to install the nptdms module.
```

pyproject.toml
bkeryan: Correction: plural `examples` makes more sense when writing `poetry install --with examples` etc.

tests/acceptance/test_examples.py
WayneDroid: noted, tdms logging example is skipped from pytest.

poetry.lock
```diff
@@ -1005,58 +1351,67 @@ files = [
 
 [[package]]
 name = "numpy"
-version = "1.26.4"
+version = "2.0.0"
```
bkeryan, line 1354: FYI @zhindes If we were using Renovate for this repo, it would automatically run the PR build & tests when upstream dependencies are updated. (Or the CI build, if we configured it to bypass PR policy.)

As-is, we basically only update poetry.lock when someone changes something unrelated, like this PR.

zhindes (approved):

poetry.lock
WayneDroid: @bkeryan Should we keep this discussion open?

bkeryan: No. I already fixed the test failures with NumPy 2.0, so this isn't breaking your PR build. I'll file a GitHub issue about moving to Renovate.

WayneDroid: Noted, I will close this conversation.

PR #645, opened 2024-10-16
nidaqmx: Fix installation with Python 3.13

### LINKED ISSUE #644: nidaqmx doesn't support Python 3.13+
To reproduce:
* Install Python 3.13
* `pip install nidaqmx`
* Run some of the shipping examples

At the very least I ran into this:
```
  File "C:\Users\User\dev\junyinta-nimbus-benchmark\ramp_up.py", line 5, in <module>
    import nidaqmx
  File "C:\Users\User\AppData\Local\pypoetry\Cache\virtualenvs\benchmark-PLHonyti-py3.13\Lib\site-packages\nidaqmx\__init__.py", line 3, in <module>
    from nidaqmx.scale import Scale
  File "C:\Users\User\AppData\Local\pypoetry\Cache\virtualenvs\benchmark-PLHonyti-py3.13\Lib\site-packages\nidaqmx\scale.py", line 3, in <module>
    import numpy
ModuleNotFoundError: No module named 'numpy'
```

That is because our numpy dependency excludes Python 3.13+ [here](https://github.com/ni/nidaqmx-python/blob/41cb6ba35c0781dec34d2e4fc91e23b22b813135/pyproject.toml#L35). There may be other issues, though. We should officially support and test Python 3.13+

bkeryan:

Update the dependency versions in `pyproject.toml` to be compatible with Python 3.13. For per-Python version constraints, make sure the last version is `^3.13` so that it will also cover 3.14, etc.

Loosen `protobuf` version constraint to allow `protobuf` 5.x because `grpcio-tools` 1.67 requires `protobuf` 5.x.

Enable running unit tests with Python 3.13 for the PR/CI workflows. I didn't do this for system tests yet because we need to update our self-hosted runner image.

Fix a couple of system test failures:
- A couple of example tests were failing due to new examples that time out if there is no signal. We didn't catch these before because they are skipped in the PR/CI workflows.
- `test_in_stream.py` had incorrect constants for +/-full-scale int16 values, which causes an overflow error with the updated `numpy`.

Partially fixes #644 

`poetry run pytest -v`

```
=============================================== short test summary info ===============================================
FAILED tests/component/system/test_device_properties.py::test___ext_cal_last_date_and_time___no_errors[library_init_kwargs] - assert 2009 > 2009
FAILED tests/component/system/test_device_properties.py::test___ext_cal_last_date_and_time___no_errors[grpc_init_kwargs] - assert 2009 > 2009
========================= 2 failed, 2286 passed, 48 skipped, 11 xfailed in 200.26s (0:03:20) ==========================
```
The failures are expected, I have a really old X Series board that has never been recalibrated.


zhindes (approved):

pyproject.toml
```diff
@@ -43,7 +44,7 @@ deprecation = ">=2.1"
 Sphinx = {version=">=5.0", optional=true}
 sphinx_rtd_theme = {version=">=1.0", optional=true}
 grpcio = {version=">=1.49.0,<2.0", optional=true}
-protobuf = {version="^4.21", optional=true}
+protobuf = {version=">=4.21", optional=true}
```
bkeryan, line 47: types-protobuf also needs >=

Issue #556, opened 2024-04-02
PR build should run ni-python-styleguide on all handwritten code


bkeryan:
`src/handwritten` is currently excluded from both `[tool.black]` and `[tool.ni-python-styleguide]`: https://github.com/ni/nidaqmx-python/blob/master/pyproject.toml#L77

```
[tool.black]
line-length = 100
extend_exclude = ".tox/|docs/|generated/|src/codegen/metadata/|src/codegen/templates/|src/handwritten/"

[tool.ni-python-styleguide]
extend_exclude = ".tox,docs,generated,src/codegen/metadata,src/codegen/templates,src/handwritten"
```

We should:
- Remove this exclusion
- Run `ni-python-styleguide fix`
- Run `ni-python-styleguide acknowledge-existing-violations`
- Update the copy in `generated/nidaqmx`

PR #664, opened 2024-12-13
add slots to DAQmx-property-containing classes

### LINKED ISSUE #656: Missing usage of slots in classes with DAQmx attributes
A few classes that contain DAQmx attributes are missing slots declarations. This can cause confusion if a user typo's one of our attribute names and then nothing useful happens.

For example, this code doesn't do what you expect, because the attribute is called `samp_quant_samp_per_chan` even though the `cfg_implicit_timing` function variable name is `samps_per_chan` for the same setting. Fun!

```python
with nidaqmx.Task() as co_task:
    co_task.co_channels.add_co_pulse_chan_freq("Dev1/ctr0", freq=10000, duty_cycle=0.5)
    co_task.co_channels.all.co_pulse_term = ""
    co_task.timing.cfg_implicit_timing(sample_mode=nidaqmx.constants.AcquisitionType.FINITE, samps_per_chan=256)

    for i in range(0, 100):
        number_of_samples = random.randint(2, 256)

        # Update the CO task
        co_task.timing.samps_per_chan = number_of_samples # <-- THIS IS DOING NOTHING!
        print(f"Number of samples: {co_task.timing.samps_per_chan}")
        co_task.start()
        co_task.wait_until_done()
        co_task.stop()
```

Templates that have attributes that don't have slots (bad!):
* https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/system/_watchdog_modules/expiration_state.py.mako
* https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/system/system.py.mako
* https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/system/watchdog.py.mako
* All of https://github.com/ni/nidaqmx-python/tree/master/src/codegen/templates/task/triggering
* https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/task/_export_signals.py.mako
* https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/task/_in_stream.py.mako
* https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/task/_out_stream.py.mako
* https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/task/_timing.py.mako
* Not a template, but https://github.com/ni/nidaqmx-python/blob/master/src/handwritten/task/_task.py

There are some iffy ones... e.g., collections don't have properties, only functions. But maybe they should have slots not to confuse people? idk. https://github.com/ni/nidaqmx-python/blob/master/src/codegen/templates/task/collections/_ai_channel_collection.py.mako. I think starting with that list above is good enough for now.

Note: this was create in response to an internal issue: [Bug 2924340](https://dev.azure.com/ni/DevCentral/_workitems/edit/2924340): Setting non-existent attribute on the CO timing does not cause assertion in python

zhindes:

Prevents inadvertent misuse of the DAQmx API by raising an `AttributeError` if a user attempts to use a DAQmx Property that doesn't exist. Closes #656.

This avoids really tricky bugs.

* Added tests for all the modified attribute-containing classes.
* Added tests for weakref support on `Task` and `WatchdogTask`, since we want to maintain that.
* Ran the rest of existing tests.


tests/component/system/test_system.py
```diff
@@ -121,3 +121,8 @@ def test_invalid_power_up_states___set_analog_power_up_states___throws_invalid_a
             system.set_analog_power_up_states(device_name, power_up_states)
 
     assert exc_info.value.error_code == DAQmxErrors.INVALID_ATTRIBUTE_VALUE
+
+
+def test___system___set_nonexistent_property___raises_exception(system):
+    with pytest.raises(AttributeError):
+        system.nonexistent_property = "foo"
```
bkeryan, line 128: Use type hints in new code.
```suggestion
def test___system___set_nonexistent_property___raises_exception(system: nidaqmx.system.System):
    with pytest.raises(AttributeError):
        system.nonexistent_property = "foo"  # type: ignore[attr-defined]
```

src/handwritten/task/_task.py
```diff
@@ -55,6 +55,10 @@ class Task:
     """
     Represents a DAQmx Task.
     """
+    __slots__ = ('_handle', '_close_on_exit', '_saved_name', '_grpc_options', '_event_handlers', '_interpreter',
```
bkeryan, line 58: FYI, we should reformat this code with Black some time (not now) - https://github.com/ni/nidaqmx-python/issues/556

tests/component/test_watchdog.py
```diff
@@ -88,3 +91,45 @@ def test___watchdog_task___clear_expiration___no_error(
...
+    watchdog_task = generate_watchdog_task(f"{sim_9189_device.name}", timeout=0.8)
+    expir_states = [
+        AOExpirationState(
+            physical_channel=sim_9263_device.ao_physical_chans[0].name,
+            expiration_state=0.0,
+            output_type=WatchdogAOExpirState.VOLTAGE,
+        )
+    ]
+
+    watchdog_task.cfg_watchdog_ao_expir_states(expir_states)
```
bkeryan, line 130: cfg_watchdog_ao_expir_states is part of arrange, the pytest.raises / nonexistent_property= is the combined act/assert section.
```suggestion
    ]
    watchdog_task.cfg_watchdog_ao_expir_states(expir_states)
```

bkeryan (approved):
Approved with suggestions

PR #674, opened 2025-02-24
New Example for Time Based Synchronization

MAnoushe-NI:

~~I've added tests applicable for this pull request~~

Adds an example for Time-Based Synchronization

No examples in nidaqmx python for this.

Confirmed working with cDAQ and AI Module

### DIFF
--- a/examples/nidaqmx - time tased synchronization.py
+++ b/examples/nidaqmx - time tased synchronization.py
@@ -2,16 +2,15 @@ import nidaqmx
 import time
 from nidaqmx.constants import AcquisitionType
 from datetime import timedelta, datetime, timezone
-import matplotlib.pyplot as plt
 
 # Function to perform the measurement
 def perform_measurement():
     with nidaqmx.Task() as task:
-        task.ai_channels.add_ai_voltage_chan("cDAQ9189-1DF40B6Mod8/ai0")
+        task.ai_channels.add_ai_voltage_chan("Dev1/ai0")
         
         # Configure finite sampling
-        num_samples = 10000
-        task.timing.cfg_samp_clk_timing(10000, sample_mode=AcquisitionType.FINITE, samps_per_chan=num_samples)
+        num_samples = 1000
+        task.timing.cfg_samp_clk_timing(1000, sample_mode=AcquisitionType.FINITE, samps_per_chan=num_samples)
         
         # Calculate the start time (1 minute from now)
         start_time = datetime.now(timezone.utc) + timedelta(minutes=1)
@@ -23,20 +22,13 @@ def perform_measurement():
         print("Starting the task...")
         task.start()
 
-        print(nidaqmx.__version__)
+        print("Waiting for the scheduled start time...")
         # Wait until the scheduled start time
         while datetime.now(timezone.utc) < start_time:
             time.sleep(1)
         
         # Read data after the task starts
         data = task.read(number_of_samples_per_channel=num_samples)
-
-        # Plot the data
-        plt.plot(data)
-        plt.title('Measurement Data')
-        plt.xlabel('Sample Number')
-        plt.ylabel('Voltage')
-        plt.show()
 
 # Perform the measurement
 perform_measurement()
--- a/examples/nidaqmx - time tased synchronization.py
+++ b/examples/nidaqmx - time tased synchronization.py
@@ -1,0 +1,42 @@
+import nidaqmx
+import time
+from nidaqmx.constants import AcquisitionType
+from datetime import timedelta, datetime, timezone
+import matplotlib.pyplot as plt
+
+# Function to perform the measurement
+def perform_measurement():
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan("cDAQ9189-1DF40B6Mod8/ai0")
+        
+        # Configure finite sampling
+        num_samples = 10000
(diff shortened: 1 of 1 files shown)


examples/nidaqmx - time tased synchronization.py
```diff
@@ -0,0 +1,34 @@
...
+        print("Waiting for the scheduled start time...")
+        # Wait until the scheduled start time
+        while datetime.now(timezone.utc) < start_time:
+            time.sleep(1)
+        
+        # Read data after the task starts
+        data = task.read(number_of_samples_per_channel=num_samples)
+
+# Perform the measurement
+perform_measurement()
```
zhindes, line 34: Our examples usually prefer simplicity. I would remove this function and simply put your `with ... task:` block at the top-level

examples/nidaqmx - time tased synchronization.py
```diff
@@ -0,0 +1,34 @@
...
+        print("Starting the task...")
+        task.start()
+
+        print("Waiting for the scheduled start time...")
+        # Wait until the scheduled start time
+        while datetime.now(timezone.utc) < start_time:
+            time.sleep(1)
+        
+        # Read data after the task starts
+        data = task.read(number_of_samples_per_channel=num_samples)
```
zhindes, line 31: This is a more standard read from a finite task:

```suggestion
        data = task.read(number_of_samples_per_channel=num_samples)
```

examples/nidaqmx - time tased synchronization.py
```diff
@@ -0,0 +1,34 @@
...
+from datetime import timedelta, datetime, timezone
+
+# Function to perform the measurement
+def perform_measurement():
+    with nidaqmx.Task() as task:
+        task.ai_channels.add_ai_voltage_chan("Dev1/ai0")
+        
+        # Configure finite sampling
+        num_samples = 1000
+        task.timing.cfg_samp_clk_timing(1000, sample_mode=AcquisitionType.FINITE, samps_per_chan=num_samples)
```
zhindes, line 13: With my suggested change to the read, this can be simplified

```suggestion
        num_samples = 1000
        task.timing.cfg_samp_clk_timing(1000, sample_mode=AcquisitionType.FINITE, samps_per_chan=1000)
```

examples/nidaqmx - time tased synchronization.py
```diff
@@ -0,0 +1,34 @@
...
+        
+        # Calculate the start time (1 minute from now)
+        start_time = datetime.now(timezone.utc) + timedelta(minutes=1)
```
zhindes, line 16: 1 minute is a lot...

```suggestion
        # Calculate the start time (10 seconds from now)
        start_time = datetime.now(timezone.utc) + timedelta(seconds=10)
```

examples/nidaqmx - time tased synchronization.py
```diff
@@ -0,0 +1,34 @@
...
+        print(f"Scheduled start time: {start_time}")
+
+        # Configure the time start trigger
+        task.triggers.start_trigger.cfg_time_start_trig(start_time)
+
+        print("Starting the task...")
+        task.start()
+
+        print("Waiting for the scheduled start time...")
+        # Wait until the scheduled start time
```
zhindes, line 26: Two things:
1. Generally printing commentary isn't really necessary, but for something that must wait, I like it. But let's not overdo it.
2. You don't need comments that repeat the code. I like the comment describing how the time delta stuff works out. These feel a bit less useful.
```suggestion
        print(f"Scheduled start time: {start_time}")

        task.triggers.start_trigger.cfg_time_start_trig(start_time)
        task.start()

        print("Waiting for the scheduled start time...")
```

zhindes (changes requested):

examples/nidaqmx - time tased synchronization.py
zhindes: Note you'll need to import READ_ALL_AVAILABLE or fully specify it

examples/nidaqmx - time tased synchronization.py
bkeryan, line 16: FYI: I was going to ask whether it's really necessary to specify UTC here, but then I looked up whether `datetime.now()` returns an "aware" datetime and now I am sad.

https://discuss.python.org/t/datetime-api-no-atomic-way-to-get-aware-current-local-datetime-when-moving-between-time-zones/55041/8
https://stackoverflow.com/questions/24281525/what-is-the-point-of-a-naive-datetime

examples/nidaqmx - time tased synchronization.py
```diff
@@ -0,0 +1,34 @@
...
+        while datetime.now(timezone.utc) < start_time:
+            time.sleep(1)
```
bkeryan, line 28: There is a DAQmx API for this, `Task.wait_for_valid_timestamp`, so you don't need to poll. 
https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/task/_task.py#L1028

I think you can wait for either `TimestampEvent.FIRST_SAMPLE` or `TimestampEvent.START_TRIGGER`.

I assume you are doing this to avoid getting a read timeout. FYI, another solution is to increase the read timeout to account for the scheduled start time, but that's probably less clear.

bkeryan (changes requested):

PR #677, opened 2025-03-04
Adding new ID Pin API introduced in mioDAQ

tjying95:
   - [Update nidaqmxbot driver feed](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/924158)

Adding new ID Pin API support introduced in mioDAQ.

Depends on the following PRs:
- [Adding new ID Pin API introduced in mioDAQ #1154](https://github.com/ni/grpc-device/pull/1154)
- [Add new attribute and APIs for mioDAQ ID Pin feature #262](https://github.com/ni/grpc-device-scrapigen/pull/262)

This is needed to enable mioDAQ ID Pin python API support.

Ran the test with a latest 25.0 DAQmx stack locally.

image


src/codegen/protos/nidaqmx.proto
tjying95, line 1: I am seeing a mismatch between [grpc-device](https://github.com/ni/grpc-device/blob/main/generated/nidaqmx/nidaqmx.proto) and this copy of nidaqmx.proto. The former has data_moniker where the current export doesn't. Without data_moniker.proto included here, the codegen will output bad result.

Thoughts? @zhindes @bkeryan

tjying95:
Quick question about CHANGELOG.md. Do I just create a new release version 1.1.1 and add this change there?

src/codegen/protos/nidaqmx.proto
zhindes: Looks like you added it, which is the right thing to do, I believe.

zhindes:
> Quick question about CHANGELOG.md. Do I just create a new release version 1.1.1 and add this change there?

Latest release is 1.0.2, so 1.1.0 is vNext. Just add it to that section

generated/nidaqmx/system/device.py
```diff
@@ -1253,6 +1313,54 @@ def self_test_device(self):
...
+        specified ID Pin.
+
+        Args:
+            id_pin_name (str): Is the name of the ID Pin to access (ex:
+                id0)
+
+        Returns:
+            Tuple[numpy.ndarray, int]: Returns a tuple containing
+            the data read from the ID Pin and the format code of the
+            data read.
```
zhindes, line 1353: I can't find an existing method that returns a "raw" tuple like this. We tend to use Named Tuples for a bit more type-safety. See [types.py](https://github.com/ni/nidaqmx-python/blob/601db52da9642f2a760ba331e42217d85921de10/generated/nidaqmx/types.py#L3) and [DriverVersion](https://github.com/ni/nidaqmx-python/blob/601db52da9642f2a760ba331e42217d85921de10/generated/nidaqmx/system/system.py#L73). I think that would be useful here, too. You could inline a named Tuple above this method in the mako pretty easily

generated/nidaqmx/system/device.py
```diff
@@ -177,6 +178,22 @@ def device_supports_cal(self):
 
     # endregion
 
+    # region ID Pin Hand-written Attributes
+
+    @property
+    def id_pin_mem_serial_nums(self):
+        """
+        List[str]: Indicates the serial number of the memory connected
```
zhindes, line 186: Is there a reason this isn't List[int] like [accessory_serial_nums](https://github.com/ni/nidaqmx-python/blob/601db52da9642f2a760ba331e42217d85921de10/generated/nidaqmx/system/device.py#L207). Are ID pin serial numbers more weird or something?

zhindes (changes requested):

generated/nidaqmx/system/device.py
tjying95: The serial number is 48bit long. I tried using 64bit int to represent it, but surprise surprise, LV attributes do not support returning array of int64. So this is the side effect of that.

zhindes: Ah, so its a string in C as well? Is there any way we can change C to be 64-bit number? Not sure if we have any of those in C, yet. I don't think its a huge deal, though. If it isn't "easy", I think its good as a string.

tjying95: Since this is an attribute, this is mostly codegenerated, which is about the same process as the LV attributes. I think it is a bit risky now to update the code-gen to support array of 64-bit number this late in the development cycle. So, I will keep this as array of strings.

src/handwritten/types.py
```diff
@@ -54,3 +54,9 @@
 
 # endregion
 
+# region ID Pin namedtuples
+
+IDPinMemory = collections.namedtuple(
```
bkeryan, line 59: This cluster is not the memory.

IDPinData would be more accurate, but that's confusing because one of the fields is called `data`. (The other field, `format_code`, is also data, but not that kind of data.)

Maybe IDPinContents? 

I honestly don't think using a simple tuple was so bad.

@zhindes

src/codegen/protos/data_moniker.proto
bkeryan, line 1: If any other NI Python modules are including this proto, make sure you can load both copies into the same Python interpreter at the same time without getting an error.

Check nimi-python and nisync-python?

src/handwritten/types.py
bkeryan, line 59: Write docstrings for types by putting a string constant after the type

```
IDPinMemory = collections.namedtuple(...)
"""Docstring goes here"""
```

src/handwritten/types.py
```diff
@@ -54,3 +54,9 @@
...
+    'IDPinMemory', ['data', 'format_code'])
```
bkeryan, line 60: What is data, an ndarray or a list[int]?

Please use https://docs.python.org/3/library/typing.html#typing.NamedTuple for this new named tuple.

I have a branch somewhere that updates the other named tuples to be typed, I should dig that up.

bkeryan (changes requested):

generated/nidaqmx/system/device.py
bkeryan: ok

src/codegen/protos/data_moniker.proto
tjying95: I quickly tested with some simulated devices on `nidmm` and `nidaqmx`:

```python
import nidaqmx
import nidmm


def example(resource_name, option_string, function, range, digits):
    with nidmm.Session(resource_name=resource_name, options=option_string) as session:
        session.configure_measurement_digits(measurement_function=nidmm.Function[function], range=range, resolution_digits=digits)
        print(session.read())


def test_example():
    options = {'simulate': True, 'driver_setup': {'Model': '4082', 'BoardType': 'PXIe', }, }
    example('PXI1Slot2', options, 'DC_VOLTS', 10, 6.5)
    
    local_system = nidaqmx.system.System.local()
    for device in local_system.devices:
        print(
            "Device Name: {0}, Product Category: {1}, Product Type: {2}".format(
                device.name, device.product_category, device.product_type
            )
        )
        print(device.id_pin_pin_names)


test_example()
```

image

It seems working.

bkeryan: nidmm isn't using data_moniker.proto yet. It looks like the other packages that are using it are not released yet, so this is fine.

bkeryan (changes requested):

zhindes (approved):

bkeryan (approved):

WayneDroid (approved):

PR #723, opened 2025-05-21
fix virtual channel name logic with library interpreter and DAQmx 24.5+

### LINKED ISSUE #579: nidaqmx does not generate numbered virtual channel names correctly
If a DAQmx channel is created for physicalChannel=`Dev1/ai0:7` with nameToAssignToChannel=`myChan09`, then DAQmx will assign `Dev1/ai0` the name `myChan09`, `Dev1/ai1` the name `myChan10`, and so on. In particular, notice the renumbering of the channels and the zero padding of the numbers. See https://www.ni.com/docs/en-US/bundle/ni-daqmx/page/nameassignment.html for reference.

nidaqmx calculates the virtual names incorrectly.

To demonstrate, run the following script:
```python
import nidaqmx

with nidaqmx.Task() as task:
    ai_channel = task.ai_channels.add_ai_voltage_chan("XSeries6363/ai0:3")
    print(ai_channel.name)
    print(ai_channel.chan_type)

with nidaqmx.Task() as task:
    ai_channel = task.ai_channels.add_ai_voltage_chan("XSeries6363/ai0:3", name_to_assign_to_channel="myChan09")
    print(ai_channel.name)
    print(ai_channel.chan_type)
```

It will fail on the last line with `nidaqmx.errors.DaqError: Specified channel is not in the task.`.

The output before that is
```text
XSeries6363/ai0:3
ChannelType.ANALOG_INPUT
myChan090:3
```

[AB#3099973](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3099973)

zhindes:

* Closes #579 when using the library interpreter and DAQmx 24.5 or newer
* Adds extensive tests to virtual channel name logic demonstrating how broken our existing implementation is
* Fixes a small, obvious bug in the bespoke DI/DO virtual channel name logic

Improves API consistency across ADEs

New unit tests:
* Some disabled on gRPC since our fix isn't portable
* Some xfail if executing in a version of DAQmx older than 24.5
* Validated that everything is as expected on 24.3 (old) and 24.5 (new)


tests/component/task/channels/test_di_channel.py
```diff
@@ -44,3 +44,151 @@ def test___task___add_di_chan_chan_per_line___sets_channel_attributes(
...
+            True,
+        ),
+        (
+            ["port0/line0:7"],
+            LineGrouping.CHAN_PER_LINE,
+            "myFirstChan0:5, mySecondChan0:5",
+            "myFirstChan0:5, mySecondChan0:1",
+            False,
+            True,
+        ),
```
bkeryan, line 162: giphy

tests/component/task/channels/test_di_channel.py
```diff
@@ -44,3 +44,151 @@ def test___task___add_di_chan_chan_per_line___sets_channel_attributes(
     for chan in chans:
         assert chan.chan_type == ChannelType.DIGITAL_INPUT
         assert chan.di_num_lines == 1
+
+
+@pytest.mark.library_only(
+    reason="Internal method we use for retrieving a channel name isn't supported on gRPC",
+)
```
bkeryan, line 51: The test cases with `xfail_if_old=False` should pass with gRPC.

bkeryan (approved):
Approved with suggestions

tests/component/task/channels/test_di_channel.py
zhindes: Good point... what's a good way to detect if the test case is running with grpc interpreter to xfail at runtime?

zhindes: I could just split it into 2 tests, the second batch that fails on grpc is "aggressive abuse" :D

bkeryan: grpc_xfail

bkeryan: > ...at runtime?

Oh, I see why you did this. 

Delete the `xfail_if_old` param and use `pytest.param` to mark the param data.
https://docs.pytest.org/en/stable/example/parametrize.html#set-marks-or-test-id-for-individual-parametrized-test

bkeryan: https://docs.pytest.org/en/6.2.x/skipping.html#skip-xfail-with-parametrize has another example using skipif to do runtime checks

bkeryan: Splitting the test is fine

bkeryan (approved):

PR #269, opened 2023-04-05
fix test_examples type annotations for python <3.9

zhindes:
# What does this Pull Request accomplish?

Makes `test_examples.py` usable in Python 3.7/3.8. Type hinting generics wasn't available until [PEP 585](https://peps.python.org/pep-0585/) without a `__future__` import.

# What testing has been done?

```
C:\dev\nidaqmx-python>poetry run pytest
================================================ test session starts =================================================
platform win32 -- Python 3.7.9, pytest-7.2.2, pluggy-1.0.0
...
========================================== 296 passed, 10 skipped in 50.03s ==========================================
```

PR #767, opened 2025-06-19
Add Active Dev to Timing Class

### LINKED ISSUE #692: Cannot set "ai_conv_rate" on NI DAQ 9209 due to missing active device modifier for timing attributes
To avoid ghosting during switching, we need to increase the ai_conv_rate. However, the standard command "task.timing.ai_conv_rate = 1.1 * sampling_rate * n_channels" does not work on the DAQ model 9209.

[AB#3064732](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3064732)

charitylxy:

- Update attributes.py metadata generated from [add lv-filters to attribute metadata by charitylxy · Pull Request #269 · ni/grpc-device-scrapigen](https://github.com/ni/grpc-device-scrapigen/pull/269)
- Add `_active_devs` to `Timing` class
- Implemented `__getitem__` to return a new Timing instance with `_active_devs` set to the index
- Updated the property getter/setter/deleter template to use the `_ex` interpreter method if `_active_devs` is set.
- Added more unit test coverage for timing property

- Fixes #692 

1. Ran simple program to test some of the attributes:
-  `ai_conv_rate` attribute with device specified
```
...
    task.timing["cdaqChassisTesterMod6"].ai_conv_rate = 100
    print (task.timing["cdaqChassisTesterMod6"].ai_conv_rate)
...

# Result: ai_conv_rate gets printed
```
 - `first_samp_timestamp_enable` attribute with device specified
```
...
    dev = Device("nidaqmxMultithreadingTester1")
    print (task.timing[dev].first_samp_timestamp_enable)
...

# Result: Error -201118 displayed
```
    
2. Unit Test
```
(nidaqmx-py3.10) PS C:\Dev\nidaqmx-python> poetry run pytest -k test___timing___
=========================================================================================== test session starts ===========================================================================================
platform win32 -- Python 3.10.8, pytest-8.4.1, pluggy-1.6.0
rootdir: C:\Dev\nidaqmx-python
configfile: pyproject.toml
testpaths: tests
plugins: cov-6.2.1, mock-3.14.1
collected 2514 items / 2398 deselected / 116 selected

tests\component\task\test_timing.py .s.s..ss...sss...sss.s                                                                                                                                           [ 18%]
tests\component\task\test_timing_properties.py .s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s                                                        [100%]

============================================================================= 58 passed, 58 skipped, 2398 deselected in 1.44s ============================================================================= 
```


charitylxy:
@bkeryan I'm not too sure how we should go about specifying `_active_devs` for unsupported methods. Do we want to allow specifying `_active_devs` only for `ai_conv_rate`, and the rest would error out?

src/codegen/templates/property_setter_template.py.mako
```diff
@@ -42,5 +42,12 @@
         function_call_args.append(hex(attribute.id))
         function_call_args.append('val')
     %>\
+    %if attribute.name == "ai_conv_rate":
```
charitylxy, line 45: I think the codegen template could be structured better for this case, I'm still figuring out on that.

charitylxy: I think the codegen template could be structured better for this case, I'm still figuring that out.

zhindes:
> @bkeryan I'm not too sure how we should go about specifying `_active_devs` for unsupported methods. Do we want to allow specifying `_active_devs` only for `ai_conv_rate`, and the rest would error out if `_active_devs` is specified?

Can you try that in LabVIEW with the Property Node - what does it do if you specify one with an attribute that doesn't support it?

Overall, it looks good :) Once we figure out the behavior here, let's add a couple unit tests. Add a simulated cDAQ module that supports this, and then validate whatever the error cases (or not error depending on what we decide) are.

zhindes:
> The timing properties that do not have an `Ex` version do not support active devices in the C API. The Python API should raise an exception rather than silently discard the active devices parameter. `DAQmxErrorMStudioOperationDoesNotSupportDeviceContext` is an appropriate error code, despite the fact that it has "MStudio" in the constant name:
> 
> "Operation must be performed on the entire task. It cannot be performed only on specific devices in the task. Do not use the indexer, Item property in Visual Basic, or index operator in C++ to specify device names when performing this operation."

...

> Normally I am in favor of parity with the LabVIEW API, but the underlying interfaces used by the LabVIEW API allow passing active devices to properties that do not support active devices in the C API.

That all sounds good to me!

generated/nidaqmx/task/_timing.py
```diff
@@ -12,11 +12,15 @@ class Timing:
...
-    __slots__ = ('_handle', '_interpreter')
+    __slots__ = ('_handle', '_interpreter', '_active_devs')
 
-    def __init__(self, task_handle, interpreter):
+    def __init__(self, task_handle, interpreter, active_devs=None):
         self._handle = task_handle
         self._interpreter = interpreter
+        self._active_devs = active_devs
+
+    def __getitem__(self, dev):
```
bkeryan: What type do we expect `dev` to be? `str`, `Device`, or `str | Device`?

I was hoping that `str(dev)` would convert `Device` to `str` like it does for `pathlib.Path`, but it doesn't because `Device` only implements `__repr__` and not `__str__`:
```python
>>> import nidaqmx
>>> x = nidaqmx.system.device.Device("Dev1")
>>> x
Device(name=Dev1)
>>> str(x)
'Device(name=Dev1)'
```

so I would do something like
```python
    def __init__(self, task_handle, interpreter, active_devs: str | Device | None = None):
        if isinstance(active_devs, Device):
            active_devs = active_devs.name
        self._handle = task_handle
        self._active_devs = active_devs
```

We could also consider implementing `Device.__str__`, but this would change behavior for applications that expect str to be the same as repr.

bkeryan: getitem is also used for indexing, so make sure that `task.timing[3]` either does something useful or raises a reasonable exception (and not a https://en.wikipedia.org/wiki/Garbage_in,_garbage_out exception like `AttributeError: 'NoneType' object has no attribute 'name'`)

charitylxy: Updated to expect `str `and `Device`

charitylxy: Updated to raise TypeError, if dev is not the expected type

generated/nidaqmx/task/_timing.py
```diff
@@ -12,11 +17,28 @@ class Timing:
...
+        else: 
+            raise TypeError(f"Invalid active_devs input: {dev!r} (type: {type(dev).__name__}). Expected str or Device.")
+
+    def _raise_device_context_not_supported_error(self):
+        raise DaqError(
+            'Operation must be performed on the entire task. It cannot be '
+            'performed only on specific devices in the task. Do not use the '
+            'indexer, Item property in Visual Basic, or index operator in '
+            'C++ to specify device names when performing this operation.',
+            DAQmxErrors.M_STUDIO_OPERATION_DOES_NOT_SUPPORT_DEVICE_CONTEXT)
```
zhindes: Let's remove the last sentence starting with "Do not use the indexer..."

zhindes (changes requested):
Still looking for tests!

src/codegen/templates/property_setter_template.py.mako
```diff
@@ -32,9 +32,10 @@
 \
 ## Script interpreter call.
 <%
+        active_devs_supported_attributes = get_advanced_attribute_function_name("Advanced:Timing (Active Device)")
```
bkeryan: This template is called for each attribute, so building a list of all activedev attributes and searching it is accidentally quadratic and also kind of confusing. 

It would be less confusing to either check whether "Advanced:Timing (Active Device)" is in the current attribute's lv_filter or to build the list only once.

src/codegen/templates/property_getter_template.py.mako
```diff
@@ -19,10 +19,11 @@
 \
 ## Script interpreter call.
 <%
+    active_devs_supported_attributes = get_advanced_attribute_function_name("Advanced:Timing (Active Device)")
     mapped_func_type = get_generic_attribute_function_type(attribute)
     generic_attribute_func = get_generic_attribute_function_name(attribute) + "_" + mapped_func_type
-    if attribute.name in ACTIVE_DEVS_SUPPORTED_ATTRIBUTES:
-        generic_attribute_func_ex = get_generic_attribute_function_name(attribute) + "_ex" + "_" + mapped_func_type
+    if attribute.name in active_devs_supported_attributes:
+            generic_attribute_func_ex = get_generic_attribute_function_name(attribute) + "_ex" + "_" + mapped_func_type
```
bkeryan: bad indent

src/codegen/templates/property_setter_template.py.mako
```diff
@@ -43,15 +44,22 @@
...
     %>\
-    %if attribute.name in ACTIVE_DEVS_SUPPORTED_ATTRIBUTES:
+    %if attribute.python_class_name == "Timing":
+        %if attribute.name in active_devs_supported_attributes:
         if self._active_devs:
             self._interpreter.set_${generic_attribute_func_ex}(${', '.join(function_call_args_ex)})
         else:
             self._interpreter.set_${generic_attribute_func}(${', '.join(function_call_args)})
+        %else:
+        if self._active_devs == None:
```
bkeryan: [style] Be consistent about using falsiness (`if self._active_devs:`) or explicitly comparing to None. Also, if you are going to explicitly compare to None, use `is` or `not is`, not `==` or `!=`.

src/codegen/templates/property_setter_template.py.mako
```diff
@@ -43,15 +44,22 @@
...
+            self._interpreter.set_${generic_attribute_func}(${', '.join(function_call_args)})
+        else:
+            self._raise_device_context_not_supported_error()
```
bkeryan, line 59: [style] The balanced if/else looks weird to me since one case is far less likely than the other.

I think this looks better:
```
    @change_detect_di_falling_edge_physical_chans.setter
    def change_detect_di_falling_edge_physical_chans(self, val):
        val = val.name
        if self._active_devs:
            self._raise_device_context_not_supported_error()
        self._interpreter.set_timing_attribute_string(self._handle, 0x2196, val)
```

or 
```
    @change_detect_di_falling_edge_physical_chans.setter
    def change_detect_di_falling_edge_physical_chans(self, val):
        val = val.name
        self._throw_if_has_context()
        self._interpreter.set_timing_attribute_string(self._handle, 0x2196, val)
```

charitylxy:
> Still looking for tests!
I'm trying to add a test for the timing boolean property, `ai_conv_dig_fltr_enable` is the only boolean property that allows active_devs. I've tried with multiple devices, but I keep getting a "property not supported" error. Do you know what device supports ai_conv_dig_fltr_enable? Or is there anywhere in our codebase that specifies devices and their supported attributes?

src/codegen/templates/property_setter_template.py.mako
charitylxy: Updated to `has_attribute_with_filter` which will check if attribute is in `Timing` group and `Advanced:Timing (Active Device)` is in the current attribute's `lv_filter`

zhindes (approved):

src/codegen/templates/task/_timing.py.mako
```diff
@@ -24,11 +29,26 @@ class Timing:
     """
     Represents the timing configurations for a DAQmx task.
     """
-    __slots__ = ('_handle', '_interpreter')
+    __slots__ = ('_handle', '_interpreter', '_active_devs')
 
-    def __init__(self, task_handle, interpreter):
+    def __init__(self, task_handle, interpreter, active_devs: Union[str, Device, None] = None):
```
bkeryan: Union is deprecated. In Python 3.10 we can use `str | Device | None` instead.

What about Python 3.9? Well, `from __future__ import annotations` causes Python to parse type hints as strings at run time, so you don't get errors from using `Union` in type hints with Python 3.9. You only need `Union` for non-hint usages, like `AnyFoo: TypeAlias = Union[Foo, Bar]`.

I removed most usages of `Union` in https://github.com/ni/nidaqmx-python/pull/681

src/codegen/templates/task/_timing.py.mako
```diff
@@ -24,11 +29,26 @@ class Timing:
...
+        if isinstance(active_devs, Device):
+            active_devs = active_devs.name
         self._handle = task_handle
         self._interpreter = interpreter
+        self._active_devs = active_devs
+
+    def __getitem__(self, dev):
```
bkeryan:
```suggestion
    def __getitem__(self, dev: str | Device) -> Timing:
```

This also needs `from __future__ import annotations`

bkeryan: getitem should raise an error if the timing already has active devs so you can't write `task.timing["cDAQ1Mod1"]["cDAQ1Mod2"]`

tests/component/task/test_timing_properties.py
```diff
@@ -21,6 +31,17 @@ def test___timing___set_boolean_property___returns_assigned_value(task, sim_6363
...
 
+def test___timing___set_boolean_property_with_device_context___throw_daqerror(
+    task, sim_6363_device
+):
+    task.ao_channels.add_ao_voltage_chan(sim_6363_device.ao_physical_chans[0].name)
+    task.timing.samp_timing_type = SampleTimingType.ON_DEMAND
+
+    with pytest.raises(DaqError) as e:
+        task.timing[sim_6363_device].simultaneous_ao_enable = True
+    assert e.value.error_type == DAQmxErrors.M_STUDIO_OPERATION_DOES_NOT_SUPPORT_DEVICE_CONTEXT
```
bkeryan: arrange, act, assert in separate blocks with 1 blank line between each block
```suggestion
    task.ao_channels.add_ao_voltage_chan(sim_6363_device.ao_physical_chans[0].name)
    task.timing.samp_timing_type = SampleTimingType.ON_DEMAND

    with pytest.raises(DaqError) as e:
        task.timing[sim_6363_device].simultaneous_ao_enable = True
        
    assert e.value.error_type == DAQmxErrors.M_STUDIO_OPERATION_DOES_NOT_SUPPORT_DEVICE_CONTEXT
```

`with pytest.raises` is part of act.

This applies everywhere, not just this specific test.

src/codegen/templates/task/_timing.py.mako
```diff
@@ -24,11 +29,26 @@ class Timing:
...
+        if isinstance (dev, str) or isinstance (dev, Device):
+            return Timing(self._handle, self._interpreter, active_devs=dev)
+        else: 
+            raise TypeError(f"Invalid active_devs input: {dev!r} (type: {type(dev).__name__}). Expected str or Device.")
+
+    def _raise_device_context_not_supported_error(self):
```
bkeryan:
```suggestion
    def _raise_device_context_not_supported_error(self) -> NoReturn:
```
also `from typing import NoReturn`

tests/component/task/test_timing_properties.py
bkeryan: See https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions?anchor=function-body

generated/nidaqmx/task/_timing.py
```diff
@@ -19,20 +21,24 @@ class Timing:
...
         self._active_devs = active_devs
 
-    def __getitem__(self, dev):
-        if isinstance (dev, str) or isinstance (dev, Device):
+    def __getitem__(self, dev: str | Device) -> Timing:
+        if self._active_devs:
+            raise DaqError( 
+                f"Cannot set active device '{dev}' because active device '{self._active_devs}' is already set.",
+                DAQmxErrors.UNKNOWN)
+        if isinstance(dev, str) or isinstance(dev, Device):
```
bkeryan:
```suggestion
        if isinstance(dev, (str, Device)):
```

PR #772, opened 2025-06-27
Fix dependency: hightime = ">=0.2.2"

mikeprosserni:

- ~~[ ] I've added tests applicable for this pull request~~

Updates the hightime dependency in pyproject.toml to allow hightime to be version 0.3 or higher.

`hightime = "^0.2.2"` is causing dependency issues for another project, which depends on daqmx, numpy, nitypes, grpcio-tools, protobuf, and more. The best way to fix this is to allow hightime to be 0.3.0.

```
poetry run pytest
2089 passed, 339 skipped, 27 xfailed, 5 xpassed in 151.76s (0:02:31)
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -98,13 +98,13 @@ uvloop = ["uvloop (>=0.15.2)"]
 
 [[package]]
 name = "cachetools"
-version = "6.0.0"
+version = "6.1.0"
 description = "Extensible memoizing collections and decorators"
 optional = false
 python-versions = ">=3.9"
 files = [
-    {file = "cachetools-6.0.0-py3-none-any.whl", hash = "sha256:82e73ba88f7b30228b5507dce1a1f878498fc669d972aef2dde4f3a3c24f103e"},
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -44,7 +44,7 @@ numpy = [
 deprecation = ">=2.1"
 grpcio = { version = ">=1.49.0,<2.0", optional = true }
 protobuf = { version = ">=4.21", optional = true }
-hightime = "^0.2.2"
+hightime = ">=0.2.2"
```
bkeryan, line 47: This will affect some users so please add it to the changelog.

mikeprosserni: I added "Updated hightime dependency from `^0.2.2` to `>=0.2.2` to allow for newer versions." to Major Changes under 1.2.0

poetry.lock
bkeryan, line 1: When you edit version constraints in `pyproject.toml`, use `poetry lock --no-update` to make only the required changes to `poetry.lock`. That way you don't introduce regressions to the PR/CI builds by updating 3rd party dependencies. This also reduces the likelihood of merge conflicts.

I recommend reverting `poetry.lock` and running `poetry lock --no-update` before merging.

mikeprosserni: reverted poetry.lock and used poetry lock --no-update instead (now the only diff there is the hash)

Issue #626, opened 2024-08-07
Add automated test coverage with PyPy


bkeryan:
The README.rst and trove classifiers claim support for PyPy, but there is no automated test coverage of this configuration.

In https://github.com/ni/nidaqmx-python/pull/625 I tried updating the GitHub workflows to run unit tests with PyPy, and ran into a number of problems with dependencies (mostly grpcio, grpcio-tools, and numpy).

bkeryan:
I think we should be able to get the tests running on PyPy in library mode, but not gRPC mode. We also ought to add some caveats about gRPC/PyPy compatibility to the readme.

## NumPy

NumPy appears to distribute binary wheels for one version of PyPy. For example, https://pypi.org/project/numpy/2.3.2/#files includes binary wheels for pypy3.11. 

I got the nitypes tests running with multiple versions of PyPy by specifying PyPy-specific version constraints: https://github.com/ni/nitypes-python/blob/main/pyproject.toml#L38

```
numpy = [
  { version = ">=1.22", python = ">=3.9,<3.12", markers = "implementation_name != 'pypy'" },
  { version = ">=1.26", python = ">=3.12,<3.13", markers = "implementation_name != 'pypy'" },
  { version = ">=2.1", python = "^3.13", markers = "implementation_name != 'pypy'" },
  { version = ">=2.1", python = ">=3.10,<3.11", markers = "implementation_name == 'pypy'" },
  { version = ">=2.3", python = "^3.11", markers = "implementation_name == 'pypy'" },
]
```

Note that specifying these constraints in the `[tool.poetry.dependencies]` section affects the distribution package's metadata, which may require consumers of the package to specify the same constraints. It's better to specify these constraints in a dev dependency group.

The PyPy FAQ has some interesting info about NumPy compatibility: https://doc.pypy.org/en/latest/faq.html#what-about-numpy-numpypy-micronumpy

## Protocol Buffers

The `protobuf` package has multiple implementations: pure Python, legacy pyext, and upb.

`protobuf` on PyPy appears to use the pure Python implementation:

```
PS C:\dev\nitypes-python> poetry run python
Python 3.11.11 (0253c85bf5f8, Feb 26 2025, 10:43:25)
[PyPy 7.3.19 with MSC v.1941 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>>> import google.protobuf.internal.api_implementation
>>>> google.protobuf.internal.api_implementation.Type()
[...]

PR #798, opened 2025-08-06
Add nidaqmx.stream_readers.AnalogSingleChannelReader.read_waveform()

mikeprosserni:

* Added `read_analog_waveform()` to `BaseInterpreter` and `LibraryInterpreter`
* (Make `GrpcStubInterpreter` raise `NotImplementedError` for now)
* Added `LibraryInterpreter._internal_read_analog_waveform_ex()` as a helper for calling `DAQmxInternalReadAnalogWaveformEx` from the C API 
* Added `nidaqmx.stream_readers.AnalogSingleChannelReader.read_waveform()` - reads into single waveform using `read_analog_waveform()`

Also: 

* Added `nitypes` as a dependency (and I had to tweak the details of numpy versioning accordingly)
* Added `[tool.pyright]` section in `pyproject.toml` to suppress some of the many pylance warnings in VS Code. 

This is the first step towards adding Waveform support to the nidaqmx-python api.

* Added integration test `test___analog_single_channel_reader___read_waveform___returns_valid_waveform`
* Added `examples\analog_in\voltage_acq_int_clk_wfm.py`


src/handwritten/stream_readers.py
```diff
@@ -238,6 +244,49 @@ def read_one_sample(self, timeout=10):
...
+
+        def set_wfm_attr_callback(
+            channel_index: int,
+            attribute_name: str,
+            attribute_type: WfmAttrType,
+            value: WfmAttrValue,
+            callback_data: object,
+        ) -> int:
+            assert channel_index == 0
+            extended_properties[attribute_name] = value
```
bkeryan: This callback is a library-specific implementation detail and belongs in LibraryInterpreter. The interpreter interface should use a dict/mapping/mutablemapping/extendedpropertydictionary. Try to avoid making an extra copy.

src/handwritten/stream_readers.py
```diff
@@ -238,6 +244,49 @@ def read_one_sample(self, timeout=10):
...
+            set_wfm_attr_callback,
+            None,
+            raw_data,
+        )
+
+        waveform = AnalogWaveform(dtype=numpy.double, raw_data=raw_data, extended_properties=extended_properties)
+        waveform.timing = Timing(
+            sample_interval_mode=SampleIntervalMode.REGULAR,
+            timestamp=_T0_EPOCH + datetime.timedelta(seconds=t0_array[0] * _INT64_WFM_SEC_PER_TICK),
+            sample_interval=datetime.timedelta(seconds=dt_array[0] * _INT64_WFM_SEC_PER_TICK),
```
bkeryan: Use of .NET timestamps is a library-specific implementation detail and belongs in LibraryInterpreter. The interpreter interface should use a sequence of timestamps or whatever.

src/handwritten/stream_readers.py
```diff
@@ -238,6 +244,49 @@ def read_one_sample(self, timeout=10):
...
+    def read_waveform(
+        self, number_of_samples_per_channel=READ_ALL_AVAILABLE, timeout=10
+    ) -> AnalogWaveform[numpy.double]:
+        number_of_samples_per_channel = (
+            self._task._calculate_num_samps_per_chan(
+                number_of_samples_per_channel))
+
+        t0_array = numpy.full(1, numpy.iinfo(numpy.int64).min, dtype=numpy.int64)
+        dt_array = numpy.full(1, numpy.iinfo(numpy.int64).min, dtype=numpy.int64)
+        raw_data = numpy.full(number_of_samples_per_channel, math.inf, dtype=numpy.float64)
```
bkeryan: The use of min and inf was for testing. Zeros may be more appropriate for production code.

src/handwritten/types.py
```diff
@@ -67,3 +78,13 @@ class IDPinContents(typing.NamedTuple):
...
 # endregion
+
+WfmAttrValue: TypeAlias = Union[bool, float, int, str]
+"""Waveform attribute value."""
+
+SetWfmAttrCallback: TypeAlias = Callable[[int, str, WfmAttrType, WfmAttrValue, object], int]
+"""
+Callback function for retrieving waveform attributes.
+
+Signature: channel_index, attribute_name, attribute_type, value, callback_data -> int (error code)
```
bkeryan: These are library implementation details and should not be public.

src/codegen/templates/constants.mako
```diff
@@ -24,3 +24,10 @@ class ${name}(Enum):
 
 
 % endfor
+class WfmAttrType(Enum):
+    """Waveform attribute datatype."""
+
+    BOOL32 = 1
+    FLOAT64 = 2
```
bkeryan: This is a library implementation detail and should not be public.

pyrightconfig.json
bkeryan, line 1: Move to pyproject.toml

pyproject.toml
```diff
@@ -53,6 +55,7 @@ click = [
 ]
 distro = { version = ">=1.9.0", platform = "linux" }
 requests = ">=2.25.0"
+nitypes = {version=">=0.1.0dev3", allow-prereleases=true}
```
bkeryan, line 58: We will need to release nitypes before the next nidaqmx-python release.

pyproject.toml
```diff
@@ -40,6 +40,8 @@ numpy = [
   { version = ">=1.22", python = ">=3.9,<3.12" },
   { version = ">=1.26", python = ">=3.12,<3.13" },
   { version = ">=2.1", python = "^3.13" },
+  { version = ">=2.1", python = ">=3.10,<3.11" },
+  { version = ">=2.3", python = "^3.11" },
```
bkeryan: These constraints overlap because they are missing the implementation markers.

bkeryan (changes requested):

src/handwritten/stream_readers.py
bkeryan: BTW, I guess it's possible for the gRPC interpreter to implement this callback signature, so this one is negotiable.

If the library interpreter wasn't doing the translation in `_get_wfm_attr_value`, that would be more of a problem.

generated/nidaqmx/stream_readers.py
```diff
@@ -247,6 +247,44 @@ def read_one_sample(self, timeout=10):
     def read_waveform(
         self, number_of_samples_per_channel=READ_ALL_AVAILABLE, timeout=10
     ) -> AnalogWaveform[numpy.double]:
```
bkeryan: Add nitypes to the intersphinx mapping and make sure it hyperlinks `AnalogWaveform` to the nitypes documentation.

https://github.com/ni/nidaqmx-python/blob/master/docs/conf.py#L48

https://nidaqmx-python--798.org.readthedocs.build/en/798/stream_readers.html#nidaqmx.stream_readers.AnalogSingleChannelReader.read_waveform

generated/nidaqmx/stream_readers.py
```diff
@@ -247,6 +247,44 @@ def read_one_sample(self, timeout=10):
...
+                If the task acquires samples continuously and you set
+                this input to nidaqmx.constants.READ_ALL_AVAILABLE, this
+                method reads all the samples currently available in the
+                buffer.
+
+                If the task acquires a finite number of samples and you
+                set this input to nidaqmx.constants.READ_ALL_AVAILABLE,
+                the method waits for the task to acquire all requested
+                samples, then reads those samples. If you set the
+                "read_all_avail_samp" property to True, the method reads
```
bkeryan, line 277: > add documentation (generated)

Generated from what? Is this copied from another method or do I need to compare it more closely?

mikeprosserni: In the previous commit ("add documentation"), I added this documentation to `src/handwritten/stream_readers.py`. The parameter documentation for `number_of_samples_per_channel` and `timeout` was copied verbatim from `read_many_sample()`. But I forgot to run code generation before committing this.

So, after committing "add documentation", I ran code generation (which copied the documentation to `src/generated/daqmx/stream_readers.py`) and then I committed that as "add documentation (generated)".

Sorry for the confusion.

examples/analog_in/voltage_acq_int_clk_wfm.py
```diff
@@ -0,0 +1,21 @@
...
+
+with nidaqmx.Task() as task:
+    task.ai_channels.add_ai_voltage_chan("Dev1/ai0")
+    task.timing.cfg_samp_clk_timing(1000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=50)
+
+    reader = AnalogSingleChannelReader(task.in_stream)
+    waveform = reader.read_waveform(READ_ALL_AVAILABLE)
+    print(f"Acquired data: {waveform.scaled_data}")
+    print(f"Channel name: {waveform.channel_name}")
+    print(f"Unit description: {waveform.unit_description}")
```
zhindes, line 19: FYI - we have an internal API for adding generic extended attributes to DAQmx waveforms that the Sound & Vibe toolkit uses for some of its advanced measurements. That's a LV toolkit, so its definitely not relevant here, but that is something that the S&V champion (Doug Bendele) has requested we make public for more general usage: [Idea 3111321](https://dev.azure.com/ni/InnovationCentral/_workitems/edit/3111321): DAQmx Custom Channel Attributes FTW). I don't think we'll do it any time soon, but I am happy to see that the wfm datatype lets you access the "raw" [extended_properties](https://github.com/ni/nitypes-python/blob/main/src/nitypes/waveform/_numeric.py#L485) dictionary directly! Definitely not useful adding that here, but if we ever did make that idea a reality, we can already use it in Python! Woo!

generated/nidaqmx/constants.py
```diff
@@ -1026,3 +1026,10 @@ class _TriggerUsageTypes(Enum):
     ARM_START = 32  #: Device supports arm start triggers
 
 
+class WfmAttrType(Enum):
```
zhindes: This looks internal... I don't think it should be here. It appears to be _library_interpreter specific, so maybe put it there?

generated/nidaqmx/types.py
```diff
@@ -67,3 +78,13 @@ class IDPinContents(typing.NamedTuple):
...
+
+WfmAttrValue: TypeAlias = Union[bool, float, int, str]
+"""Waveform attribute value."""
+
+SetWfmAttrCallback: TypeAlias = Callable[[int, str, WfmAttrType, WfmAttrValue, object], int]
+"""
+Callback function for retrieving waveform attributes.
+
+Signature: channel_index, attribute_name, attribute_type, value, callback_data -> int (error code)
+"""
```
zhindes: Similarly, these don't seem customer facing. Also move to _library_interpreter?

zhindes (changes requested):
I only looked at the customer-facing elements of this. I mostly love it! I had already reviewed the nitypes stuff, so this is a great first step at exposing this functionality in one of our APIs. The only issues are minor - I think you're leaking some internal types to users that we should keep hidden.

Otherwise, I think Brad has the nitty gritty under control.

examples/analog_in/voltage_acq_int_clk_wfm.py
bkeryan, line 19: Perhaps this should be renamed: https://github.com/ni/nitypes-python/issues/168

src/handwritten/stream_readers.py
mikeprosserni: I refactored this so that LibraryInterpreter.internal_read_analog_waveform_ex() returns a sequence of datetimes and a sequence of timestamps.

src/handwritten/stream_readers.py
mikeprosserni: I moved set_wfm_attr_callback into LibraryInterpreter. Now LibraryInterpreter.internal_read_analog_waveform_ex() returns `Sequence[Mapping[str, ExtendedPropertyValue]]` for the properties.

src/handwritten/types.py
mikeprosserni: I've reverted types.py. These are no longer public.

I'm now using `from nitypes.waveform.typing import ExtendedPropertyValue` instead of `WfmAttrValue`.

`SetWfmAttrCallback` has been moved into _library_interpreter.py.

src/codegen/templates/constants.mako
mikeprosserni: constants.mako has been reverted. 

`WfmAttrType` is now in src/codegen/templates/_library_interpreter.py.mako

pyproject.toml
mikeprosserni: Should I create a work item or something to track that?

generated/nidaqmx/constants.py
mikeprosserni: WfmAttrType has been moved to _library_interpreter

generated/nidaqmx/types.py
mikeprosserni: SetWfmAttrCallback has been moved to _library_interpreter

WfmAttrValue has been deleted. Now we use from nitypes.waveform.typing import ExtendedPropertyValue instead.

src/handwritten/stream_readers.py
mikeprosserni: changed all three to initialize with zeroes

pyproject.toml
mikeprosserni: added implementation markers (matching https://github.com/ni/nitypes-python/blob/main/pyproject.toml)

pyrightconfig.json
mikeprosserni: moved these settings to pyproject.toml, and deleted pyrightconfig.json

generated/nidaqmx/stream_readers.py
mikeprosserni: done

image

links to: https://nitypes.readthedocs.io/en/latest/autoapi/nitypes/waveform/AnalogWaveform.html#AnalogWaveform

examples/analog_in/voltage_acq_int_clk_wfm.py
mikeprosserni: I'll work on that bug next.

CHANGELOG.md
bkeryan, line 1: @mikeprosserni @zhindes Perhaps this new functionality should be gated by a feature toggle until it's ready for consumption?

See https://github.com/ni/measurement-plugin-python/blob/main/packages/service/ni_measurement_plugin_sdk_service/_featuretoggles.py for a feature toggle implementation you can copy.

tests/component/test_stream_readers_ai.py
```diff
@@ -110,6 +136,25 @@ def test___analog_single_channel_reader___read_many_sample_with_wrong_dtype___ra
     assert "float64" in exc_info.value.args[0]
 
 
+@pytest.mark.grpc_skip(reason="internal_read_analog_waveform_ex not implemented in GRPC")
```
bkeryan: I think this function was added in NI-DAQmx 24.5 or 24.8. 
- Should the tests be marked accordingly or do we always run a newer version than this?
- I guess LibImporter already handles the error case.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6338,6 +6368,126 @@ def get_extended_error_info(self):
...
+                        CSetWfmAttrCallbackPtr,
+                        ctypes.c_void_p,
+                        wrapped_ndpointer(dtype=numpy.float64, flags=("C", "W")),
+                        ctypes.c_uint,
+                        ctypes.POINTER(ctypes.c_int),
+                        ctypes.POINTER(c_bool32),
+                    ]
+
+        t0_array = numpy.full(channel_count, 0, dtype=numpy.int64)
+        dt_array = numpy.full(channel_count, 0, dtype=numpy.int64)
```
bkeryan: https://numpy.org/doc/stable/reference/generated/numpy.zeros.html

generated/nidaqmx/stream_readers.py
```diff
@@ -238,6 +241,73 @@ def read_one_sample(self, timeout=10):
...
+            number_of_samples_per_channel,
+            timeout,
+            FillMode.GROUP_BY_CHANNEL.value,
+            raw_data,
+        )
+
+        waveform = AnalogWaveform(
+            dtype=numpy.double, 
+            raw_data=raw_data, 
+            extended_properties=properties[0])
```
bkeryan: This makes an unnecessary copy. To avoid this, construct an ExtendedPropertyDictionary instead of a dict and set copy_extended_properties=False when constructing the waveform, or use a pre-constructed waveform's extended properties.

generated/nidaqmx/stream_readers.py
```diff
@@ -238,6 +241,73 @@ def read_one_sample(self, timeout=10):
         """
         return self._interpreter.read_analog_scalar_f64(self._handle, timeout)
 
+    def read_waveform(
+        self, number_of_samples_per_channel=READ_ALL_AVAILABLE, timeout=10
+    ) -> AnalogWaveform[numpy.double]:
```
bkeryan: [deferrable] This always constructs a new waveform. Cifra's planned shared memory workflow will require reading into an existing waveform.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3600,6 +3602,20 @@ def set_runtime_environment(
     def internal_get_last_created_chan(self):
         raise NotImplementedError
 
+    def internal_read_analog_waveform_ex(
```
bkeryan: [deferrable] Thinking ahead to gRPC support: if y'all use ni.protobuf.types.DoubleAnalogWaveform to nidaqmx.proto, then you can push all this waveform stuff into the interpreter.

pyproject.toml
bkeryan: FYI, this overlaps with https://github.com/ni/nidaqmx-python/issues/626

Also, we're currently not claiming PyPy support for https://github.com/ni/ni-apis-python/ because there are no binary wheels of grpcio for PyPy, AFAIK. But it should be possible to test nidaqmx w/library-only.

tests/component/test_stream_readers_ai.py
mikeprosserni: Since all the tests are passing in the PR Checks, I'm guessing we always run a newer version.

generated/nidaqmx/stream_readers.py
mikeprosserni: I've gone ahead and updated `read_waveform()` to optionally take in an existing waveform to read the data into. And I added a test for this: `test___analog_single_channel_reader___read_waveform_in_place___populates_valid_waveform`

generated/nidaqmx/stream_readers.py
mikeprosserni: It now uses the pre-constructed waveform's extended_properties.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: It now uses numpy.zeros in all three places.

src/handwritten/stream_readers.py
```diff
@@ -238,6 +238,79 @@ def read_one_sample(self, timeout=10):
...
+                timeout is 10 seconds. If you set timeout to
+                nidaqmx.constants.WAIT_INFINITELY, the method waits
+                indefinitely. If you set timeout to 0, the method tries
+                once to read the requested samples and returns an error
+                if it is unable to.
+            waveform (Optional[AnalogWaveform]): Specifies an existing
+                AnalogWaveform object to use for reading samples. If
+                provided, the raw_data array of this waveform will be
+                used to store the samples. If not provided, a new
+                AnalogWaveform will be created.
```
bkeryan: Use the word "into"

https://github.com/ni/nidaqmx-python/blob/d7ab6111fb4dd20502a7b64e322f9f089068b0ba/src/codegen/templates/task/_in_stream.py.mako#L280

src/handwritten/stream_readers.py
```diff
@@ -238,6 +238,79 @@ def read_one_sample(self, timeout=10):
...
+        if waveform is None:
+            waveform = AnalogWaveform(raw_data=numpy.zeros(number_of_samples_per_channel, dtype=numpy.float64))
+
+        timestamps, sample_intervals = self._interpreter.internal_read_analog_waveform_ex(
+            self._handle,
+            1, # single channel
+            number_of_samples_per_channel,
+            timeout,
+            FillMode.GROUP_BY_CHANNEL.value,
+            waveform.raw_data,
```
bkeryan: What happens if you read into a waveform that is too small? Too large?

See MSDaqDotNET for ideas.

generated/nidaqmx/_grpc_interpreter.py
mikeprosserni: I like that plan. I've gone ahead and moved all the waveform stuff into the interpreter. The internal_read_analog_waveform_ex interface is a lot cleaner now.

src/handwritten/stream_readers.py
mikeprosserni: Updated to "Reads one or more floating-point samples from a single analog input channel **into** a waveform."

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6416,30 +6408,39 @@ def set_wfm_attr_callback(
...
+            sample_interval=self._get_wfm_timedelta(dt_array[0]),
+        )
+
+    def _get_wfm_datetime(self, t0_value: numpy.int64) -> datetime.datetime:
+        _T0_EPOCH = datetime.datetime(1, 1, 1, tzinfo=datetime.timezone.utc)        
+        return _T0_EPOCH + self._get_wfm_timedelta(t0_value)
 
-        return timestamps, sample_intervals
+    def _get_wfm_timedelta(self, value: numpy.int64) -> datetime.timedelta:
+        _INT64_WFM_SEC_PER_TICK = 100e-9
```
bkeryan: Why construct these every time?

src/handwritten/stream_readers.py
mikeprosserni: Also updated to "Specifies an existing AnalogWaveform object to use for reading samples **into**."

generated/nidaqmx/_library_interpreter.py
mikeprosserni: I've changed them back to module-level constants.

src/handwritten/stream_readers.py
mikeprosserni: If you read into a waveform that is too small, you get a DaqReadError, like this:

```
  File "c:\dev\fireserp\nidaqmx-python\examples\analog_in\voltage_acq_int_clk_wfm.py", line 17, in <module>
    reader.read_waveform(READ_ALL_AVAILABLE, waveform=waveform)
  File "C:\dev\fireserp\nidaqmx-python\generated\nidaqmx\stream_readers.py", line 306, in read_waveform
    self._interpreter.internal_read_analog_waveform_ex(
  File "C:\dev\fireserp\nidaqmx-python\generated\nidaqmx\_library_interpreter.py", line 6431, in internal_read_analog_waveform_ex
    self.check_for_error(error_code, samps_per_chan_read=samps_per_chan_read.value)
  File "C:\dev\fireserp\nidaqmx-python\generated\nidaqmx\_library_interpreter.py", line 6626, in check_for_error
    raise DaqReadError(extended_error_info, error_code, samps_per_chan_read)
nidaqmx.errors.DaqReadError: Buffer is too small to fit read data.

Buffer Length: 49
Required Buffer Size in Samples: 50
```

If you read into a waveform that is too big, you get extra zeroes in the buffer:
```
Acquired data: [-0.15140415  0.51282051  1.22588523  1.73870574  2.42735043  2.74969475
  3.37973138  3.67277167  3.96581197  4.27838828  4.75213675  4.64468864
  4.68864469  4.84493284  4.78632479  4.50793651  4.36630037  3.95604396
  3.81440781  3.32112332  2.95970696  2.46642247  1.79242979  1.19169719
  0.71306471  0.08302808 -0.45909646 -1.17216117 -1.85592186 -2.23687424
 -2.88644689 -3.26739927 -3.88278388 -4.16117216 -4.36141636 -4.50793651
 -4.76678877 -4.76678877 -4.84981685 -4.68864469 -4.58119658 -4.31746032
 -4.07326007 -3.83882784 -3.34065934 -2.84737485 -2.26129426 -1.88522589
 -1.21123321 -0.55189255  0.          0.          0.          0.
  0.          0.          0.          0.          0.          0.
  0.          0.          0.          0.          0.          0.
  0.          0.          0.          0.          0.          0.
  0.          0.          0.          0.          0.          0.
  0.          0.          0.          0.          0.          0.
  0.          0.          0.          0.          0.          0.
  0.          0.          0.          0.          0.          0.
  0.          0.          0.          0.        ]
```

From what I can tell by looking at the source code, this is the same way MSDaqDotNET handles things.

https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/nidaqmxSharp/tools/generate_classes.py&_a=contents&version=GBmain

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -156,6 +187,127 @@ class LibraryInterpreter(BaseInterpreter):
...
+            callback_data: object,
+        ) -> int:
+            waveform.extended_properties[attribute_name] = value
+            return 0
+
+        error_code = cfunc(
+            task_handle,
+            num_samps_per_chan,
+            timeout,
+            FillMode.GROUP_BY_CHANNEL.value,
```
bkeryan: fill mode should be a parameter.

bkeryan (changes requested):

generated/nidaqmx/_base_interpreter.py
```diff
@@ -1845,4 +1846,14 @@ def write_to_teds_from_file(
...
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def internal_read_analog_waveform_ex(
+        self,
+        task_handle: object,
+        num_samps_per_chan: int,
+        timeout: float,
+        waveform: AnalogWaveform[numpy.float64]
+    ) -> None:
```
bkeryan: This does not match DAQmxInternalReadAnalogWaveformEx's capabilities. This function can read waveform samples and attributes for multiple channels. Its limitation is that it uses a flat sample buffer rather than a separate sample buffer for each channel.

src/handwritten/stream_readers.py
bkeryan: See https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/MSDaqDotNET/source/ninetai2005/API/Infrastructure/AnalogSingleChannelReaderImpl.cpp&version=GC7e7645fe2fcc6793d944b2e3093cb3b875b12abe&line=339&lineEnd=340&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

Please do not post the code here on GitHub.

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -156,6 +187,127 @@ class LibraryInterpreter(BaseInterpreter):
...
+            t0_array,
+            dt_array,
+            t0_array.size,
+            self._get_wfm_attr_callback_ptr(set_wfm_attr_callback),
+            None,
+            waveform.raw_data,
+            waveform.raw_data.size,
+            ctypes.byref(samps_per_chan_read),
```
bkeryan, line 283: Looking at MSDaqDotNET reminds me of another very important case; short/partial reads.

Reading N samples may return fewer than N samples. When this happens, the read function should resize the waveform to indicate how many samples were actually read, not how many samples it tried to read. Note that this does not necessarily involve reallocating the waveform's data array or changing its capacity; that's why start index and size are separate from capacity. You may need to extend the nitypes waveforms to support changing the size without changing the data.

A short read may also be accompanied by an error. If this happens, the exception indicates the number of samples read. If the read function is returning a new waveform, the data is lost, but if the read function is reading into an existing waveform, the data is available in the waveform.

bkeryan (changes requested):

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: But shouldn't it always be GROUP_BY_CHANNEL? Isn't a `waveform.raw_data` always a single channel?

generated/nidaqmx/_base_interpreter.py
mikeprosserni: Right, but that flat sample buffer prevents this method from reading into multiple existing waveforms, right? That's why I changed this to be specifically single-channel when I changed it to take in a preexisting waveform. My expectation was that AnalogMultiChannelReader.read_waveform() would use DAQmxInternalReadAnalogWaveformPerChan instead.

Alternately, I could have internal_read_analog_waveform_ex() only return new waveforms, and have a separate internal_read_into_analog_waveform() or something that takes in an existing waveform.

Or does all this really mean that internal_read_analog_waveform_ex() should not use the waveform data type after all?

mikeprosserni: Or, as another alternative, we could just use internal_read_analog_waveform_per_chan() for both AnalogSingleChannelReader and AnalogMultiChannelReader, and not have a internal_read_analog_waveform_ex() at all.

src/codegen/templates/_library_interpreter.py.mako
bkeryan: > But shouldn't it always be GROUP_BY_CHANNEL?

For your use case; yes.
In general: no. NumPy supports both layouts: https://numpy.org/doc/stable/reference/generated/numpy.asfortranarray.html

> Isn't a waveform.raw_data always a single channel?

Yes but DAQmxInternalReadAnalogWaveformEx can read multiple channels. It's the same as DAQmxReadAnalogF64 with additional parameters for t0, dt, and attributes.

generated/nidaqmx/_base_interpreter.py
bkeryan: > Right, but that flat sample buffer prevents this method from reading into multiple existing waveforms, right? 

No. It means that using this method to read into multiple existing waveforms requires copying the data. That's how we implemented waveform reads before I added the per-chan version.

> That's why I changed this to be specifically single-channel when I changed it to take in a preexisting waveform. My expectation was that AnalogMultiChannelReader.read_waveform() would use DAQmxInternalReadAnalogWaveformPerChan instead.
> 
> Alternately, I could have internal_read_analog_waveform_ex() only return new waveforms, and have a separate internal_read_into_analog_waveform() or something that takes in an existing waveform.
> 
> Or does all this really mean that internal_read_analog_waveform_ex() should not use the waveform data type after all?

In light of this discussion, I now think it could make sense for internal_read_analog_waveform_ex() to be a lower-level method that only exists in the library interpreter. The gRPC interpreter can natively implement read_waveform() and the library interpreter can implement read_waveform in terms of internal_read_analog_waveform_ex().

You will have to be careful about the method names when adding grpc-device support.

> Or, as another alternative, we could just use internal_read_analog_waveform_per_chan() for both AnalogSingleChannelReader and AnalogMultiChannelReader, and not have a internal_read_analog_waveform_ex() at all.

Using a flat buffer for single-channel reads is a performance optimization, at least in .NET. I don't know how much it will matter in Python.

bkeryan: > > Right, but that flat sample buffer prevents this method from reading into multiple existing waveforms, right?
> 
> No. It means that using this method to read into multiple existing waveforms requires copying the data. That's how we implemented waveform reads before I added the per-chan version.

BTW, this is also how it will work for digital: read/write with a flat buffer, and copy the data to/from the user's waveforms. We didn't add a per-chan version of the digital read/write methods for .NET-specific reasons. Python would benefit from per-chan digital read/write methods, but this is probably outside the scope of your feature.

mikeprosserni: I've refactored along those lines: 

BaseInterpreter, LibraryInterpreter, and GrpcInterpreter now have `read_analog_waveform()`, which uses the AnalogWaveform data type. 

LibraryInterpreter now has a private `_internal_read_analog_waveform_ex()` which matches DAQmxInternalReadAnalogWaveformEx's capabilities. (and doesn't use AnalogWaveform)

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: fill_mode is now a parameter again on _internal_read_analog_waveform_ex()

generated/nidaqmx/_base_interpreter.py
mikeprosserni: > We didn't add a per-chan version of the digital read/write methods

Without a per-chan digital read method, is it possible to write a `DigitalMultiChannelReader.read_waveform()` that supports read-into without copying any arrays? Should it support read-into even if copies are required?

bkeryan: Yes, it should support read-into for the shared memory workflow, even if copies are required.

mikeprosserni: I've updated the HLD to say that DigitalMultiChannelReader.read_waveform() will support read-into, even if copies are required.

src/handwritten/stream_readers.py
mikeprosserni: Based on that, what we should do here is this:

1. If you read into a waveform that is too small, you get a DaqReadError by default.
2. If you read into a waveform that is too small, but specify something like `ReallocationPolicy.TO_GROW`, we increase the size of the waveform for you.
3. If you read into a waveform that is too big, we decrease the size of the waveform for you.

I have implemented the first one by raising a DaqError in read_waveform(). The other two cannot be implemented until I add support for changing Waveform.sample_count, which I've added a [Task](https://dev.azure.com/ni/DevCentral/_workitems/edit/3228924) for.

I've added TODOs so that those can be done in a subsequent PR.

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: I've added a [Task](https://dev.azure.com/ni/DevCentral/_workitems/edit/3228924) for adding support for changing the size of a waveform. Once that is done, we can handle short reads here by setting the waveform size to the number of samples read before throwing the read exception. 

I've added TODOs so that this can be done in a subsequent PR.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -1,24 +1,55 @@
...
+    if sys.version_info >= (3, 10):
+        from typing import TypeAlias
+    else:
+        from typing_extensions import TypeAlias
 
 _logger = logging.getLogger(__name__)
 _was_runtime_environment_set = None
 
+_INT64_WFM_SEC_PER_TICK = 100e-9
+_T0_EPOCH = datetime.datetime(1, 1, 1, tzinfo=datetime.timezone.utc)
```
bkeryan: .NET timestamps have 100 ns resolution. datetime has 1 us resolution. Use hightime to avoid losing precision.

src/codegen/templates/_library_interpreter.py.mako
bkeryan: Ok, thanks

tests/component/test_stream_readers_ai.py
```diff
@@ -110,6 +137,57 @@ def test___analog_single_channel_reader___read_many_sample_with_wrong_dtype___ra
...
+
+@pytest.mark.grpc_skip(reason="read_analog_waveform not implemented in GRPC")
+def test___analog_single_channel_reader___read_waveform_in_place___populates_valid_waveform(
+    ai_single_channel_task_with_timing: nidaqmx.Task,
+) -> None:
+    reader = AnalogSingleChannelReader(ai_single_channel_task_with_timing.in_stream)
+    samples_to_read = 10
+
+    waveform = AnalogWaveform(raw_data=numpy.zeros(samples_to_read, dtype=numpy.float64))
+    reader.read_waveform(number_of_samples_per_channel=samples_to_read, waveform=waveform)
```
bkeryan, line 204: Test reading into a previously used waveform. It should overwrite the data/timing/attributes.

generated/nidaqmx/stream_readers.py
```diff
@@ -238,6 +238,87 @@ def read_one_sample(self, timeout=10):
...
+            waveform = AnalogWaveform(raw_data=numpy.zeros(number_of_samples_per_channel, dtype=numpy.float64))
+        elif number_of_samples_per_channel > waveform.sample_count:
+            # TODO: AB#3228924 - if allowed by the caller, increase the sample count of the waveform
+            raise DaqError(
+                f'The provided waveform does not have enough space ({waveform.sample_count}) to hold '
+                f'the requested number of samples ({number_of_samples_per_channel}). Please provide a larger '
+                'waveform or adjust the number of samples requested.',
+                DAQmxErrors.READ_BUFFER_TOO_SMALL, task_name=self._task.name)
+
+        self._interpreter.read_analog_waveform(
```
bkeryan, line 313: Does the .NET API clear the incoming waveform's attributes? (Please try it, don't just look at the code.)

tests/component/test_stream_readers_ai.py
```diff
@@ -110,6 +137,57 @@ def test___analog_single_channel_reader___read_many_sample_with_wrong_dtype___ra
...
+
+@pytest.mark.grpc_skip(reason="read_analog_waveform not implemented in GRPC")
+def test___analog_single_channel_reader___read_into_undersized_waveform___throws_exception(
+    ai_single_channel_task_with_timing: nidaqmx.Task,
+) -> None:
+    reader = AnalogSingleChannelReader(ai_single_channel_task_with_timing.in_stream)
+    samples_to_read = 10
+
+    waveform = AnalogWaveform(raw_data=numpy.zeros(samples_to_read - 1, dtype=numpy.float64))
+    with pytest.raises(DaqError, match="The provided waveform does not have enough space"):
```
bkeryan: FYI, most of our code uses a separate `assert exc.value.args[0].startswith("...")` but I looked at the output for a failed match and it seems reasonable to debug, so it's ok.

tests/component/test_stream_readers_ai.py
```diff
@@ -110,6 +137,57 @@ def test___analog_single_channel_reader___read_many_sample_with_wrong_dtype___ra
...
 
 
+@pytest.mark.grpc_skip(reason="read_analog_waveform not implemented in GRPC")
+def test___analog_single_channel_reader___read_waveform___returns_valid_waveform(
+    ai_single_channel_task_with_timing: nidaqmx.Task,
+) -> None:
+    reader = AnalogSingleChannelReader(ai_single_channel_task_with_timing.in_stream)
+    samples_to_read = 10
+
+    waveform = reader.read_waveform(number_of_samples_per_channel=samples_to_read)
```
bkeryan, line 164: Test read_waveform() with no arguments. Because this is a finite acquisition, it should read the number of samples passed to cfg_samp_clk_timing.

tests/component/test_stream_readers_ai.py
```diff
@@ -110,6 +137,57 @@ def test___analog_single_channel_reader___read_many_sample_with_wrong_dtype___ra
...
+
+    assert isinstance(waveform, AnalogWaveform)
+    expected = _get_voltage_offset_for_chan(0)
+    assert waveform.scaled_data == pytest.approx(expected, abs=VOLTAGE_EPSILON)
+    assert isinstance(waveform.timing.timestamp, datetime.datetime)
+    assert _is_timestamp_close_to_now(waveform.timing.timestamp)
+    assert waveform.timing.sample_interval == datetime.timedelta(seconds=1 / 1000)
```
bkeryan: Test read_waveform() with a sample rate over 1 MS/s and make sure it returns the correct sample interval. You can use the simulated PXIe-4480 named "chargeTester", which supports 20 MS/s.

```
>>> dt.timedelta(seconds=1/20e6)
datetime.timedelta(0)
```

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6338,6 +6369,152 @@ def get_extended_error_info(self):
...
+
+        def set_wfm_attr_callback(
+            channel_index: int,
+            attribute_name: str,
+            attribute_type: WfmAttrType,
+            value: ExtendedPropertyValue,
+            callback_data: object,
+        ) -> int:
+            properties[channel_index][attribute_name] = value
+            return 0
```
bkeryan, line 6450: [separate PR] In LV and .NET, there is a public API property to allow users to disable timing and/or attributes. You should add it to Python.

https://www.ni.com/docs/en-US/bundle/ni-daqmx-labview-api-ref/page/resource/objmgr/daqmx-rc/daqmx-read/attr1225.html

https://www.ni.com/docs/en-US/bundle/ni-daqmx-net-framework-40-api-ref/page/ninetdaqmxfx40ref/html/p_nationalinstruments_daqmx_daqstream_waveformattributemode.htm

src/handwritten/stream_readers.py
bkeryan: ok

generated/nidaqmx/_library_interpreter.py
mikeprosserni: Updated to use hightime.datetime and hightime.timedelta

tests/component/test_stream_readers_ai.py
```diff
@@ -80,6 +80,22 @@ def ai_single_channel_task_with_timing(
...
+    task: nidaqmx.Task, sim_charge_device: nidaqmx.system.Device
+) -> nidaqmx.Task:
+    offset = _get_voltage_offset_for_chan(0)
+    task.ai_channels.add_ai_voltage_chan(
+        sim_charge_device.ai_physical_chans[0].name,
+        min_val=offset,
+        max_val=offset + VOLTAGE_EPSILON,
+    )
+    task.timing.cfg_samp_clk_timing(
+        rate=2_000_000, sample_mode=AcquisitionType.FINITE, samps_per_chan=50
```
bkeryan:
```suggestion
        rate=20_000_000, sample_mode=AcquisitionType.FINITE, samps_per_chan=50
```
or
```suggestion
        rate=20e6, sample_mode=AcquisitionType.FINITE, samps_per_chan=50
```

tests/component/test_stream_readers_ai.py
```diff
@@ -184,9 +221,32 @@ def test___analog_single_channel_reader___read_into_undersized_waveform___throws
...
+    samples_to_read = 50
+
+    waveform = reader.read_waveform(number_of_samples_per_channel=samples_to_read)
+
+    assert isinstance(waveform, AnalogWaveform)
+    expected = _get_voltage_offset_for_chan(0)
+    assert waveform.scaled_data == pytest.approx(expected, abs=VOLTAGE_EPSILON)
+    assert isinstance(waveform.timing.timestamp, ht_datetime)
+    assert _is_timestamp_close_to_now(waveform.timing.timestamp)
+    assert waveform.timing.sample_interval == ht_timedelta(seconds=1 / 2_000_000)
```
bkeryan:
```suggestion
    assert waveform.timing.sample_interval == ht_timedelta(seconds=1 / 20_000_000)
```
or
```suggestion
    assert waveform.timing.sample_interval == ht_timedelta(seconds=1 / 20e6)
```

tests/component/test_stream_readers_ai.py
bkeryan: 20, not 2

mikeprosserni: DAQmxInternalReadAnalogWaveformEx only supports sample rates up to 10 MS/s. At that point, the dt_array contains an `int64(1)`. 

At 20 MS/s, the dt_array just contains an `int64(0)`. Is this expected?

bkeryan: I see. I filed [AB#3229757](https://dev.azure.com/ni/DevCentral/_workitems/edit/3229757). Please test 10 MS/s for now.

tests/component/test_stream_readers_ai.py
mikeprosserni: added `test___analog_single_channel_reader___reuse_waveform_in_place___overwrites_data_timing_and_attributes`

tests/component/test_stream_readers_ai.py
mikeprosserni: Actually, I think I like that style better. I changed this test to use that style, so I can also assert the error code as well as the message.

tests/component/test_stream_readers_ai.py
mikeprosserni: added `test___analog_single_channel_reader___read_waveform_no_args___returns_valid_waveform`

tests/component/test_stream_readers_ai.py
mikeprosserni: now testing with 10_000_000

tests/component/test_stream_readers_ai.py
mikeprosserni: changed to 10_000_000

tests/component/test_stream_readers_ai.py
mikeprosserni: changed to 10_000_000

generated/nidaqmx/_library_interpreter.py
mikeprosserni: added a task for this: https://dev.azure.com/ni/DevCentral/_workitems/edit/3229778

tests/component/test_stream_readers_ai.py
```diff
@@ -213,6 +214,39 @@ def test___analog_single_channel_reader___read_waveform_in_place___populates_val
...
+
+    reader0 = _make_single_channel_reader(chan_index=0, offset=0, rate=1000.0)
+    reader1 = _make_single_channel_reader(chan_index=1, offset=1, rate=2000.0)
+    waveform = AnalogWaveform(raw_data=numpy.zeros(10, dtype=numpy.float64))
+
+    reader0.read_waveform(number_of_samples_per_channel=10, waveform=waveform)
+    timestamp1 = waveform.timing.timestamp
+    assert waveform.scaled_data == pytest.approx(0, abs=VOLTAGE_EPSILON)
+    assert waveform.timing.sample_interval == ht_timedelta(seconds=1 / 1000)
+    assert waveform.channel_name == "nidaqmxMultithreadingTester1/ai0"
```
bkeryan: Use sim_6363_device instead of hardcoding the device name.

bkeryan (approved):
I'm ok with resolving the remaining open issues in separate PRs
- Feature toggles
- WaveformAttributeMode
- etc.

pyproject.toml
bkeryan: Sure

mikeprosserni: added task: https://dev.azure.com/ni/DevCentral/_workitems/edit/3229809

CHANGELOG.md
mikeprosserni: added task: https://dev.azure.com/ni/DevCentral/_workitems/edit/3229811

generated/nidaqmx/stream_readers.py
mikeprosserni: added task: https://dev.azure.com/ni/DevCentral/_workitems/edit/3229822

tests/component/test_stream_readers_ai.py
mikeprosserni: it's now `assert waveform.channel_name == f"{sim_6363_device.name}/ai0"` and `assert waveform.channel_name == f"{sim_6363_device.name}/ai1"`

zhindes (approved):

PR #800, opened 2025-08-14
Add in_stream.waveform_attribute_mode

mikeprosserni:

~~[ ] I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~

- Added `WaveformAttributeMode`, matching the one in [.NET](https://www.ni.com/docs/en-US/bundle/ni-daqmx-net-framework-40-api-ref/page/ninetdaqmxfx40ref/html/t_nationalinstruments_daqmx_waveformattributemodes.htm)
- Added a `waveform_attribute_mode` property on `InStream`
- Use the `WaveformAttributeMode` in `LibraryInterpreter` to control the setting of timing information and extended properties

In LV and .NET, there is a public "WaveformAttributeMode" property to allow users to disable timing and/or attributes. We should have the same in python.

```
test___analog_single_channel_reader_with_timing_flag___read_waveform___only_includes_timing_data
test___analog_single_channel_reader_with_extended_properties_flag___read_waveform___only_includes_extended_properties
test___analog_single_channel_reader_with_both_flags___read_waveform___includes_both_timing_and_extended_properties
test___analog_single_channel_reader_with_none_flag___read_waveform___minimal_waveform_data
```


generated/nidaqmx/_library_interpreter.py
```diff
@@ -6446,7 +6450,8 @@ def set_wfm_attr_callback(
             value: ExtendedPropertyValue,
```
bkeryan: Don't allocate t0_array and dt_array when timing is disabled.

That's what the `0 if t0_array is None else t0_array.size` is for.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6386,14 +6387,16 @@ def read_analog_waveform(
             timeout,
             FillMode.GROUP_BY_CHANNEL.value,
             waveform.raw_data,
-            [waveform.extended_properties]
+            [waveform.extended_properties],
```
bkeryan: Don't allocate a list when extended properties are disabled. Pass None and update the type hint to allow None.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6446,7 +6450,8 @@ def set_wfm_attr_callback(
             value: ExtendedPropertyValue,
             callback_data: object,
         ) -> int:
-            properties[channel_index][attribute_name] = value
+            if WaveformAttributeMode.EXTENDED_PROPERTIES in waveform_attribute_mode:
```
bkeryan: Instead of checking the flag for every property, pass a null callback (`CSetWfmAttrCallbackPtr()`) when extended properties are disabled.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: I've refactored _internal_read_analog_waveform_ex() so that it can take in None for timestamps and sample_intervals, and it doesn't allocation t0_array or dt_array in that case.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: It now passes in None when properties are disabled.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: It now uses a null callback when properties is None.

bkeryan (approved):

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6457,18 +6484,21 @@ def set_wfm_attr_callback(
...
         )
 
-        timestamps = [_T0_EPOCH + ht_timedelta(seconds=t0 * _INT64_WFM_SEC_PER_TICK) for t0 in t0_array]
-        sample_intervals = [ht_timedelta(seconds=dt * _INT64_WFM_SEC_PER_TICK) for dt in dt_array]
+        if timestamps is not None and t0_array is not None:
+            timestamps[:] = [_T0_EPOCH + ht_timedelta(seconds=t0 * _INT64_WFM_SEC_PER_TICK) for t0 in t0_array]
 
-        return error_code, samps_per_chan_read.value, timestamps, sample_intervals
+        if sample_intervals is not None and dt_array is not None:
+            sample_intervals[:] = [ht_timedelta(seconds=dt * _INT64_WFM_SEC_PER_TICK) for dt in dt_array]
```
bkeryan: On further thought: this is doing unnecessary list allocations and copies. It would be better to allocate the numpy arrays in the caller and convert to hightime in the for-loop that allocates the timing objects.

bkeryan (changes requested):

zhindes (approved):

generated/nidaqmx/_library_interpreter.py
mikeprosserni: I moved the numpy array allocation to the callers (read_analog_waveform and read_analog_waveforms), and the hightime conversion is now done like this:

```
    def _set_waveform_timings(
        self, 
        waveforms: Sequence[AnalogWaveform[numpy.float64]], 
        t0_array: numpy.typing.NDArray[numpy.int64], 
        dt_array: numpy.typing.NDArray[numpy.int64]
    ) -> None:
        for i, waveform in enumerate(waveforms):
            waveform.timing = Timing(
                sample_interval_mode=SampleIntervalMode.REGULAR,
                timestamp=_T0_EPOCH + ht_timedelta(seconds=t0_array[i] * _INT64_WFM_SEC_PER_TICK),
                sample_interval=ht_timedelta(seconds=dt_array[i] * _INT64_WFM_SEC_PER_TICK),
            )
```

PR #807, opened 2025-08-18
Add task.read_waveform() for analog inputs

mikeprosserni:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).

- [ ] ~~I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~

- [x] I've added tests applicable for this pull request

### What does this Pull Request accomplish?

This adds `Task.read_waveform()`, which is similar to `Task.read()` in that the return type is automatically determined based on the task configuration. `Task.read_waveform()` will return a single waveform if there's a single channel, or a list of waveforms if there are multiple channels. Currently, only analog channels are supported. Digital channel support will be added in a future user story.

### Why should this Pull Request be merged?

Implements [AB#3233521](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3233521)

### What testing has been done?

```
test___analog_single_channel___read_waveform___returns_valid_waveform
test___analog_single_channel___read_waveform_one_sample___returns_waveform_with_one_sample
test___analog_single_channel___read_waveform_many_sample___returns_waveform_with_many_samples
test___analog_multi_channel___read_waveform___returns_valid_waveforms
test___analog_multi_channel___read_waveform_one_sample___returns_waveforms_with_single_sample
test___analog_multi_channel___read_waveform_many_samples___returns_waveforms_with_many_samples
```

Issue #528, opened 2024-03-05
Task.read interprets data incorrectly for short reads


bkeryan:
When `DAQmxReadAnalogF64` with `DAQmx_Val_GroupByChannel` returns fewer samples than requested, it squashes the valid samples together at the beginning of the buffer. The way that `nidaqmx.Task.read` handles short reads doesn't take this into account, so it may return samples that have been overwritten.

Test case (from `tests/component/test_task_read_ai.py`, under development):
```python
def test___analog_multi_channel_finite___read_too_many_sample___returns_valid_2d_channels_samples_truncated(
    ai_multi_channel_task: nidaqmx.Task,
) -> None:
    samples_to_acquire = 5
    ai_multi_channel_task.timing.cfg_samp_clk_timing(rate=1000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=samples_to_acquire)
    num_channels = ai_multi_channel_task.number_of_channels
    samples_to_read = 10

    data = ai_multi_channel_task.read(samples_to_read)

    expected = [
        [_get_voltage_offset_for_chan(chan_index) for _ in range(samples_to_acquire)]
        for chan_index in range(num_channels)
    ]
    _assert_equal_2d(data, expected, abs=VOLTAGE_EPSILON)
```

Result (lib and grpc fail the same way, because the underlying interpreters have consistent behavior here):
```
_ test___analog_multi_channel_finite___read_too_many_sample___returns_valid_2d_channels_samples_truncated[library_init_kwargs] _

ai_multi_channel_task = Task(name=_unnamedTask<0>)

    def test___analog_multi_channel_finite___read_too_many_sample___returns_valid_2d_channels_samples_truncated(
        ai_multi_channel_task: nidaqmx.Task,
    ) -> None:
        samples_to_acquire = 5
        ai_multi_channel_task.timing.cfg_samp_clk_timing(rate=1000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=samples_to_acquire)
        num_channels = ai_multi_channel_task.number_of_channels
        samples_to_read = 10

        data = ai_multi_channel_task.read(samples_to_read)

        expected = [
            [_get_voltage_offset_for_chan(chan_index) for _ in range(samples_to_acquire)]
            for chan_index in range(num_channels)
        ]
>       _assert_equal_2d(data, expected, abs=VOLTAGE_EPSILON)

tests\component\test_task_read_ai.py:144:
[...]

Issue #421, opened 2023-06-22
`stream_readers` array shape validation is incompatible with `READ_ALL_AVAILABLE`


bkeryan:
If you use `stream_readers` to read into a preallocated NumPy array with `number_of_samples_per_channel=READ_ALL_AVAILABLE`, the stream reader validates the NumPy array shape against the available samples per channel, which is not a predictable number.

## Steps to reproduce
Install an X Series or M Series as Dev1 and run this code:
```
import pprint
import time

import nidaqmx
import numpy
from nidaqmx.constants import AcquisitionType
from nidaqmx.stream_readers import AnalogMultiChannelReader

pp = pprint.PrettyPrinter(indent=4)

with nidaqmx.Task() as task:
    task.ai_channels.add_ai_voltage_chan("Dev1/ai0:1")
    task.timing.cfg_samp_clk_timing(1000.0, sample_mode=AcquisitionType.CONTINUOUS)
    task.start()
    time.sleep(100e-3)

    reader = AnalogMultiChannelReader(task.in_stream)
    data = numpy.zeros((2, 1000), dtype=numpy.double)
    reader.read_many_sample(data, timeout=0.0)

    pp.pprint(data)
```

## Expected output
Prints array data

## Actual output
The required number of samples varies from run to run but is around 100 to 110.
```
Traceback (most recent call last):
  File "D:\dev\nidaqmx-python\examples\ai_raw.py", line 19, in <module>
    reader.read_many_sample(data, timeout=0.0)
  File "D:\dev\nidaqmx-python\generated\nidaqmx\stream_readers.py", line 322, in read_many_sample
    self._verify_array(data, number_of_samples_per_channel, True, True)
  File "D:\dev\nidaqmx-python\generated\nidaqmx\stream_readers.py", line 83, in _verify_array
    raise DaqError(
nidaqmx.errors.DaqError: Read cannot be performed because the NumPy array passed into this function is not shaped correctly. You must pass in a NumPy array of the correct shape based on the number of channels in task and the number of samples per channel requested.

Shape of NumPy Array provided: (2, 1000)
Shape of NumPy Array required: (2, 105)

Task Name: _unnamedTask<0>
```

zhindes:
Hmm, good catch. We could allow for a NumPy array that is bigger than necessary since `read_many_sample` returns how much data is valid.

bkeryan:
With GROUP_BY_CHANNEL, a larger array will not have the right binary layout.

[...]

Issue #90, opened 2020-03-27
read_many_sample cannot use views


spalatofhi:
This is an improvement suggestion.

The `read_many_sample` methods of `stream_reader` classes have odd requirements for the data array. The array shape must be `(n_channels, n_samples)` and must be `C_CONTIGUOUS`. In the language of `nidaq-mx`, the data array must be 'non-interleaved'. This causes extra difficulties when using continuous acquisition mode. In this case, the exact number of samples may not be always known. As an example, this prevents the use of array views as an argument to `read_many_sample`. Let's review the root of this problem, show some example code, and introduce some possible solutions.

Numpy arrays can be either `C_CONTIGUOUS` (or row-major, C-order) or `F_CONTIGUOUS` (or column-major). This controls the layout in memory for multidimensional arrays. For 2D arrays, `C_CONTIGUOUS` arrays store the elements one row after the other. Ie: `arr[i,j]` is next to `arr[i, j+1]` in memory. See: https://docs.scipy.org/doc/numpy/glossary.html#term-row-major. This means that changing the number of rows can be done without reorganizing the entire array. Furthermore, a slice containing all rows but not all columns is not `C_CONTIGUOUS`, whereas a slice containing all columns but some rows is not. Since `nidaqmx` maps rows onto channels and samples onto columns, a slice onto some samples is not `C_CONTIGUOUS`, and cannot be used as an argument for `stream_reader.AnalogMultiChannelReader.read_many_samples`. This also means the transfer from the buffer to the numpy array is not in native order, thus not so efficient.

```python
import numpy as np

n_chan, n_samp = 4, 100
c = np.zeros((n_chan, n_samp), order="C") # default order
assert c.flags.c_contiguous
c_cols = c[:,10:15]
assert not c_cols.flags.c_contiguous
```

As an example demonstration, let's perform continuous acquisition. This example requires a temporary buffer for each call. This exact example can be achieved using finite acquisition, but the pattern can be useful for other applications:

 ```python
import numpy as np
import nidaqmx as ni
from nidaqmx.constants import VoltageUnits, AcquisitionType, READ_ALL_AVAILABLE
from nidaqmx.stream_readers import AnalogMultiChannelReader
from time import sleep

##### SETUP
n_tot = 100000
sample_rate = 200000

with ni.Task("signals") as task:
    task.ai_channels.add_ai_voltage_chan(
        "DevT/ai0:1", 
        min_val=-10, max_val=10,
    ) 
[...]

PR #809, opened 2025-08-21
Add support for reading digital waveforms

mikeprosserni:

- Add `read_digital_waveform()`, `read_digital_waveforms()`, and `_internal_read_digital_waveform()` to `LibraryInterpreter`
- Extend `nidaqmx.Task.read_waveform()` - reads a single waveform or list of waveforms using `Interpreter.read_digital_waveform()` or `Interpreter.read_digital_waveforms()`
- Add `nidaqmx.stream_readers.DigitalSingleChannelReader.read_waveform()` - reads a single waveform using `Interpreter.read_digital_waveform()`
- Add `nidaqmx.stream_readers.DigitalMultiChannelReader.read_waveform()` - reads a list of waveforms using `Interpreter.read_digital_waveforms()`

Implements [AB#3066730](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3066730)

Tests added to tests/component/test_stream_readers_di.py:
```
test___digital_single_line_reader___read_waveform_feature_disabled___raises_feature_not_supported_error
test___digital_multi_channel_multi_line_reader___read_waveforms_feature_disabled___raises_feature_not_supported_error

test___digital_single_line_reader___read_waveform___returns_valid_waveform
test___digital_single_line_reader___read_waveform_no_args___returns_valid_waveform
test___digital_single_line_reader___read_waveform_in_place___returns_valid_waveform
test___digital_single_line_reader___reuse_waveform_in_place___overwrites_data_timing_and_attributes

test___digital_single_line_reader___read_waveform_high_sample_rate___returns_correct_sample_interval
test___digital_single_line_reader___read_into_undersized_waveform___throws_exception

test___digital_single_line_reader_with_timing_flag___read_waveform___only_includes_timing_data
test___digital_single_line_reader_with_extended_properties_flag___read_waveform___only_includes_extended_properties
test___digital_single_line_reader_with_both_flags___read_waveform___includes_both_timing_and_extended_properties
test___digital_single_line_reader_with_none_flag___read_waveform___minimal_waveform_data

test___digital_single_channel_port_uint32_reader___read_waveform___returns_valid_waveform (failing due to [AB#3178052](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3178052))
test___digital_single_channel_lines_and_port___read_waveform___returns_valid_waveform

test___digital_single_channel_multi_line_reader___read_waveform___returns_valid_waveform
test___digital_single_channel_multi_line_reader___read_waveform_no_args___returns_valid_waveform
test___digital_single_channel_multi_line_reader___read_waveform_in_place___returns_valid_waveform
test___digital_single_channel_multi_line_reader___reuse_waveform_in_place___overwrites_data_timing_and_attributes

test___digital_multi_channel_multi_line_reader___read_waveforms___returns_valid_waveforms
test___digital_multi_channel_multi_line_reader___read_waveforms_no_args___returns_valid_waveforms
test___digital_multi_channel_multi_line_reader___read_waveforms_in_place___populates_valid_waveforms
test___digital_multi_channel_multi_line_reader___reuse_waveform_in_place___overwrites_data_timing_and_attributes

test___digital_multi_channel_multi_line_reader___read_with_wrong_number_of_waveforms___throws_exception
test___digital_multi_channel_multi_line_reader___read_into_undersized_waveforms___throws_exception

test___digital_multi_channel_multi_line_reader_with_timing_flag___read_waveforms___only_includes_timing_data
test___digital_multi_channel_multi_line_reader_with_extended_properties_flag___read_waveforms___only_includes_extended_properties
test___digital_multi_channel_multi_line_reader_with_both_flags___read_waveforms___includes_both_timing_and_extended_properties
test___digital_multi_channel_multi_line_reader_with_none_flag___read_waveforms___minimal_waveform_data

test___digital_multi_channel_different_lines_reader___read_waveforms___returns_valid_waveforms
test___digital_multi_channel_different_lines_reader___read_mismatched_waveforms___throws_exception
test___digital_multi_channel_lines_and_port_reader___read_waveforms___returns_valid_waveforms
```

Tests added to tests/component/test_task_read_waveform_di.py:
```
test___digital_single_channel___read_waveform___returns_valid_waveform
test___digital_single_channel___read_waveform_one_sample___returns_waveform_with_one_sample
test___digital_single_channel___read_waveform_many_sample___returns_waveform_with_many_samples
test___digital_single_channel___read_waveform_too_many_samples___returns_waveform_with_correct_number_of_samples (failing due to [AB#3228924](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3228924))
test___digital_single_channel___read_waveform_lines_and_port___returns_valid_waveform

test___digital_multi_channel___read_waveform___returns_valid_waveforms
test___digital_multi_channel___read_waveform_one_sample___returns_waveforms_with_single_sample
test___digital_multi_channel___read_waveform_many_samples___returns_waveforms_with_many_samples
test___digital_multi_channel___read_waveform_too_many_samples___returns_waveforms_with_correct_number_of_samples (failing due to [AB#3228924](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3228924))
test___digital_multi_channel___read_waveform_different_lines___returns_valid_waveforms
test___digital_multi_channel___read_waveform_lines_and_port___returns_valid_waveforms
```


generated/nidaqmx/_base_interpreter.py
bkeryan, line 1:
```
Run poetry run mypy
generated/nidaqmx/_base_interpreter.py:1895: error: X | Y syntax for unions requires Python 3.10  [syntax]
Found 1 error in 1 file (checked 184 source files)
```

Add `from __future__ import annotations`. This prevents the Python interpreter from parsing type hints at run time, and I think it will probably make mypy stop erroring here too. Note that this doesn't work for type aliases like `T: TypeAlias = int | str`; we still use `Union` for those.

mikeprosserni: Thanks, I was really scratching my head about that.

src/handwritten/stream_readers.py
```diff
@@ -2634,4 +2716,94 @@ def read_one_sample_port_uint32(self, data, timeout=10):
...
+        if waveforms is not None:
+            if len(waveforms) != number_of_channels:
+                raise DaqError(
+                    f'The number of waveforms provided ({len(waveforms)}) does not match '
+                    f'the number of channels in the task ({number_of_channels}). Please provide '
+                    'one waveform for each channel.',
+                    DAQmxErrors.MISMATCHED_INPUT_ARRAY_SIZES, task_name=self._task.name)
+
+            for i, waveform in enumerate(waveforms):
+                if number_of_samples_per_channel > waveform.sample_count:
```
jfriedri-ni: Maybe we should use `waveform.capacity` instead of `waveform.sample_count` if/when we do AB#3228924 .

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6637,6 +6637,168 @@ def _set_waveform_timings(
...
+        number_of_samples_per_channel: int,
+        number_of_signals_per_sample: int,
+        timeout: float,
+        waveform_attribute_mode: WaveformAttributeMode,
+        waveforms: Sequence[DigitalWaveform[numpy.uint8]] | None = None,
+    ) -> Sequence[DigitalWaveform[numpy.uint8]]:
+        """Read a digital waveform with timing and attributes."""
+        read_array = numpy.zeros(
+            (channel_count, number_of_samples_per_channel, number_of_signals_per_sample),
+            dtype=numpy.uint8)
```
zhindes, line 6709: Why does this allocate?

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6637,6 +6637,168 @@ def _set_waveform_timings(
...
+
+    def read_digital_waveforms(
+        self,
+        task_handle: object,
+        channel_count: int,
+        number_of_samples_per_channel: int,
+        number_of_signals_per_sample: int,
+        timeout: float,
+        waveform_attribute_mode: WaveformAttributeMode,
+        waveforms: Sequence[DigitalWaveform[numpy.uint8]] | None = None,
```
zhindes: optional parameter in the interpreter seems a bit off. I understand why `_internal_read_digital_waveform` has them since some of those are truly optional. but for this, why not make the caller always provide it and make this impl more straight-forward?

generated/nidaqmx/stream_readers.py
```diff
@@ -2165,6 +2164,89 @@ def read_one_sample_port_uint32(self, timeout=10):
         """
         return self._interpreter.read_digital_scalar_u32(self._handle, timeout)
 
+    @requires_feature(WAVEFORM_SUPPORT)
+    def read_waveform(
+        self,
+        number_of_samples_per_channel: int = READ_ALL_AVAILABLE,
+        timeout: float = 10.0,
+        waveform: DigitalWaveform[numpy.uint8] | None = None
```
zhindes: I didn't notice this on the analog PRs, but I don't like this being optional either. The point of the stream readers is that they are performance optimized. Therefore we force callers to allocate memory for us. I don't think we should give the option to do it on their behalf.

generated/nidaqmx/_library_interpreter.py
zhindes: It's also odd that this one is optional, but read a single waveform isn't. Unclear why its not symmetric.

generated/nidaqmx/stream_readers.py
zhindes: @bkeryan  - this is the root of my issues here... everything else follows

examples/digital_in/acq_dig_port_int_clk_wfm.py
```diff
@@ -0,0 +1,29 @@
+﻿"""Example for reading digital signals.
+
+This example demonstrates how to input a finite digital pattern
+using the DAQ device's internal clock.
+"""
+
+import os
+
+os.environ["NIDAQMX_ENABLE_WAVEFORM_SUPPORT"] = "1"
```
bkeryan, line 9: FYI: It's a shame that you can't specify this in the shebang line on Windows.

```
#!/usr/bin/env NIDAQMX_ENABLE_WAVEFORM_SUPPORT=1 python
```

src/codegen/templates/_base_interpreter.py.mako
```diff
@@ -79,4 +81,28 @@ class BaseInterpreter(abc.ABC):
...
+    def read_digital_waveforms(
+        self,
+        task_handle: object,
+        channel_count: int,
+        number_of_samples_per_channel: int,
+        number_of_signals_per_sample: int,
+        timeout: float,
+        waveform_attribute_mode: WaveformAttributeMode,
+        waveforms: Sequence[DigitalWaveform[numpy.uint8]] | None = None,
+    ) -> Sequence[DigitalWaveform[numpy.uint8]]:
```
bkeryan, line 119: The return value of read_digital_waveforms() is inconsistent with the other read_*_waveform(s) methods.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: The root of the problem here is that there isn't a digital equivalent of `DAQmxInternalReadAnalogWaveformPerChan`, which takes in a list of separate arrays to read separate channels into. Instead, we have to use `DAQmxInternalReadDigitalWaveform`, which can read multiple channels, but only into a single contiguous array.

So we have to create a single, contiguous read_array to send into `_internal_read_digital_waveform`. We can then either create waveforms using slices of that read_array, or copy the data from read_array into pre-existing waveforms. 

I made waveforms optional here because that allows us to create waveforms using the read_array, without an additional data copy. I thought that this might more performance optimized than the alternative of copying the data into pre-existing waveforms. (But of course, much less optimized than it would be to read directly into the arrays of the existing waveforms, as we do for analog)

But you could make the argument that this create-waveforms-from-read_array-slices option isn't significantly (or at all?) more performant that the copy-waveform-data-from-read_array-slices option, and isn't worth the complexity and confusion. In that case, I'd be happy to made the waveforms parameter required, and just always copy the data.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: DAQmxInternalReadDigitalWaveform requires a single contiguous array to read data from multiple channels into. Unfortunately, there isn't a DAQmxInternalReadDigitalWaveformPerChan method to call.

generated/nidaqmx/stream_readers.py
mikeprosserni: That's fair. If Brad agrees, I can make all the waveform parameters in the stream readers required. Then only Task.read_waveform() would create waveform objects, right?

generated/nidaqmx/_library_interpreter.py
mikeprosserni: see: https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Fdaqmx%2Fapi%2Fnicai%2Fsource%2Fnicai%2Ftemplates%2FtemplateInternal.h&version=GBmain&_a=contents

zhindes: Bah forgot to hit comment before resolve. I see two options here:
1. Force the user to also allocate the shadow array
2. Leave an allocation "in the loop", keeping the interface clean, and then we can eliminate it later with some C API additions.

Let's leave it as-is, (2), I think.

generated/nidaqmx/_library_interpreter.py
zhindes: Hmm, after reading this, I'm less confident about my opinion above...

> 1. Force the user to also allocate the shadow array
> 2. Leave an allocation "in the loop", keeping the interface clean, and then we can eliminate it later with some C API additions.

Like you said, with (1) we could have the user do a single allocation, and then we return the waveform-views into them. That sounds nicer from a performance perspective, but it does make the interface very weird since the user passes us two "buffers" - a raw buffer and some waveforms. Or we return the waveforms for them.

Agh, my brain hurts. I have a slight preference for keeping the interface clean and keeping an intermediate allocation in the loop. But like very slight at this point. Leaking some cruft of the internal API doesn't seem entirely out of line for the opt-in performance of Stream Readers.

I hope Brad has a stronger opinion that I do.

mikeprosserni: I've gone ahead and implemented option 2, which makes the API nice and clean. (I also made all the waveform parameters required). 

If we still want to improve performance, we could allow the user to pass in the 'shadow array' (aka `read_array`) as an optional parameter. That would allow for max performance for users who need it, without compromising on the API too much. 

Or perhaps we could store the `read_array` on the DigitalMultiChannelReader object, so it could be reused across calls to read_waveforms()?

generated/nidaqmx/stream_readers.py
mikeprosserni: I have made all the waveform parameters in the stream readers required.

bkeryan: > The point of the stream readers is that they are performance optimized. Therefore we force callers to allocate memory for us. I don't think we should give the option to do it on their behalf.

I think we should really think through what this means, starting with the existing `read_many_sample` APIs.

If you want to read 1000 samples, you must allocate a NumPy array where one dimension is 1000. Okay.

If you want to read all available samples, you must allocate a NumPy array where one dimension is equal to the number of available samples. However, the number of available samples is always changing, so you can't predict it, and this case doesn't work: https://github.com/ni/nidaqmx-python/issues/421

Even if this did work, it would mean that you have to query the number of available samples before calling `reader.read_many_sample`. There are issues with this:
- This extra driver call is redundant because the underlying read function already queries this information. However, this isn't a performance issue because when a higher level of the API translates -1 into querying `Read.AvailSampPerChan`, it passes the queried value (instead of -1) to the next lower level of the API. 
- The number of samples to read when you specify -1 is not always the same as the `Read.AvailSampPerChan` attribute. It's the internal `Read.DefaultNumberOfSamplesToRead` attribute, which behaves differently for finite acquisitions (see AB#2692099). It also supports being queried before the task is started (see AB#2681936) and I think querying it does not verify the task.
- [Task._calculate_num_samps_per_chan](https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/task/_task.py#L296) queries multiple attributes in order to emulate the behavior of `Read.DefaultNumberOfSamplesToRead`, but I think it is missing the task state corner cases.

Anyway, I don't think it's desirable to force users of the more-advanced API to always preallocate their waveforms. It would be nice if users could ask DAQmx to allocate a waveform for them, or if they could control whether DAQmx is allowed to grow their waveform capacity or not.

In the .NET API, the stream readers support both [easy-to-use methods that allocate memory](https://www.ni.com/docs/en-US/bundle/ni-daqmx-net-framework-40-api-ref/page/ninetdaqmxfx40ref/html/overload_nationalinstruments_daqmx_analogsinglechannelreader_readwaveform.htm) and [harder-to-use memory-optimized methods](https://www.ni.com/docs/en-US/bundle/ni-daqmx-net-framework-40-api-ref/page/ninetdaqmxfx40ref/html/overload_nationalinstruments_daqmx_analogsinglechannelreader_memoryoptimizedreadwaveform.htm). The .NET API doesn't have a `Task.Read` method, so the "easy-to-use" methods are on the stream reader classes.

The .NET memory-optimized methods allow the user to specify a [reallocation policy](https://www.ni.com/docs/en-US/bundle/ni-daqmx-net-framework-40-api-ref/page/ninetdaqmxfx40ref/html/t_nationalinstruments_daqmx_reallocationpolicy.htm), so they can choose whether DAQmx must not reallocate or may grow the buffer. Allowing DAQmx to grow the buffer still provides a performance speedup when the buffer is large enough and does not need to be resized.

It's true that the Python stream readers do not support the easy-to-use case. They do not support the pre-allocated array case correctly either, when using READ_ALL_AVAILABLE. @mikeprosserni's waveform code handles this better than `read_many_sample`, because it allows you to pre-allocate more samples than you need rather than forcing you to pre-allocate exactly the right number of samples. But I'm not convinced that the stream readers should always be 100% allocation-free.

bkeryan (changes requested):
I am not done with my first review of this PR.

generated/nidaqmx/_library_interpreter.py
bkeryan: Using DAQmxInternalReadDigitalWaveform for multi-channel is sort of a temporary condition. There should be a DAQmxInternalReadDigitalWaveformPerChan, but I didn't add one last year because it was not needed by the .NET API, for .NET-specific reasons.

Since DAQmxInternalReadDigitalWaveformPerChan doesn't exist today, if you added it to DAQmx and used it in the Python API, then it would require a newer DAQmx version like 2025 Q4 or later, so you would still want to fall back to DAQmxInternalReadDigitalWaveform when the per-chan version is not available.

tests/component/test_task_read_waveform_di.py
```diff
@@ -0,0 +1,176 @@
...
+    task.timing.cfg_samp_clk_timing(1000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=50)
+    return task
+
+
+@pytest.fixture
+def di_multi_channel_timing_task(
+    task: nidaqmx.Task, sim_6363_device: nidaqmx.system.Device
+) -> nidaqmx.Task:
+    task.di_channels.add_di_chan(
+        flatten_channel_string(sim_6363_device.di_lines.channel_names[:8]),
```
bkeryan, line 42: @mikeprosserni @jfriedri-ni Please test a port channel as well. I think the line order is reversed.

This is related to [AB#3178052](https://dev.azure.com/ni/DevCentral/_workitems/edit/3178052).

tests/component/test_task_read_waveform_di.py
```diff
@@ -0,0 +1,176 @@
...
+        line_grouping=LineGrouping.CHAN_PER_LINE,
```
bkeryan, line 43: @mikeprosserni Please test multiple channels that each contain multiple lines.

See [DigitalMultiChannelReaderTests.cs](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Fdaqmx%2Ftests%2FnidaqmxNETAPITester%2Fsource%2FnidaqmxNETAPITester%2FTests%2FComponent%2FDAQmx%2FDigitalMultiChannelReaderTests.cs&version=GBmain&_a=contents) and [DigitalReaderTestsBase.cs](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Fdaqmx%2Ftests%2FnidaqmxNETAPITester%2Fsource%2FnidaqmxNETAPITester%2FTests%2FComponent%2FDAQmx%2FDigitalReaderTestsBase.cs&_a=contents&version=GBmain) for more channel groupings that are worth testing.

tests/component/test_task_read_waveform_di.py
jfriedri-ni: There is one port test so far `test___digital_single_channel_port_uint32_reader___read_waveform___returns_valid_waveform` and it's marked with `xfail` and the work item number.

generated/nidaqmx/stream_readers.py
zhindes: > It's true that the Python stream readers do not support the easy-to-use case. They do not support the pre-allocated array case correctly either, when using READ_ALL_AVAILABLE. @mikeprosserni's waveform code handles this better than read_many_sample, because it allows you to pre-allocate more samples than you need rather than forcing you to pre-allocate exactly the right number of samples. But I'm not convinced that the stream readers should always be 100% allocation-free.

Good points. Some more thoughts...
* Existing non-waveform methods do not allocate, which is part of the `READ_ALL_AVAILABLE` problem.
* It is OK for for stream readers to allocate on behalf of the user, but we should make that a clear opt-in as part of the interface. Not passing in a waveform/array is pretty clear, IMO.
* Therefore, I think it is OK if the waveform methods introduce this new methodology as long as we're 100% consistent between all of them. They either all support it or none do.
* It is OK that digital_waveform**s** does a temporary allocation until we can make that better in the future, but let's make that transparent for now.
* We can make a pass later to add new methods or loosen up the interface to allow existing methods to allocate to address things like `REAL_ALL_AVAILABLE`.

All that being said, I kind of like @mikeprosserni 's latest commit that leans into maintaining existing behavior. Then we can come back later to clean stuff up as an explicit action.

bkeryan: > [Task._calculate_num_samps_per_chan](https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/task/_task.py#L296) queries multiple attributes in order to emulate the behavior of Read.DefaultNumberOfSamplesToRead, but I think it is missing the task state corner cases.

I filed #812 about this, and also #811 about a related property.

tests/component/test_task_read_waveform_di.py
bkeryan: Thanks, I overlooked this because "Large diffs are not rendered by default."

src/handwritten/task/_task.py
```diff
@@ -847,32 +847,61 @@ def read_waveform(self, number_of_samples_per_channel=READ_ALL_AVAILABLE,
...
+                    timeout,
+                    digital_waveform,
+                    self._in_stream.waveform_attribute_mode,
+                )
+                return digital_waveform
+            else:
+                digital_waveforms = [
+                    DigitalWaveform(
+                        number_of_samples_per_channel,
+                        self.in_stream.di_num_booleans_per_chan
```
bkeryan: This is incorrect in some multi-channel cases. See the docs for `di_num_booleans_per_chan`:

```
    @property
    def di_num_booleans_per_chan(self):
        """
        int: Indicates the number of booleans per channel that NI-DAQmx
            returns in a sample for line-based reads. If a channel has
            fewer lines than this number, the extra booleans are False.
```

`di_num_booleans_per_chan` describes the format of the flat array used by `read_digital_lines` and `internal_read_digital_waveform`.

However, waveform reads in LabVIEW and C# use the correct number of lines for each channel. For example, if your channels are `["Dev1/port0/line0:31", "Dev1/port1/line0:3", "Dev1/port1/line4:7,Dev1/port2/line0:7"]`, then DAQmx Read should return 3 waveforms with 32, 4, and 12 lines, respectively.

How do you know the correct number of lines for each waveform? `DAQmxInternalReadDigitalWaveform` returns this information via the `bytesPerChanArray` parameter.

Doesn't this cause a chicken-and-egg problem with waveform allocation? Sort of.
- When allocating new waveforms, you must call `DAQmxInternalReadDigitalWaveform` first so that you know how many lines to allocate to each waveform. Note that you can still pre-allocate the `ExtendedPropertyDictionary` and specify `copy_extended_properties=False` or whatever. Also note that you can load data into an empty waveform without copying, but I don't think you can currently change the number of lines. 
- When reusing existing waveforms, you should probably validate that each one has the right number of lines. Please look at how the .NET API handles this.

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -457,6 +459,158 @@ class LibraryInterpreter(BaseInterpreter):
...
+            dt_array,
+            0 if t0_array is None else t0_array.size,
+            self._get_wfm_attr_callback(properties),
+            None,
+            read_array,
+            read_array.size,
+            ctypes.byref(samps_per_chan_read),
+            ctypes.byref(num_bytes_per_samp),
+            None,
+            0,
```
bkeryan: Add another parameter for `bytes_per_chan_array`.

tests/component/test_task_read_waveform_di.py
jfriedri-ni: One configuration not covered by those tests is a channel created with _both_ lines and ports.

This might not be a useful configuration because reading waveforms usually implies that the user wants hardware-timed sampling and DAQ devices only have _one_ port with that capability. Using lines and a port in a task's channel would then require software-timed acquisition.

Still, a user can create a task like this without an error from the driver, and the resulting bit order is unsurprisingly a blend of the two
- LSB-first for the lines
- MSB-first for the port

When I tried this in Python using the lines from `port0` line-by-line and then all of `port2`, the waveform attributes had this

'NI_LineNames': 'NIPCIe-6363/port0/**line0**,NIPCIe-6363/port0/**line1**,NIPCIe-6363/port0/**line2**,(so many...),NIPCIe-6363/port0/**line30**,NIPCIe-6363/port0/**line31**,NIPCIe-6363/port2/**line7**,NIPCIe-6363/port2/**line6**,(snip again),NIPCIe-6363/port2/**line1**,NIPCIe-6363/port2/**line0**'

where the line-by-line names are _increasing_ from 0...31 and the by-port names are _decreasing_ from 7...0.

How well should we accommodate this?

tests/component/test_stream_readers_ai.py
```diff
@@ -167,9 +167,10 @@ def test___analog_single_channel_reader___read_waveform_feature_disabled___raise
     ai_single_channel_task_with_timing: nidaqmx.Task,
 ) -> None:
     reader = AnalogSingleChannelReader(ai_single_channel_task_with_timing.in_stream)
+    waveform = AnalogWaveform[numpy.float64](50)
```
bkeryan: https://github.com/ni/nidaqmx-python/pull/807#discussion_r2283511149

> Just pass the sample count. AnalogWaveform will allocate that many samples. The default dtype is np.float64.

Please remember feedback from previous PRs and continue to apply it.

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -457,6 +459,158 @@ class LibraryInterpreter(BaseInterpreter):
...
+            dt_array = numpy.zeros(1, dtype=numpy.int64)
+        else:
+            t0_array = None
+            dt_array = None
+
+        error_code, samples_read = self._internal_read_digital_waveform(
+            task_handle,
+            number_of_samples_per_channel,
+            timeout,
+            FillMode.GROUP_BY_CHANNEL.value,
```
jfriedri-ni, line 488: Just to confirm my understanding... the API deliberately hardcodes the fill mode when reading waveforms and does not allow the user to request grouping by scan?

tests/component/test_task_read_waveform_di.py
bkeryan: Using `DAQmxInternalReadDigitalWaveform` instead of `DAQmxReadDigitalLines` makes DAQmx reverse the line order for ports, and it should set the `NI_ChannelName` and `NI_LineNames` waveform attributes appropriately.

I don't think the .NET tests cover the case you described.

tests/component/test_stream_readers_di.py
```diff
@@ -630,3 +711,611 @@ def test___digital_multi_channel_reader___read_many_sample_port_uint32_with_wron
...
+    waveforms = [
+        DigitalWaveform(samples_to_read),
+        DigitalWaveform(samples_to_read),
+        DigitalWaveform(samples_to_read),
+        DigitalWaveform(samples_to_read),
+        DigitalWaveform(samples_to_read),
+        DigitalWaveform(samples_to_read),
+        DigitalWaveform(samples_to_read),
+        DigitalWaveform(samples_to_read),
+    ]
```
bkeryan:
```suggestion
    waveforms = [DigitalWaveform(samples_to_read) for _ in range(num_channels)]
```

bkeryan (changes requested):
Done with my first-pass review

src/codegen/templates/_library_interpreter.py.mako
bkeryan: This API deliberately hardcodes the fill mode, in general. This is not ideal. See these open issues:
- https://github.com/ni/nidaqmx-python/issues/90 - requesting support for NumPy views in either C or Fortran layout
- https://github.com/ni/nidaqmx-python/issues/528 - consequence of using non-interleaved fill mode

From a public API perspective, waveform reads don't have a fill mode. Analog waveforms use a separate array for each channel. DIO line read/write arrays have an additional axis, so the fill mode selects between (samples x channels x lines) and (I think) (channels x samples x lines), but this doesn't affect waveforms because they represent each channel with a separate array that has a fixed order of axes.

tests/component/test_stream_readers_ai.py
mikeprosserni: oops, fixed

tests/component/test_stream_readers_di.py
mikeprosserni: done

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: added bytes_per_chan_array

src/handwritten/task/_task.py
mikeprosserni: For the stream reader use case (where the user provides the waveforms), I've updated `read_digital_waveforms()` to use `bytes_per_chan_array` when copying the `read_arrary` into `waveform.data[:]`. This throws a ValueException if the number of signals doesn't match the size of the waveform.

For the task use case (where the user doesn't provide the waveforms), I've added `read_new_digital_waveforms()`, which allocates new waveforms after calling `DAQmxInternalReadDigitalWaveform`. This uses `DigitalWaveform.from_lines()` and re-uses the slices of the `read_array`, without making copies.

I wasn't able to look at how the .NET API handles this, as I could not find where the C# DigitalMultiChannelReader is implemented. Can you point me to it?

mikeprosserni: Also, I added these tests:
test_stream_readers_di.py:
test___digital_multi_channel_different_lines_reader___read_waveforms___returns_valid_waveforms
test___digital_multi_channel_different_lines_reader___read_mismatched_waveforms___throws_exception

test_task_read_waveform.py:
test___digital_multi_channel___read_waveform_different_lines___returns_valid_waveforms

src/codegen/templates/_base_interpreter.py.mako
mikeprosserni: this return value has been changed to `int` (samples actually read), like the others.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6780,11 +6846,11 @@ def _internal_read_digital_waveform(
             self._get_wfm_attr_callback(properties),
             None,
             read_array,
-            read_array.size,
+            0 if read_array is None else read_array.size,
```
bkeryan: I am not aware of a use case for omitting read_array. None of the other read functions let you do this.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6728,6 +6736,63 @@ def read_digital_waveforms(
...
+
+        waveforms = []
+        for i in range(channel_count):
+            signal_count = bytes_per_chan_array[i]
+            waveform = DigitalWaveform.from_lines(
+                array=read_array[i, :, :signal_count],
+                copy=False,
+                sample_count=samples_read,
+                signal_count=signal_count,
+                extended_properties=properties[i] if properties else None)
```
bkeryan, line 6786: This makes an unnecessary copy of the properties. Use the constructor instead of `from_lines` and set `copy_extended_properties=False`. I don't think you really need `from_lines` anyway because you're passing in a NumPy array view and saving a reference to it.

src/handwritten/task/_task.py
bkeryan: > I wasn't able to look at how the .NET API handles this, as I could not find where the C# DigitalMultiChannelReader is implemented. Can you point me to it?

It's C++/CLI: 
- https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/daqmx/api/MSDaqDotNET/source/ninetai2005/API/DigitalMultiChannelReader.cpp&version=GBmain&_a=contents
- https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Fdaqmx%2Fapi%2FMSDaqDotNET%2Fsource%2Fninetai2005%2FAPI%2FInfrastructure%2FDigitalMultiChannelReaderImpl.cpp&version=GBmain&_a=contents

tests/component/test_task_read_waveform_di.py
mikeprosserni: I've added these tests that test multiple channels that each contain multiple lines (and a different number of lines in each channel)
```
stream: test___digital_multi_channel_different_lines_reader___read_waveforms___returns_valid_waveforms
stream: test___digital_multi_channel_different_lines_reader___read_mismatched_waveforms___throws_exception
task: test___digital_multi_channel___read_waveform_different_lines___returns_valid_waveforms
```

And I've added these tests that test mixing lines and ports, either in multiple channels, or together in a single channel:
```
stream: test___digital_single_channel_lines_and_port___read_waveform___returns_valid_waveform
stream: test___digital_multi_channel_lines_and_port_reader___read_waveforms___returns_valid_waveforms
task: test___digital_single_channel___read_waveform_lines_and_port___returns_valid_waveform
task: test___digital_multi_channel___read_waveform_lines_and_port___returns_valid_waveforms
```

These tests also assert that NI_ChannelName and NI_LineNames were set as expected (via `waveform.channel_name` and `waveform._get_signal_names()`)

generated/nidaqmx/_library_interpreter.py
mikeprosserni: Oops, I was trying out something and neglected to revert this before committing. I've reverted it now.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: Updated to use the constructor instead of from_lines.

generated/nidaqmx/stream_readers.py
mikeprosserni: Task AB#3228924 (which I'll be working on next) will add the ability for the stream reader methods to resize the waveform, if the user allows (with a ReallocationPolicy parameter). This discussion can continue on that PR as needed.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6779,11 +6779,12 @@ def read_new_digital_waveforms(
...
         for i in range(channel_count):
             signal_count = bytes_per_chan_array[i]
-            waveform = DigitalWaveform.from_lines(
-                array=read_array[i, :, :signal_count],
-                copy=False,
+            waveform = DigitalWaveform(
                 sample_count=samples_read,
                 signal_count=signal_count,
+                dtype=numpy.uint8,
+                data=read_array[i, :, :signal_count],
```
bkeryan: FYI: I think you can omit sample_count, signal_count, and dtype since you are providing a NumPy array.

src/handwritten/task/_task.py
mikeprosserni: The only version of `DigitalMultiChannelReaderImpl::ReadWaveform()` that takes in existing waveforms (on line 186 of DigitalMultiChannelReaderImpl.cpp) just throws NotImplemented, so I think the .NET API just doesn't handle this.

Is my current behavior (on line 550 of _library_interpreter.py.mako, throw a ValueError if the waveform has the wrong number of lines) good for the python API?

bkeryan: > The only version of `DigitalMultiChannelReaderImpl::ReadWaveform()` that takes in existing waveforms (on line 186 of DigitalMultiChannelReaderImpl.cpp) just throws NotImplemented, so I think the .NET API just doesn't handle this.

Yeah, I forgot. Memory-optimized digital waveform reads are not supported in the .NET API for .NET-specific reasons.

> Is my current behavior (on line 550 of _library_interpreter.py.mako, throw a ValueError if the waveform has the wrong number of lines) good for the python API?

Yes, I think this is the best you can do with the current DAQmxInternalReadDigitalWaveform entry point.

Ideally, you would pass the waveform sizes to DAQmxInternalReadDigitalWaveform and let it validate that they are correct. I think a per-chan version of this function would probably do that.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: I've removed signal_count and dtype, but I'm keeping sample_count since it should handle short reads that way.

zhindes (approved):

PR #819, opened 2025-09-03
Handle reads that don't match Waveform length

mikeprosserni:
- [ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).

- [ ] I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.

- [ ] I've added tests applicable for this pull request

TODO: Include high-level description of the changes in this pull request.

TODO: Justify why this contribution should be part of the project.

### DIFF
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -6406,7 +6406,7 @@ class LibraryInterpreter(BaseInterpreter):
         if t0_array is not None and dt_array is not None:
             self._set_waveform_timings([waveform], t0_array, dt_array)
 
-        # TODO: AB#3228924 - if the read was short, set waveform.sample_count before throwing the exception
+        waveform.sample_count = samples_read
         self.check_for_error(error_code, samps_per_chan_read=samples_read)
         return samples_read
 
@@ -6444,7 +6444,9 @@ class LibraryInterpreter(BaseInterpreter):
         if t0_array is not None and dt_array is not None:
             self._set_waveform_timings(waveforms, t0_array, dt_array)
 
-        # TODO: AB#3228924 - if the read was short, set waveform.sample_count before throwing the exception
+        for waveform in waveforms:
+            waveform.sample_count = samples_read
+            
         self.check_for_error(error_code, samps_per_chan_read=samples_read)
         return samples_read
 
@@ -6675,7 +6677,7 @@ class LibraryInterpreter(BaseInterpreter):
         if t0_array is not None and dt_array is not None:
             self._set_waveform_timings([waveform], t0_array, dt_array)
 
-        # TODO: AB#3228924 - if the read was short, set waveform.sample_count before throwing the exception
+        waveform.sample_count = samples_read
         self.check_for_error(error_code, samps_per_chan_read=samples_read)
         return samples_read
 
@@ -6705,7 +6707,7 @@ class LibraryInterpreter(BaseInterpreter):
         # Since there's no DAQmxInternalReadDigitalWaveformPerChan, we have to allocate a
         # temporary contiguous array to read the data from multiple channels into.
         read_array = numpy.zeros(
-            (channel_count, number_of_samples_per_channel, number_of_signals_per_sample),
+            channel_count * number_of_samples_per_channel * number_of_signals_per_sample,
             dtype=numpy.uint8)
 
         bytes_per_chan_array = numpy.zeros(channel_count, dtype=numpy.uint32)
@@ -6722,17 +6724,20 @@ class LibraryInterpreter(BaseInterpreter):
             bytes_per_chan_array,
         )
 
+        channel_array_length = channel_count * samples_read * number_of_signals_per_sample
+        channel_array = read_array[:channel_array_length].reshape(
+            channel_count, samples_read, number_of_signals_per_sample)
+
         for i, waveform in enumerate(waveforms):
-            waveform_signals = waveform.data.shape[1]
-            channel_signals = bytes_per_chan_array[i]
-            if waveform_signals != channel_signals:
-                raise ValueError(f"waveforms[{i}].data has {waveform_signals} signals, but expected {channel_signals}")
-            waveform.data[:] = read_array[i, :, :channel_signals]
+            waveform_signal_count = waveform.data.shape[1]
+            channel_signal_count = bytes_per_chan_array[i]
+            if waveform_signal_count != channel_signal_count:
+                raise ValueError(f"waveforms[{i}].data has {waveform_signal_count} signals, but expected {channel_signal_count}")
+            waveform.data[:] = channel_array[i, :, :channel_signal_count]
 
         if t0_array is not None and dt_array is not None:
             self._set_waveform_timings(waveforms, t0_array, dt_array)
 
-        # TODO: AB#3228924 - if the read was short, set waveform.sample_count before throwing the exception
         self.check_for_error(error_code, samps_per_chan_read=samples_read)
         return samples_read
 
@@ -6759,7 +6764,7 @@ class LibraryInterpreter(BaseInterpreter):
             dt_array = None
 
         read_array = numpy.zeros(
-            (channel_count, number_of_samples_per_channel, number_of_signals_per_sample),
+            channel_count * number_of_samples_per_channel * number_of_signals_per_sample,
             dtype=numpy.uint8)
 
         bytes_per_chan_array = numpy.zeros(channel_count, dtype=numpy.uint32)
@@ -6776,12 +6781,16 @@ class LibraryInterpreter(BaseInterpreter):
             bytes_per_chan_array,
         )
 
+        channel_array_length = channel_count * samples_read * number_of_signals_per_sample
+        channel_array = read_array[:channel_array_length].reshape(
+            channel_count, samples_read, number_of_signals_per_sample)
+
         waveforms = []
         for i in range(channel_count):
-            signal_count = bytes_per_chan_array[i]
+            channel_signal_count = bytes_per_chan_array[i]
             waveform = DigitalWaveform(
                 sample_count=samples_read,
-                data=read_array[i, :, :signal_count],
+                data=channel_array[i, :, :channel_signal_count],
                 copy_extended_properties=False,
                 extended_properties=properties[i] if properties else None)
             waveforms.append(waveform)
--- a/src/codegen/templates/_library_interpreter.py.mako
+++ b/src/codegen/templates/_library_interpreter.py.mako
@@ -225,7 +225,7 @@ class LibraryInterpreter(BaseInterpreter):
         if t0_array is not None and dt_array is not None:
             self._set_waveform_timings([waveform], t0_array, dt_array)
 
-        # TODO: AB#3228924 - if the read was short, set waveform.sample_count before throwing the exception
+        waveform.sample_count = samples_read
         self.check_for_error(error_code, samps_per_chan_read=samples_read)
         return samples_read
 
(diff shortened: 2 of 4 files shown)


zhindes (approved):

tests/component/readers/test_analog_multi_channel_reader.py
bkeryan, line 1: [previous PR] This should be tests/component/stream_readers/ to parallel the production code.

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -543,17 +545,20 @@ class LibraryInterpreter(BaseInterpreter):
             bytes_per_chan_array,
         )
 
+        channel_array_length = channel_count * samples_read * number_of_signals_per_sample
+        channel_array = read_array[:channel_array_length].reshape(
+            channel_count, samples_read, number_of_signals_per_sample)
```
bkeryan: Why not use GROUP_BY_SCAN_NUMBER, like the .NET API does? With that data layout, short reads are pre-squashed and DAQmx doesn't need to copy data around in order to squash them.

See https://github.com/ni/nidaqmx-python/issues/528 for details and examples.

src/codegen/templates/task/_in_stream.py.mako
```diff
@@ -27,13 +27,14 @@ class InStream:
...
     """
-    __slots__ = ('_task', '_handle', '_interpreter', '_timeout', '_waveform_attribute_mode')
+    __slots__ = ('_task', '_handle', '_interpreter', '_timeout', '_reallocation_policy', '_waveform_attribute_mode')
 
     def __init__(self, task, interpreter):
         self._task = task
         self._handle = task._handle
         self._interpreter = interpreter
         self._timeout = 10.0
+        self._reallocation_policy = ReallocationPolicy.DO_NOT_REALLOCATE
```
bkeryan: Why is this a stream property? In .NET it's an optional parameter on the memory-optimized read methods, which defaults to ToGrow.

src/handwritten/stream_readers/analog_multi_channel_reader.py
```diff
@@ -211,12 +211,15 @@ def read_waveforms(
         
         for i, waveform in enumerate(waveforms):
             if number_of_samples_per_channel > waveform.sample_count:
```
bkeryan: Users may hit this when reusing a waveform after a short read. It would be better to limit based on capacity.

src/handwritten/stream_readers/analog_multi_channel_reader.py
```diff
@@ -4,7 +4,7 @@
 from nidaqmx import DaqError
 
 from nidaqmx._feature_toggles import WAVEFORM_SUPPORT, requires_feature
-from nidaqmx.constants import FillMode, READ_ALL_AVAILABLE
+from nidaqmx.constants import FillMode, READ_ALL_AVAILABLE, ReallocationPolicy
 from nidaqmx.error_codes import DAQmxErrors
 from nitypes.waveform import AnalogWaveform
```
bkeryan: FYI: https://github.com/ni/python-styleguide/blob/main/docs/Coding-Conventions.md#o15-%EF%B8%8F-do-use-absolute-imports-

> [O.1.5] ✔️ DO Use absolute imports 💻
>🐍 This rule stems from [PEP 8](https://www.python.org/dev/peps/pep-0008)

> 💻 This rule is enforced by error code I252

You didn't get an error due to https://github.com/ni/nidaqmx-python/issues/556

tests/component/readers/test_analog_multi_channel_reader.py
mikeprosserni: I'm fixing this in https://github.com/ni/nidaqmx-python/pull/825

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: yes, GROUP_BY_SCAN_NUMBER works nicely

src/codegen/templates/task/_in_stream.py.mako
mikeprosserni: I've changed reallocation_policy to an optional parameter instead of a stream property.

src/handwritten/stream_readers/analog_multi_channel_reader.py
mikeprosserni: I'm changing the relative imports to absolute imports in https://github.com/ni/nidaqmx-python/pull/825

src/handwritten/stream_readers/analog_multi_channel_reader.py
mikeprosserni: I changed this to compare against capacity, and added tests that reuse waveforms after short reads.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6759,7 +6760,7 @@ def read_new_digital_waveforms(
             dt_array = None
 
         read_array = numpy.zeros(
-            (channel_count, number_of_samples_per_channel, number_of_signals_per_sample),
+            channel_count * number_of_samples_per_channel * number_of_signals_per_sample,
```
bkeryan: read_new_digital_waveforms should use GROUP_BY_SCAN_NUMBER too.

generated/nidaqmx/stream_readers/_analog_multi_channel_reader.py
bkeryan, line 1: Delete stream_readers.py and stream_writers.py here or in a separate PR, I think they are redundant with the subpackage.

generated/nidaqmx/stream_readers/_analog_multi_channel_reader.py
```diff
@@ -210,13 +214,17 @@ def read_waveforms(
...
+                    raise DaqError(
+                        f'The waveform at index {i} does not have enough space ({waveform.capacity}) to hold '
+                        f'the requested number of samples ({number_of_samples_per_channel}). Please provide larger '
+                        'waveforms or adjust the number of samples requested.',
+                        DAQmxErrors.READ_BUFFER_TOO_SMALL, task_name=self._task.name)
+                
+            waveform.sample_count = number_of_samples_per_channel
 
         return self._interpreter.read_analog_waveforms(
             self._handle,
```
bkeryan: This leaves waveform.sample_count set to the wrong value after a read error.

If you are trying to read 1000 samples and you get a timeout or another error after reading 500, then waveform.sample_count should be 500, not 1000. Many other errors will return 0 samples read, but we still leave sample_count set to 1000.

The partial read sample count is reported via DaqReadError.samps_per_chan_read. The other read methods cannot resize the array that they are reading into, so we leave it to the customer to trap DaqReadError and query its samps_per_chan_read property. For waveforms, it seems like we ought to do this automatically: if we get DaqReadError, set sample_count to ex.samps_per_chan_read and re-raise the exception. For other exceptions, set sample_count to 0 and re-raise.

generated/nidaqmx/stream_readers/_analog_single_channel_reader.py
```diff
@@ -115,6 +115,7 @@ def read_waveform(
         self, 
         waveform: AnalogWaveform[numpy.float64],
         number_of_samples_per_channel: int = READ_ALL_AVAILABLE, 
+        reallocation_policy: ReallocationPolicy = ReallocationPolicy.DO_NOT_REALLOCATE,
```
bkeryan: In .NET this defaults to TO_GROW, doesn't it?

generated/nidaqmx/_library_interpreter.py
mikeprosserni: done

generated/nidaqmx/stream_readers/_analog_multi_channel_reader.py
mikeprosserni: I've deleted both in this PR.

generated/nidaqmx/stream_readers/analog_multi_channel_reader.py
```diff
@@ -214,16 +214,17 @@ def read_waveforms(
                 DAQmxErrors.MISMATCHED_INPUT_ARRAY_SIZES, task_name=self._task.name)
         
         for i, waveform in enumerate(waveforms):
-            if number_of_samples_per_channel > waveform.sample_count:
+            if number_of_samples_per_channel > waveform.capacity:
```
bkeryan: I think this should be `waveform.start_index + number_of_samples_per_channel > waveform.capacity`, since we are not resetting start_index to 0.

generated/nidaqmx/stream_readers/_analog_single_channel_reader.py
mikeprosserni: Yes, according to [this documentation](https://www.ni.com/docs/en-US/bundle/ni-daqmx-.net-framework-4.5.1-class-library-getting-started/page/netdaqmxmemoptanalogreads.html#:~:text=Getting%20Started-,Memory%2DOptimized%20Analog%20Reads,C%23). 

I've updated my changes so reallocation_policy defaults to TO_GROW.

generated/nidaqmx/stream_readers/analog_multi_channel_reader.py
mikeprosserni: I updated tests (for all four analog/digital and single/multi variants) with `waveform = AnalogWaveform(10, start_index=3, capacity=13)` to confirm this was a problem, and then changed this to `if waveform._start_index + number_of_samples_per_channel > waveform.capacity:` and `waveform.capacity = waveform._start_index + number_of_samples_per_channel` to fix the tests.

generated/nidaqmx/stream_readers/_analog_multi_channel_reader.py
mikeprosserni: Partial read sample count handling is happening in _library_interpreter.py on lines 6408, 6451, 6685, and 6737. In each case, the `samples_read` that was returned by `_internal_read_x_waveform()` is used to set the `waveform.sample_count` before `check_for_error()` is called (which is what raises the DaqReadError). 

I've moved the `waveform.sample_count = number_of_samples_per_channel` lines from the channel readers into the library interpreter to make things clearer and more consistent.

bkeryan: Oops, I overlooked that.

src/handwritten/stream_readers/_analog_multi_channel_reader.py
```diff
@@ -210,13 +214,15 @@ def read_waveforms(
...
         
         for i, waveform in enumerate(waveforms):
-            if number_of_samples_per_channel > waveform.sample_count:
-                # TODO: AB#3228924 - if allowed by the caller, increase the sample count of the waveform
-                raise DaqError(
-                    f'The waveform at index {i} does not have enough space ({waveform.sample_count}) to hold '
-                    f'the requested number of samples ({number_of_samples_per_channel}). Please provide larger '
-                    'waveforms or adjust the number of samples requested.',
-                    DAQmxErrors.READ_BUFFER_TOO_SMALL, task_name=self._task.name)
+            if waveform._start_index + number_of_samples_per_channel > waveform.capacity:
```
bkeryan, line 217: _start_index is private because it starts with an underscore. Consider adding a public start_index property to the waveforms and updating this. It doesn't have to gate this PR, though.

mikeprosserni: Will do - AB#3258929

PR #816, opened 2025-09-02
Refactor stream_readers and stream_writers into smaller files

mikeprosserni:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).
- [ ] ~~I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~
- [ ] ~~I've added tests applicable for this pull request~~

### What does this Pull Request accomplish?

`stream_readers.py` and `stream_writers.py` were very large files, which made development and review more difficult. This PR splits the classes in them out into individual files to make things more manageable. 

I was careful to keep each class unchanged through this refactor. I copy-pasted each entire class and made no edits within the classes. The only changes would be in the imports sections.

### Why should this Pull Request be merged?

This should make reviewing future changes easier.

### What testing has been done?

- All existing tests pass
- lint and mypy report no errors
- I checked the generated sphinx documentation for any changes. The files are largely unchanged, with only a few minor modifications:
  - An extra top level blurb: "NI-DAQmx stream writers.  This package provides classes for writing samples to NI-DAQmx tasks."
  - Source code links changed from (for example) `stream_writers.html` to `stream_writers/analog_multi_channel_writer.html`
  - `<nidaqmx.stream_writers.UnsetAutoStartSentinel object>` changed to `<nidaqmx.stream_writers._channel_writer_base.UnsetAutoStartSentinel object>`

image

image

PR #825, opened 2025-09-08
Address additional feedback on the refactoring of stream_readers and stream_writers

mikeprosserni:

- Make the stream_reader and stream_writer modules private (like the task submodules)
- Fix test folder structure to parallel the production code
- Fix `AUTO_START_UNSET` compatibility breakage
- Change relative imports to absolute imports

Addressing these comments:
https://github.com/ni/nidaqmx-python/pull/819#discussion_r2325883663
https://github.com/ni/nidaqmx-python/pull/816#discussion_r2325654273
https://github.com/ni/nidaqmx-python/pull/816#discussion_r2325674832
https://github.com/ni/nidaqmx-python/pull/819#discussion_r2327977205

Ran all autotests

Testing the import of AUTO_START_UNSET:
```
PS C:\dev\fireserp\nidaqmx-python> C:/dev/fireserp/nidaqmx-python/.venv/Scripts/python.exe -c "from nidaqmx.stream_writers import AUTO_START_UNSET; print('SUCCESS: AUTO_START_UNSET import works!'); print(f'AUTO_START_UNSET: {AUTO_START_UNSET}'); print(f'Type: {type(AUTO_START_UNSET)}')"
SUCCESS: AUTO_START_UNSET import works!
AUTO_START_UNSET: <nidaqmx.stream_writers._channel_writer_base.UnsetAutoStartSentinel object at 0x0000024392016790>
Type: <class 'nidaqmx.stream_writers._channel_writer_base.UnsetAutoStartSentinel'>
```

### DIFF
--- a/generated/nidaqmx/stream_writers/__init__.py
+++ b/generated/nidaqmx/stream_writers/__init__.py
@@ -12,7 +12,7 @@ from nidaqmx.stream_writers._analog_unscaled_writer import AnalogUnscaledWriter
 from nidaqmx.stream_writers._counter_writer import CounterWriter
 from nidaqmx.stream_writers._digital_single_channel_writer import DigitalSingleChannelWriter
 from nidaqmx.stream_writers._digital_multi_channel_writer import DigitalMultiChannelWriter
-from nidaqmx.stream_writers._channel_writer_base import AUTO_START_UNSET
+from nidaqmx.stream_writers._channel_writer_base import UnsetAutoStartSentinel, AUTO_START_UNSET
 
 __all__ = [
     'AnalogSingleChannelWriter', 
@@ -21,4 +21,5 @@ __all__ = [
     'CounterWriter',
     'DigitalSingleChannelWriter', 
     'DigitalMultiChannelWriter',
+    'UnsetAutoStartSentinel',
     'AUTO_START_UNSET']
--- a/generated/nidaqmx/stream_writers/_channel_writer_base.py
+++ b/generated/nidaqmx/stream_writers/_channel_writer_base.py
@@ -3,12 +3,13 @@ from nidaqmx.error_codes import DAQmxErrors
 
 
 class UnsetAutoStartSentinel:
-    pass
+    """Sentinel class for unset auto_start parameter."""
+    
+    def __init__(self):
+        raise RuntimeError("Cannot instantiate UnsetAutoStartSentinel. Use AUTO_START_UNSET instead.")
 
 
-AUTO_START_UNSET = UnsetAutoStartSentinel()
-
-del UnsetAutoStartSentinel
+AUTO_START_UNSET = object.__new__(UnsetAutoStartSentinel)
 
 
 class ChannelWriterBase:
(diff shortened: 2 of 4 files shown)


generated/nidaqmx/stream_writers/__init__.py
```diff
@@ -6,17 +6,19 @@
...
 
 __all__ = [
     'AnalogSingleChannelWriter', 
     'AnalogMultiChannelWriter',
     'AnalogUnscaledWriter', 
     'CounterWriter',
     'DigitalSingleChannelWriter', 
-    'DigitalMultiChannelWriter']
+    'DigitalMultiChannelWriter',
+    'AUTO_START_UNSET']
```
bkeryan, line 25: We should probably also export the sentinel class so clients can use it in type hints. If there's a way to do typeof(AUTO_START_UNSET) in type hints, I don't know what it is.

generated/nidaqmx/stream_readers/__init__.py
```diff
@@ -8,13 +8,13 @@
...
 from nidaqmx import DaqError
 
-from .analog_single_channel_reader import AnalogSingleChannelReader
-from .analog_multi_channel_reader import AnalogMultiChannelReader
-from .analog_unscaled_reader import AnalogUnscaledReader
-from .counter_reader import CounterReader
-from .digital_single_channel_reader import DigitalSingleChannelReader
-from .digital_multi_channel_reader import DigitalMultiChannelReader
-from .power_readers import (
+from ._analog_single_channel_reader import AnalogSingleChannelReader
```
bkeryan: FYI, ni-python-styleguide discourages relative imports, but these files are not style-checked yet.

mikeprosserni: I've (just now) changed all the relative imports to absolute.

generated/nidaqmx/stream_writers/__init__.py
mikeprosserni: Currently, UnsetAutoStartSentinel is deleted right after being declared, to prevent callers from importing it. Should I remove that `del UnsetAutoStartSentinel`, so that UnsetAutoStartSentinel can be exported also?

```
class UnsetAutoStartSentinel:
    pass


AUTO_START_UNSET = UnsetAutoStartSentinel()

del UnsetAutoStartSentinel
```

mikeprosserni: How about this instead, which allows exporting UnsetAutoStartSentinel, while not allowing users to accidently create their own?

```
class UnsetAutoStartSentinel:
    """Sentinel class for unset auto_start parameter."""
    
    def __init__(self):
        raise RuntimeError("Cannot instantiate UnsetAutoStartSentinel. Use AUTO_START_UNSET instead.")


AUTO_START_UNSET = object.__new__(UnsetAutoStartSentinel)
```

generated/nidaqmx/stream_readers.py
bkeryan, line 1: Not sure how I missed this, but I think the submodule is dead code. `import nidaqmx.stream_readers` imports the subpackage, doesn't it?

https://stackoverflow.com/questions/6393861/how-python-deals-with-module-and-package-having-the-same-name

mikeprosserni: I think you are right. I've removed stream_readers.py and stream_writers.py in https://github.com/ni/nidaqmx-python/pull/819 and everything works fine.

PR #826, opened 2025-09-09
Refactor tests for stream writers into smaller files

mikeprosserni:
- [ ] ~I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~
- [ ] ~I've added tests applicable for this pull request~

Along the same lines as my other PR https://github.com/ni/nidaqmx-python/pull/817, this splits up large test files into smaller more manageable test files. It also moves common helper functions into a conftest.py.

This should make reviewing future changes easier.

All the tests pass. Note that the writer tests require real hardware, which my machine now has:

image


tests/component/conftest.py
```diff
@@ -139,11 +190,17 @@ def _assert_equal_2d(data: list[list[float]], expected: list[list[float]], abs:
 
 _D = TypeVar("_D", bound=numpy.generic)
 
-VOLTAGE_EPSILON = 1e-3
```
zhindes, line 142: This was 1e-3 for AI and 1e-2 for AO before. For AI we use simulation so we can get exact values. For AO we use real hardware which is often uncalibrated, so we widened it a bit at some point. I think I would prefer if you create a new variable, VOLTAGE_EPSILON_FOR_AO with some sort of comment.

```python
# NOTE: We must use real signals for AO validation, but we aren't validating hardware accuracy here.
# This should be wide enough tolerance to allow for uncalibrated boards while still ensuring we are
# correctly configuring hardware.
VOLTAGE_EPSILON_FOR_AO = 1e-2
```

Also, feel free to rename the current ones to be less ambiguous, too. This whole block of 4 constants is for simulation for AI/Power.

Then make sure the new helpers you copied into here use that new value.

zhindes (changes requested):

tests/component/test_stream_writers_ao.py
bkeryan, line 1: What about test_stream_writers.py? Should it be split or deleted?

tests/component/stream_writers/test_analog_multi_channel_writer.py
```diff
@@ -0,0 +1,80 @@
...
+
+import numpy
+import pytest
+
+import nidaqmx
+from nidaqmx.stream_writers import AnalogMultiChannelWriter
+from tests.component.conftest import (
+    _get_expected_voltage_for_chan,
+    VOLTAGE_EPSILON,
+)
```
bkeryan: Note that conftest.py is really for fixtures and pytest callbacks, not general purpose helper functions. 

https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20935/Sharing-Code-in-Tests suggests putting other shared helper functions in `_utils.py`, `_foo_utils.py`, etc. 

I realize this feedback is a bit late since you already did this for the reader tests.

bkeryan (changes requested):

tests/component/conftest.py
bkeryan: If you put the helpers in `_ai_utils.py` etc. then there is no name conflict.

tests/component/test_stream_writers_ao.py
mikeprosserni: The four tests in it are duplicates of other tests that I already moved. I think we can just delete `test_stream_writers.py`.

tests/component/stream_writers/test_analog_multi_channel_writer.py
mikeprosserni: I've moved all the utility methods out of conftest.py, into `_utils.py`, `_analog_utils.py`, and `_digital_utils.py`.

tests/component/conftest.py
mikeprosserni: I've renamed these to `AI_VOLTAGE_EPSILON`, `AO_VOLTAGE_EPSILON`, and `RAW_VOLTAGE_EPSILON`. And I added that note to AO_VOLTAGE_EPSILON.

(There was enough overlap between AI and AO utilities that it didn't make sense to have separate files for them, so these needed different names)

zhindes (approved):

zhindes:
> Something weird is going on with the CI linter. It's reporting lots of errors that don't make sense. The first one is this:
> 
> `./tests/component/stream_readers/test_analog_multi_channel_reader.py:227:68: F821 undefined name 'VOLTAGE_EPSILON'`
> 
> but line 227 of [tests/component/stream_readers/test_analog_multi_channel_reader.py](https://github.com/ni/nidaqmx-python/blob/9cc7dd6f21581badeba47f74a7e34e0d70ed0b6d/tests/component/stream_readers/test_analog_multi_channel_reader.py) is this:
> 
> `in_stream.waveform_attribute_mode = WaveformAttributeMode.TIMING`
> 
> and `VOLTAGE_EPSILON` (without any prefix) doesn't appear anywhere in the codebase.
> 
> What's going on here? And how might I fix it?

The build checked out this ref, which looks wrong...

```
Checking out the ref
/usr/bin/git log -1 --format=%H
3a2a8256df2090abbbaef732d885069b799f0380
```

The build 18 hours ago said...

> Results for commit https://github.com/ni/nidaqmx-python/commit/deb610f25626c06502fc193838ecedfa2f27f5a7. ±
> Comparison against base commit https://github.com/ni/nidaqmx-python/commit/faf34c90c8404728a1a50eba51a40ddcb6f58058.
> ♻️ This comment has been updated with latest results.

deb610f is your commit from like an hour ago, so that last comment indicates it reran it and its fine?

This isn't helpful, probably. I'm confused, too :)

bkeryan:
> Something weird is going on with the CI linter. It's reporting lots of errors that don't make sense. The first one is this:
> 
> `./tests/component/stream_readers/test_analog_multi_channel_reader.py:227:68: F821 undefined name 'VOLTAGE_EPSILON'`
> 
> but line 227 of [tests/component/stream_readers/test_analog_multi_channel_reader.py](https://github.com/ni/nidaqmx-python/blob/9cc7dd6f21581badeba47f74a7e34e0d70ed0b6d/tests/component/stream_readers/test_analog_multi_channel_reader.py) is this:
> 
> `in_stream.waveform_attribute_mode = WaveformAttributeMode.TIMING`
> 
> and `VOLTAGE_EPSILON` (without any prefix) doesn't appear anywhere in the codebase.
> 
> What's going on here? And how might I fix it?

You added new test cases in https://github.com/ni/nidaqmx-python/pull/819/

There are no merge conflicts, but the new test cases use variables that you have renamed. I would rebase this branch on top of origin/master and update the variable names.

PR #827, opened 2025-09-10
Use the new waveform.start_index instead of _start_index

mikeprosserni:

Uses the new public `waveform.start_index` instead of the private `waveform._start_index`.

We shouldn't use private fields.

Ran autotests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ python-decouple = ">=3.8"
 click = ">=8.0.0"
 distro = { version = ">=1.9.0", platform = "linux" }
 requests = ">=2.25.0"
+typing_extensions = { version = ">=4.0.0" }
 nitypes = {version=">=0.1.0dev10", allow-prereleases=true}
 
 [tool.poetry.extras]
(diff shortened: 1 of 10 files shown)


zhindes (approved):

pyproject.toml
```diff
@@ -50,7 +50,8 @@ python-decouple = ">=3.8"
 click = ">=8.0.0"
 distro = { version = ">=1.9.0", platform = "linux" }
 requests = ">=2.25.0"
-nitypes = {version=">=0.1.0dev9", allow-prereleases=true}
+typing_extensions = { version = ">=4.0.0" }
```
bkeryan, line 53: This is fine. I have given up on backporting typing stuff to old Python versions, because typing_extensions is that backport and if I add my own helper module then some tools (like sphinx-autoapi) cannot recognize some typing usages (like TypeAliases).

Also, nitypes already depends on typing_extensions, so it's already required.

mikeprosserni: ok, I'll go ahead and submit this here.

PR #828, opened 2025-09-11
Add  AnalogSingleChannelWriter.write_waveform()

mikeprosserni:
- ~~[ ] I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~

- Add `LibraryInterpreter.write_analog_waveform()` and `LibraryInterpreter._internal_write_analog_waveform_per_chan()`
- Add `AnalogSingleChannelWriter.write_waveform()`
- Add support in `Task.write()` for `AnalogWaveform`
- Made `feature_toggle.raise_if_disabled()` public

[AB#3258807](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3258807)

for AnalogSingleChannelWriter:
```
test___analog_single_channel_reader___read_waveform_feature_disabled___raises_feature_not_supported_error
test___analog_single_channel_writer___write_waveform___updates_output
test___analog_single_channel_writer___write_waveform_with_float32_dtype___updates_output
test___analog_single_channel_writer___write_waveform_with_scaling___updates_output
test___analog_single_channel_writer___write_waveform_with_non_contiguous_data___updates_output
```

for Task:
```
test___task___write_waveform_feature_disabled___raises_feature_not_supported_error
test___task___write_linear_ramp_waveform___output_matches_final_value
test___task___write_waveform_with_auto_start___waveform_executes_successfully
test___task_with_multiple_channels___write_single_channel_waveform___raises_daq_error
test___task___write_waveform_with_float32_dtype___output_matches_final_value
test___task___write_waveform_with_scaling___output_matches_final_value
test___task___write_waveform_with_non_contiguous_data___updates_output
test___task___write_waveform_with_timing___all_samples_match_waveform_data
```

### DIFF
--- a/generated/nidaqmx/_base_interpreter.py
+++ b/generated/nidaqmx/_base_interpreter.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import abc
 import numpy
 from nitypes.waveform import AnalogWaveform, DigitalWaveform
-from typing import Sequence
+from typing import Any, Sequence
 from nidaqmx.constants import WaveformAttributeMode
 
 
@@ -1914,7 +1914,7 @@ class BaseInterpreter(abc.ABC):
     def write_analog_waveform(
         self,
         task_handle: object,
-        waveform: AnalogWaveform[numpy.float64],
+        waveform: AnalogWaveform[Any],
         auto_start: bool,
         timeout: float
     ) -> int:
--- a/src/handwritten/stream_writers/_analog_single_channel_writer.py
+++ b/src/handwritten/stream_writers/_analog_single_channel_writer.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import numpy
+from typing import Any
 
 from nidaqmx._feature_toggles import WAVEFORM_SUPPORT, requires_feature
 from nidaqmx.constants import FillMode
@@ -89,7 +90,7 @@ class AnalogSingleChannelWriter(ChannelWriterBase):
     @requires_feature(WAVEFORM_SUPPORT)
     def write_waveform(
         self, 
-        waveform: AnalogWaveform[numpy.float64], 
+        waveform: AnalogWaveform[Any], 
         timeout: float = 10.0
     ) -> int:
         """
@@ -105,7 +106,7 @@ class AnalogSingleChannelWriter(ChannelWriterBase):
         the device generated all samples.
 
         Args:
-            waveform (AnalogWaveform[numpy.float64]): Specifies the
+            waveform (AnalogWaveform[Any]): Specifies the
                 waveform to write to the task.
             timeout (Optional[float]): Specifies the amount of time in
                 seconds to wait for the method to write all samples.
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -3654,11 +3654,11 @@ class GrpcStubInterpreter(BaseInterpreter):
         waveform_attribute_mode: WaveformAttributeMode,
     ) -> Sequence[DigitalWaveform[numpy.uint8]]:
         raise NotImplementedError
-        
+
     def write_analog_waveform(
         self,
         task_handle: object,
-        waveform: AnalogWaveform[numpy.float64],
+        waveform: AnalogWaveform[typing.Any],
         auto_start: bool,
         timeout: float
     ) -> int:
--- a/generated/nidaqmx/stream_writers/_analog_single_channel_writer.py
+++ b/generated/nidaqmx/stream_writers/_analog_single_channel_writer.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import numpy
+from typing import Any
 
 from nidaqmx._feature_toggles import WAVEFORM_SUPPORT, requires_feature
 from nidaqmx.constants import FillMode
@@ -89,7 +90,7 @@ class AnalogSingleChannelWriter(ChannelWriterBase):
     @requires_feature(WAVEFORM_SUPPORT)
     def write_waveform(
         self, 
-        waveform: AnalogWaveform[numpy.float64], 
+        waveform: AnalogWaveform[Any], 
         timeout: float = 10.0
     ) -> int:
         """
@@ -105,7 +106,7 @@ class AnalogSingleChannelWriter(ChannelWriterBase):
         the device generated all samples.
 
         Args:
-            waveform (AnalogWaveform[numpy.float64]): Specifies the
+            waveform (AnalogWaveform[Any]): Specifies the
                 waveform to write to the task.
             timeout (Optional[float]): Specifies the amount of time in
                 seconds to wait for the method to write all samples.
--- a/src/codegen/templates/_base_interpreter.py.mako
+++ b/src/codegen/templates/_base_interpreter.py.mako
@@ -15,7 +15,7 @@ from __future__ import annotations
 import abc
 import numpy
 from nitypes.waveform import AnalogWaveform, DigitalWaveform
-from typing import Sequence
+from typing import Any, Sequence
 from nidaqmx.constants import WaveformAttributeMode
 
 
@@ -123,7 +123,7 @@ class BaseInterpreter(abc.ABC):
     def write_analog_waveform(
         self,
         task_handle: object,
-        waveform: AnalogWaveform[numpy.float64],
+        waveform: AnalogWaveform[Any],
         auto_start: bool,
         timeout: float
     ) -> int:
--- a/src/codegen/templates/_grpc_interpreter.py.mako
+++ b/src/codegen/templates/_grpc_interpreter.py.mako
@@ -299,11 +299,11 @@ class GrpcStubInterpreter(BaseInterpreter):
         waveform_attribute_mode: WaveformAttributeMode,
     ) -> Sequence[DigitalWaveform[numpy.uint8]]:
         raise NotImplementedError
-        
+
     def write_analog_waveform(
         self,
         task_handle: object,
-        waveform: AnalogWaveform[numpy.float64],
+        waveform: AnalogWaveform[typing.Any],
         auto_start: bool,
         timeout: float
     ) -> int:


generated/nidaqmx/_base_interpreter.py
```diff
@@ -1909,3 +1909,13 @@ def read_new_digital_waveforms(
...
         raise NotImplementedError
+
+    @abc.abstractmethod
+    def write_analog_waveform(
+        self,
+        task_handle: object,
+        waveform: AnalogWaveform[numpy.float64],
+        auto_start: bool,
+        timeout: float
+    ) -> int:
```
bkeryan: Single channel doesn't need a waveform-specific C entry point. DAQmxWriteAnalogF64 works fine for this use case. See AnalogSingleChannelWriterImpl.cpp.

That said, if you want to implement this in the interpreter in order to set the stage for gRPC support, that's fine. You can change the single-channel implementation to use write_analog_f64.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6967,6 +6967,79 @@ def read_raw(self, task, num_samps_per_chan, timeout, read_array):
...
+    def write_analog_waveform(
+        self,
+        task_handle: object,
+        waveform: AnalogWaveform[numpy.float64],
+        auto_start: bool,
+        timeout: float
+    ) -> int:
+        """Write an analog waveform."""
+        if waveform.raw_data.dtype != numpy.float64:
+            raise TypeError(f"waveform.raw_data must have dtype numpy.float64, got {waveform.raw_data.dtype}")
```
bkeryan: Waveform writes should accept a generic waveform and use scaled_data. See AnalogSingleChannelWriterImpl.cpp.

If this fails because the computed scaled_data is not a contiguous NumPy array, we may need logic to detect this and force a copy. Add a test that uses linear scaling.

Some use cases are to write a waveform that was read from a compressed TDMS file or an NI-SCOPE I16 fetch.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6967,6 +6967,79 @@ def read_raw(self, task, num_samps_per_chan, timeout, read_array):
...
+        int, # error code
+        int, # The number of samples per channel that were written
+    ]:
+        assert isinstance(task_handle, TaskHandle)
+        samps_per_chan_written = ctypes.c_int()
+
+        channel_count = len(write_arrays)
+        assert channel_count > 0
+        array_size = write_arrays[0].size
+        assert all(write_array.size == array_size for write_array in write_arrays)
```
bkeryan: I think >= is more appropriate, in case the waveform capacity is larger than the sample count. I don't know if the waveform's use of views avoids this.

generated/nidaqmx/task/_task.py
```diff
@@ -1316,6 +1317,14 @@ def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
                 number_of_samples_per_channel = len(data)
                 element = data[0]
 
+            elif isinstance(data, AnalogWaveform):
```
bkeryan, line 1321: Should single-channel write accept a 1-element list of waveforms? It seems very inconvenient for write() to require using a different data structure when N == 1.

generated/nidaqmx/task/_task.py
```diff
@@ -1392,6 +1405,11 @@ def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
                     auto_start, timeout, FillMode.GROUP_BY_CHANNEL.value, data)
 
         elif write_chan_type == ChannelType.COUNTER_OUTPUT:
+            if isinstance(data, AnalogWaveform):
```
bkeryan: Is it really appropriate to enumerate the types that aren't supported? Because DigitalWaveform and Spectrum aren't supported either...

tests/component/task/test_task_write_waveform_ao.py
```diff
@@ -0,0 +1,126 @@
...
+def test___task___write_waveform_with_wrong_dtype___raises_type_error(
+    ao_single_channel_task: nidaqmx.Task,
+) -> None:
+    wrong_dtype_samples = np.array([1, 2, 3], dtype=np.int32)
+    wrong_waveform = AnalogWaveform.from_array_1d(wrong_dtype_samples)
+
+    with pytest.raises((TypeError, ValueError)) as exc_info:
+        ao_single_channel_task.write(wrong_waveform)
+
+    error_message = str(exc_info.value)
```
bkeryan: Is this any different than exc_info.value.args[0]?

generated/nidaqmx/_library_interpreter.py
bkeryan: This also means you don't have to check the dtype. AnalogWaveform.scaled_data.dtype is always np.float64.

generated/nidaqmx/task/_task.py
mikeprosserni: I'm planning to add support in the next PR for task.write() to accept a list of waveforms. If the user passes a 1-element list of waveforms into task.write(), should we detect that and use the single-channel code path, or should we just use the multi-channel code path?

bkeryan: Using the multi-channel code path is fine.

generated/nidaqmx/task/_task.py
mikeprosserni: I need some kind of guard here so that pyright doesn't complain about `data` not being iterable on lines below this. But instead of checking against AnalogWaveform, I've changed it to this:

```
            elif not isinstance(data, Iterable):
                raise DaqError(
                    'Write failed, because the provided data type is not supported '
                    'for counter output channels.',
                    DAQmxErrors.UNKNOWN, task_name=self.name)
```

tests/component/task/test_task_write_waveform_ao.py
mikeprosserni: They are equivalent. I've changed all usages of `str(exc_info.value)` to `exc_info.value.args[0]` for consistency.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: I changed this to `assert all(write_array.size >= num_samps_per_chan for write_array in write_arrays)`

generated/nidaqmx/_base_interpreter.py
mikeprosserni: I've removed `_internal_write_analog_waveform_per_chan()` (It will be back in the next PR, of course). `write_analog_waveform()` is now just a thin wrapper for `write_analog_f64()`.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: I've changed `write_analog_waveform()` to use `scaled_data` instead of `raw_data`. As far as I can tell, `scaled_data` is always contiguous, even when `raw_data` isn't. I've added tests that use non-contiguous raw_data, and tests that use linear scaling.

tests/component/_analog_utils.py
```diff
@@ -55,16 +55,37 @@ def _get_voltage_code_offset_for_chan(chan_index: int) -> int:
...
 
-def _create_constant_waveform(num_samples: int) -> AnalogWaveform:
-    samples = np.full(num_samples, 1, dtype=np.float64)
-    return AnalogWaveform.from_array_1d(samples)
+def _create_scaled_linear_ramp_waveform(
+    num_samples: int, start_val: float, end_val: float, gain: float, offset: float
+) -> AnalogWaveform:
+    samples = np.linspace(start_val, end_val, num_samples, dtype=np.float64)
+    return AnalogWaveform.from_array_1d(
+        samples, scale_mode=LinearScaleMode(gain=gain, offset=offset)
```
bkeryan: FYI: Waveform scaling is usually used with an integer dtype, like np.int16. But I guess testing with np.float32 would also cover the raw dtype mismatch.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -6970,74 +6970,20 @@ def read_raw(self, task, num_samps_per_chan, timeout, read_array):
     def write_analog_waveform(
         self,
         task_handle: object,
-        waveform: AnalogWaveform[numpy.float64],
+        waveform: AnalogWaveform,
```
bkeryan: Use `AnalogWaveform[Any]`.

mypy warns about missing typevars in strict mode, but this codebase isn't ready for that.

tests/component/stream_writers/test_analog_single_channel_writer.py
```diff
@@ -81,30 +79,73 @@ def test___analog_single_channel_writer___write_waveform___updates_output(
...
+@pytest.mark.grpc_skip(reason="write_analog_waveform not implemented in GRPC")
+def test___analog_single_channel_writer___write_waveform_with_float32_dtype___updates_output(
+    ao_single_channel_task: nidaqmx.Task,
+    ai_single_channel_loopback_task: nidaqmx.Task,
+) -> None:
+    writer = AnalogSingleChannelWriter(ao_single_channel_task.out_stream)
+    num_samples = 20
+    start_value = 0.0
+    end_value = 1.0
+    waveform = _create_linear_ramp_waveform(num_samples, start_value, end_value)
```
bkeryan: I don't see how this is using float32. Did you intend to add a dtype=np.float32 to the create waveform?

tests/component/stream_writers/test_analog_single_channel_writer.py
```diff
@@ -81,30 +79,73 @@ def test___analog_single_channel_writer___write_waveform___updates_output(
...
 
-    assert "float64" in exc_info.value.args[0]
+    assert samples_written == num_samples
+    assert ai_single_channel_loopback_task.read() == pytest.approx(
+        end_value * gain + offset, abs=AO_VOLTAGE_EPSILON
+    )
+
+
+@pytest.mark.grpc_skip(reason="write_analog_waveform not implemented in GRPC")
+def test___analog_single_channel_writer___write_waveform_with_non_contiguous_data___updates_output(
```
bkeryan, line 133: This shows that AnalogWaveform.scaled_data is making an unnecessary copy. I expected it to fail unless you update nidaqmx to force a copy when the array is non-contigous, because numpy.ctypeslib.ndpointer validates the array flags and raises TypeError.

https://github.com/ni/nitypes-python/issues/197

I would expect nidaqmx to make a deliberate copy only when needed:
```python
array = waveform.scaled_data
if not array.flags.c_contiguous:
    array = array.copy(order="C")
```

tests/component/_analog_utils.py
mikeprosserni: I've changed this to `_create_scaled_int32_ramp_waveform` so we can test with a more usual scaling situation.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: I've changed this to `AnalogWaveform[Any]`

tests/component/stream_writers/test_analog_single_channel_writer.py
mikeprosserni: yeah, I originally wrote this test with dtype=np.float32 and then I accidently messed it up while I was cleaning up the test utilities. I've now fixed it, using `_create_float32_ramp_waveform()`

tests/component/stream_writers/test_analog_single_channel_writer.py
mikeprosserni: I've added those lines to `write_analog_waveform()`, so it should keep working once the bug is fixed.

src/handwritten/stream_writers/_analog_single_channel_writer.py
```diff
@@ -80,3 +86,45 @@ def write_one_sample(self, data, timeout=10):
         return self._interpreter.write_analog_scalar_f64(
             self._handle, auto_start, timeout, data)
 
+    @requires_feature(WAVEFORM_SUPPORT)
+    def write_waveform(
+        self, 
+        waveform: AnalogWaveform[numpy.float64],
```
bkeryan: The public methods should also accept AnalogWaveform[Any].

src/handwritten/task/_task.py
```diff
@@ -1356,10 +1366,14 @@ def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
...
         if write_chan_type == ChannelType.ANALOG_OUTPUT:
-            data = numpy.asarray(data, dtype=numpy.float64)
-            return self._interpreter.write_analog_f64(
-                self._handle, number_of_samples_per_channel, auto_start,
-                timeout, FillMode.GROUP_BY_CHANNEL.value, data)
+            if isinstance(data, AnalogWaveform):
+                return self._interpreter.write_analog_waveform(
+                    self._handle, data, auto_start, timeout)
+            else:
+                data = numpy.asarray(data, dtype=numpy.float64)
```
bkeryan, line 1373: FYI, numpy.asarray has copy=None, so it only copies when needed. Kind of weird that astype doesn't do that.

src/handwritten/stream_writers/_analog_single_channel_writer.py
mikeprosserni: I've updated to `AnalogWaveform[Any]` for all the public write methods

PR #834, opened 2025-09-15
Add AnalogMultiChannelWriter.write_waveforms

mikeprosserni:
- [ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).

- [ ] I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.

- [ ] I've added tests applicable for this pull request

TODO: Include high-level description of the changes in this pull request.

TODO: Justify why this contribution should be part of the project.


zhindes (approved):

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -802,19 +802,104 @@ class LibraryInterpreter(BaseInterpreter):
...
+        auto_start: bool,
+        timeout: float
+    ) -> int:
+        """Write analog waveforms."""
+        assert len(waveforms) > 0
+        num_samps_per_chan = waveforms[0].sample_count
+
+        write_arrays = []
+        for waveform in waveforms:
+            if not waveform.sample_count == num_samps_per_chan:
```
bkeryan:
```suggestion
            if waveform.sample_count != num_samps_per_chan:
```

src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -802,19 +802,104 @@ class LibraryInterpreter(BaseInterpreter):
...
+                raise DaqError(
+                    "The waveforms must all have the same sample count.",
+                     DAQmxErrors.UNKNOWN
+                )
+            write_array = waveform.scaled_data
+            if not write_array.flags.c_contiguous:
+                write_array = write_array.copy(order="C")
+            write_arrays.append(write_array)
```
bkeryan: list.append() is amortized O(1), but you can do better by preallocating with the correct size. 

One way is to initialize write_arrays to `[waveform.scaled_data for waveform in waveforms]` and then do a second pass and reassign any non-contiguous arrays.

Another way is to push the `waveform.scaled_data ... copy(order="C")` logic into a function so you can call it in a list comprehension.

src/handwritten/stream_writers/_analog_multi_channel_writer.py
```diff
@@ -84,3 +90,48 @@ def write_one_sample(self, data, timeout=10):
         return self._interpreter.write_analog_f64(
             self._handle, 1, auto_start, timeout, FillMode.GROUP_BY_CHANNEL.value, data
         )
+
+    @requires_feature(WAVEFORM_SUPPORT)
+    def write_waveforms(
+        self, waveforms: Sequence[AnalogWaveform[Any]], timeout: float = 10.0
+    ) -> int:
+        """Writes one or more waveforms to one or more analog output channels in a task.
```
bkeryan: Unlike write_many_sample, both "one or more" are describing the same quantity, so consider simplifying

```suggestion
        """Writes waveforms to one or more analog output channels in a task.
```

generated/nidaqmx/task/_task.py
```diff
@@ -1295,6 +1295,15 @@ def _raise_invalid_write_num_chans_error(self, number_of_channels, number_of_cha
...
+    def _raise_invalid_write_mixed_data_error(self):
+        raise DaqError(
+            "Write cannot be performed, because the list contains a mix of "
+            "AnalogWaveform and non-AnalogWaveform objects. All elements in "
+            "the list must be the same type.",
+            DAQmxErrors.UNKNOWN,
+            task_name=self.name,
+        )
+
     def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
```
bkeryan, line 1326: I wonder if Task should also provide write_waveforms, perhaps as an alias for write.

Currently, Task is asymmetrical: read/read_waveform vs. write
Also, Task is inconsistent with writers about whether there is a write_waveform(s)

bkeryan: On the one hand, [There should be one-- and preferably only one --obvious way to do it.](https://peps.python.org/pep-0020/)

On the other hand, intellisense users might incorrectly conclude that task can't write waveforms.

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: fixed

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: I added a `_get_write_array()` helper so I could do `write_arrays = [self._get_write_array(waveform) for waveform in waveforms]`

src/handwritten/stream_writers/_analog_multi_channel_writer.py
mikeprosserni: fixed

generated/nidaqmx/task/_task.py
mikeprosserni: I added a `write_waveforms()` on task that just calls `write()`. Take a look at that and let me know what you think.

We can say that `write_waveforms()` is the one (but perhaps not only) "obvious way to do it". We could even go further, and remove the documentation on `write()` that mentions waveforms, if we wanted to further emphasize `write_waveforms()`.

One other advantage of `write_waveforms()` is that it can have waveform-exclusive type annotations.

generated/nidaqmx/_library_interpreter.py
```diff
@@ -7072,6 +7065,11 @@ def _internal_write_analog_waveform_per_chan(
 
         return error_code, samps_per_chan_written.value
 
+    def _get_write_array(self, waveform: AnalogWaveform[Any]) -> numpy.typing.NDArray[numpy.float64]:  
+        if waveform.scaled_data.flags.c_contiguous:
+            return waveform.scaled_data
+        return waveform.scaled_data.copy(order="C")
```
bkeryan: Query scaled_data only once. If the waveform is scaled, querying this property scales the data, making a copy in the process.

src/handwritten/task/_task.py
```diff
@@ -1588,6 +1589,59 @@ def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
                 task_name=self.name,
             )
 
+    def write_waveforms(
```
bkeryan: Task.write_waveforms is inconsistent with Task.read_waveform (singular vs. plural)

generated/nidaqmx/task/_task.py
bkeryan: Does it make sense to remove waveform support from write() and move it to write_waveforms?

Or move it to write_waveforms and have write() forward to write_wvaeforms() when it sees a waveform?

mikeprosserni: I moved the waveform support out of `write()` and into `write_waveforms()`, to see how it looked, and yeah, I think that makes sense. 

I still have `write()` forward to `write_waveforms()` though, because `write()` bills itself as accepting any and all types that can be written.

generated/nidaqmx/_library_interpreter.py
mikeprosserni: updated to query `scaled_data` only once.

src/handwritten/task/_task.py
mikeprosserni: renamed to singular

src/handwritten/task/_task.py
```diff
@@ -1573,20 +1572,98 @@ def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
...
+
+        Returns:
+            int:
+
+            Specifies the actual number of samples this method
+            successfully wrote.
+        """
+        channels_to_write = self.channels
+        number_of_channels = len(channels_to_write.channel_names)
+        write_chan_type = channels_to_write.chan_type
```
bkeryan, line 1634: FYI, this asymmetry is correct. Input uses the `task.in_stream.channels_to_read` property, but output always writes to all channels in the task.

src/handwritten/task/_task.py
```diff
@@ -1573,20 +1572,98 @@ def write(self, data, auto_start=AUTO_START_UNSET, timeout=10.0):
...
+
+        if isinstance(waveforms, AnalogWaveform):
+            number_of_samples_per_channel = waveforms.sample_count
+        elif isinstance(waveforms, list):
+            number_of_samples_per_channel = len(waveforms)
```
bkeryan: This looks wrong. len(waveforms) is the number of channels.

src/handwritten/task/_task.py
```diff
@@ -1267,7 +1268,7 @@ def wait_until_done(self, timeout=10.0):
         """
         self._interpreter.wait_until_task_done(self._handle, timeout)
 
-    def _raise_invalid_num_lines_error(self, num_lines_expected, num_lines_in_data):
+    def _raise_invalid_num_lines_error(self, num_lines_expected, num_lines_in_data) -> NoReturn:
```
bkeryan, line 1271: good

src/handwritten/task/_task.py
mikeprosserni: oops, changed to `number_of_samples_per_channel = waveforms[0].sample_count`

bkeryan: FYI, that raises IndexError if the list is empty, but I guess that's ok

bkeryan (approved):

PR #838, opened 2025-09-19
Users/mprosser/task 3271835 digital multiple waveform writer

mikeprosserni:

- Add LibraryInterpreter.write_digital_waveforms()
- Add DigitalMultiChannelWriter.write_waveforms()

AB#3271835

### DIFF
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -7111,7 +7111,14 @@ class LibraryInterpreter(BaseInterpreter):
         channel_count = len(waveforms)
         assert channel_count > 0
         sample_count = waveforms[0].sample_count
-        assert all(wf.sample_count == sample_count for wf in waveforms)
+
+        for waveform in waveforms:
+            if waveform.sample_count != sample_count:
+                raise DaqError(
+                    "The waveforms must all have the same sample count.",
+                    DAQmxErrors.UNKNOWN
+                )
+                
         bytes_per_chan_array = numpy.array([wf.signal_count for wf in waveforms], dtype=numpy.uint32)
 
         # build a temporary contiguous array to write the data from multiple channels into.
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -7103,22 +7103,34 @@ class LibraryInterpreter(BaseInterpreter):
     def write_digital_waveforms(
         self,
         task_handle: object,
-        waveform: Sequence[DigitalWaveform[Any]],
+        waveforms: Sequence[DigitalWaveform[Any]],
         auto_start: bool,
         timeout: float,
     ) -> int:
         """Write digital waveforms."""
-        # TODO: Support multiple waveforms
+        channel_count = len(waveforms)
+        assert channel_count > 0
+        sample_count = waveforms[0].sample_count
+        assert all(wf.sample_count == sample_count for wf in waveforms)
+        bytes_per_chan_array = numpy.array([wf.signal_count for wf in waveforms], dtype=numpy.uint32)
 
-        bytes_per_chan_array = numpy.array([waveform[0].signal_count], dtype=numpy.uint32)
+        # build a temporary contiguous array to write the data from multiple channels into.
+        # write_array must be in the format (numChans x numSampsPerChan x maxDataWidth)
+        write_array = numpy.zeros(
+            (channel_count, sample_count, max(bytes_per_chan_array)),
+            dtype=numpy.uint8,
+        )
+        for i, wf in enumerate(waveforms):
+            signal_count = wf.signal_count
+            write_array[i, :, :signal_count] = self._get_digital_write_array(wf)
 
         error_code, samples_written = self._internal_write_digital_waveform(
             task_handle,
-            waveform[0].sample_count,
+            sample_count,
             auto_start,
             timeout,
             FillMode.GROUP_BY_CHANNEL.value,
-            self._get_digital_write_array(waveform[0]),
+            write_array,
             bytes_per_chan_array,
         )
 
--- a/generated/nidaqmx/_library_interpreter.py
+++ b/generated/nidaqmx/_library_interpreter.py
@@ -7100,6 +7100,31 @@ class LibraryInterpreter(BaseInterpreter):
         self.check_for_error(error_code, samps_per_chan_written=samples_written)
         return samples_written
 
+    def write_digital_waveforms(
+        self,
+        task_handle: object,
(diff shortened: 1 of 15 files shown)


generated/nidaqmx/_library_interpreter.py
```diff
@@ -7100,6 +7100,50 @@ def write_digital_waveform(
...
+
+        # build a temporary contiguous array to write the data from multiple channels into.
+        # write_array must be in the format (numChans x numSampsPerChan x maxDataWidth)
+        write_array = numpy.zeros(
+            (channel_count, sample_count, max(bytes_per_chan_array)),
+            dtype=numpy.uint8,
+        )
+        for i, wf in enumerate(waveforms):
+            signal_count = wf.signal_count
+            write_array[i, :, :signal_count] = self._get_digital_write_array(wf)
```
bkeryan: _get_digital_write_array makes a copy if the waveform array is not contiguous, but in this case that isn't necessary, because you're copying the waveform array into the flat array rather than passing it to DAQmx.

tests/component/_digital_utils.py
```diff
@@ -133,19 +150,56 @@ def _get_waveform_data_msb(waveform: DigitalWaveform) -> list[int]:
...
+def _create_waveform_for_line(num_samples: int, line_number: int) -> DigitalWaveform:
+    waveform = DigitalWaveform(num_samples, 1)
+    expected_data = _get_expected_data_for_line(num_samples, line_number)
+    _set_waveform_data(num_samples, 1, waveform, expected_data)
+    return waveform
 
+
+def _create_waveform_for_lines(
+    num_samples: int, first_line: int, num_lines: int, dtype: type[numpy.generic]
+) -> DigitalWaveform:
```
bkeryan: `DigitalWaveform` has an implicit `Any`, use `DigitalWaveform[Any]` or `DigitalWaveform[np.bool]` instead.

This only applies to type hints, not constructor calls.

tests/component/_digital_utils.py
```diff
@@ -133,19 +150,56 @@ def _get_waveform_data_msb(waveform: DigitalWaveform) -> list[int]:
...
+    # create waveforms for lines 2-4, 0-1, and 5-7, matching the channel configuration
+    # in the do_multi_channel_mixed_line_task fixture
+    return [
+        _create_waveform_for_lines(num_samples, first_line=2, num_lines=3, dtype=numpy.uint8),
+        _create_waveform_for_lines(num_samples, first_line=0, num_lines=2, dtype=numpy.int8),
+        _create_waveform_for_lines(num_samples, first_line=5, num_lines=3, dtype=numpy.bool),
+    ]
+
+
+def _set_waveform_data(num_samples, num_lines, waveform, expected_data, invert=False):
```
bkeryan: This is missing type hints.

bkeryan (changes requested):

zhindes (approved):

generated/nidaqmx/_library_interpreter.py
mikeprosserni: I updated this to use `waveform.data` directly instead of `_get_digital_write_array()`

tests/component/_digital_utils.py
mikeprosserni: I updated all the DigitalWaveform type hints that were missing `[]`s

tests/component/_digital_utils.py
mikeprosserni: added type hints

generated/nidaqmx/_library_interpreter.py
```diff
@@ -7127,9 +7135,9 @@ def write_digital_waveforms(
             (channel_count, sample_count, max(bytes_per_chan_array)),
             dtype=numpy.uint8,
         )
-        for i, wf in enumerate(waveforms):
-            signal_count = wf.signal_count
-            write_array[i, :, :signal_count] = self._get_digital_write_array(wf)
+        for i, waveform in enumerate(waveforms):
+            signal_count = waveform.signal_count
+            write_array[i, :, :signal_count] = waveform.data
```
bkeryan, line 7140: FYI: good, this seems to allow mixing dtypes

```
>>> import numpy as np
>>> x=np.array([0,1,0],np.bool)
>>> x
array([False,  True, False])
>>> y=np.array([0,0,0,0,0],np.uint8)
>>> y[1:4]=x
>>> y
array([0, 0, 1, 0, 0], dtype=uint8)
```

tests/component/_digital_utils.py
```diff
@@ -171,14 +171,14 @@ def _create_waveform_for_line(num_samples: int, line_number: int) -> DigitalWave
 
 def _create_waveform_for_lines(
     num_samples: int, first_line: int, num_lines: int, dtype: type[numpy.generic]
-) -> DigitalWaveform:
+) -> DigitalWaveform[Any]:
```
bkeryan: Functions that take a dtype (this and _create_digital_waveform) should use a typevar to propagate it to the return type.
```python
def _create_waveform_for_lines(
    num_samples: int, first_line: int, num_lines: int, dtype: type[_D]
) -> DigitalWaveform[_D]:
```

(If _D doesn't work, try nitypes.waveform.typing.TDigitalState.)

tests/component/_digital_utils.py
```diff
@@ -188,7 +188,13 @@ def _create_waveforms_for_mixed_lines(num_samples: int) -> list[DigitalWaveform]
...
 
 
-def _set_waveform_data(num_samples, num_lines, waveform, expected_data, invert=False):
+def _set_waveform_data(
+    num_samples: int,
+    num_lines: int,
+    waveform: DigitalWaveform[Any],
+    expected_data: list[int],
+    invert: bool = False,
+):
```
bkeryan:
```suggestion
) -> None:
```

tests/component/_digital_utils.py
bkeryan: Don't forget the return. If you don't use `-> None`, then it returns `-> Any`.

bkeryan (approved):
Approved with suggestions

tests/component/_digital_utils.py
mikeprosserni: I've added TDigitalState annotations, although that makes the default value for the dtype tricky, so I added a more strictly typed `_create_digital_waveform_uint8()` for most callers to use.

tests/component/_digital_utils.py
mikeprosserni: done

tests/component/_digital_utils.py
mikeprosserni: added -> None

Issue #639, opened 2024-10-03
first_samp_timestamp_val property does not work because LibraryInterpreter is missing a method


bkeryan:
The `nidaqmx.Timing.first_samp_timestamp_val` property does not work because `LibraryInterpreter` is missing the `get_timing_attribute_timestamp` method.

```
>>> t = nidaqmx.Task()
>>> t.timing.first_samp_timestamp_val
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\dev\nidaqmx-python\generated\nidaqmx\task\_timing.py", line 424, in first_samp_timestamp_val
    val = self._interpreter.get_timing_attribute_timestamp(self._handle, 0x313a)
AttributeError: 'LibraryInterpreter' object has no attribute 'get_timing_attribute_timestamp'
```

bkeryan:
@zhindes @WayneDroid We should have added a test for this in https://github.com/ni/nidaqmx-python/pull/592 . Is there any reason we didn't?

Also, I think mypy would have caught this if the constructor arguments in https://github.com/ni/nidaqmx-python/blob/c09f2e286b8486272086d96f6a9846b20fa14520/generated/nidaqmx/task/_timing.py had type hints. We should try adding them, and if that enables mypy to catch errors like this, we should add these type hints to all of the subobject classes.

WayneDroid:
> @zhindes @WayneDroid We should have added a test for this in #592 . Is there any reason we didn't?
> 
> Also, I think mypy would have caught this if the constructor arguments in https://github.com/ni/nidaqmx-python/blob/c09f2e286b8486272086d96f6a9846b20fa14520/generated/nidaqmx/task/_timing.py had type hints. We should try adding them, and if that enables mypy to catch errors like this, we should add these type hints to all of the subobject classes.

I can't recall why no test case added for those properties, maybe overlooked.

PR #857, opened 2025-10-13
Fix first_samp_timestamp_val property does not work due to missing method in LibraryInterpreter

lau-yeexuan:

Exclude _GetTimingAttributeTimestamp_ from interpreter ignored functions. The nidaqmx.Timing.first_samp_timestamp_val now works because LibraryInterpreter contains the get_timing_attribute_timestamp method.

Fixes https://github.com/ni/nidaqmx-python/issues/639
[AB#3064756](https://dev.azure.com/ni/DevCentral/_workitems/edit/3064756)

Tests to retrieve first_samp_timestamp_val of simulated 9205.

### DIFF
--- a/src/codegen/utilities/interpreter_helpers.py
+++ b/src/codegen/utilities/interpreter_helpers.py
@@ -43,8 +43,8 @@ INTERPRETER_IGNORED_FUNCTIONS = [
     "SetRealTimeAttributeInt32",
     "SetRealTimeAttributeUInt32",
     "WaitForNextSampleClock",
-    # Time triggers 
-    # Single-attribute get/set functions are not used 
+    # Time triggers
+    # Single-attribute get/set functions are not used
     # Generic Get/SetTimingAttribute{Type} functions are used instead
     "GetArmStartTrigTimestampVal",
     "GetArmStartTrigTrigWhen",


.gitignore
```diff
@@ -18,4 +18,5 @@ dist/
 docs/_build/
 
 # Common editor metadata
-.vscode/
\ No newline at end of file
+.vscode/
+.vs/
```
zhindes: Thanks for this, but do you mind making a separate PR for this?

src/codegen/utilities/interpreter_helpers.py
```diff
@@ -53,7 +53,6 @@
     "GetStartTrigTrigWhen",
     "GetSyncPulseTimeWhen",
     "GetTimingAttributeExTimestamp",
```
zhindes, line 57: There are a lot of Time Trigger functions being removed at the moment... what makes the timestamps more important than the rest? Aren't all of these worth making public?

tests/component/task/test_timing_properties.py
```diff
@@ -1,3 +1,5 @@
+from datetime import datetime as std_datime
```
bkeryan: spelling: std_datetime

src/codegen/utilities/interpreter_helpers.py
bkeryan: We don't use single-attribute get/set functions in this API. We only use generic get/set functions like `GetTimingAttribute{Type}`.

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ All notable changes to this project will be documented in this file.
     * Adopt ni-python-styleguide for hand-written code.
     * Upgrade to Poetry 2.x and migrate `pyproject.toml` format.
     * Add support for Python 3.14.
+    * Add support and tests for GetTimingAttributeTimestamp
```
bkeryan: This is a bugfix so it goes under "resolved issues"

src/codegen/utilities/interpreter_helpers.py
zhindes: Ah hah! Ok, so just an oversight after we (I?) enabled the time types. Gotcha!

zhindes: Actually, I'm unresolving this. Can you update the comment above this, Yee Xuan? We do support Time, now, but we don't need these overly-specific entrypoints.

.gitignore
lau-yeexuan: Drop the change for this.

src/codegen/utilities/interpreter_helpers.py
lau-yeexuan: Updated the comment

CHANGELOG.md
lau-yeexuan: Done modified.

PR #859, opened 2025-10-20
Add ni-apis as a submodule

mikeprosserni:

This adds https://github.com/ni/ni-apis as a submodule of nidaqmx-python. (similar to https://github.com/ni/grpc-device/pull/1201)

This allows us to access the canonical versions of proto files like session.proto. I'll also be wanting to access waveform.proto in the future, which is the main reason I'm doing this PR.

[AB#3439518](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3439518)

CI Checks Passed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ documentation = "https://nidaqmx-python.readthedocs.io"
 nidaqmx = 'nidaqmx.__main__:main'
 
 [project.optional-dependencies]
-grpc = ['grpcio (>=1.49.0,<2.0)', 'protobuf (>=4.21)']
+grpc = ['grpcio (>=1.49.0,<2.0)', 'protobuf (>=4.21)', 'ni-grpcdevice-v1-proto (>=1.0.0)']
 
 [tool.poetry]
 
@@ -64,7 +64,6 @@ distro = { version = ">=1.9.0", platform = "linux" }
 requests = ">=2.25.0"
 typing_extensions = { version = ">=4.0.0" }
 nitypes = {version=">=0.1.0dev10", allow-prereleases=true}
-ni-grpcdevice-v1-proto = { version = "^1.0.0" }
 
 [tool.poetry.group.codegen.dependencies]
 Mako = "^1.2"
--- a/poetry.lock
+++ b/poetry.lock
@@ -908,6 +908,7 @@ files = [
     {file = "grpcio-1.75.1-cp39-cp39-win_amd64.whl", hash = "sha256:9fe51e4a1f896ea84ac750900eae34d9e9b896b5b1e4a30b02dc31ad29f36383"},
     {file = "grpcio-1.75.1.tar.gz", hash = "sha256:3e81d89ece99b9ace23a6916880baca613c03a799925afb2857887efa8b1b3d2"},
 ]
+markers = {main = "extra == \"grpc\""}
 
 [package.dependencies]
 typing-extensions = ">=4.12,<5.0"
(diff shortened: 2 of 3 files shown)


generated/nidaqmx/_stubs/session_pb2.pyi
bkeryan, line 1: You can't update this because it will break compatibility with nimi-python and measurement-plugin-python. See https://github.com/ni/ni-apis-python/tree/main/proto

Here is the plan:
https://github.com/ni/ni-apis-python/issues/75

You can either keep this file the same for now, or you can delete it and add a dependency on ni.grpcdevice.v1.proto.

third_party/ni-apis
bkeryan, line 1: Move to third_party/ni-apis/

pyproject.toml
```diff
@@ -129,10 +129,10 @@ numpy = [
 
 [tool.black]
 line-length = 100
-extend_exclude = ".tox/|docs/|generated/|src/codegen/metadata/|src/codegen/templates/"
+extend_exclude = ".tox/|docs/|generated/|src/codegen/metadata/|src/codegen/templates/|ni-apis/"
```
bkeryan: Exclude third_party/

.github/workflows/build.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - name: Initialize ni-apis submodule
+        run: git submodule update --init ni-apis
```
bkeryan: The reasons why grpc-device uses a separate git command don't apply here, so let actions/checkout handle this.
```suggestion
      - name: Check out repo
        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
        with:
          submodules: true
```

third_party/ni-apis
mikeprosserni: moved submodule to third_party

pyproject.toml
mikeprosserni: updated

.github/workflows/build.yml
mikeprosserni: updated

generated/nidaqmx/_stubs/session_pb2.pyi
mikeprosserni: I've deleted this (stub_generator no longer generates files for session.proto), and added a dependency on ni-grpcdevice-v1-proto instead. 

However, to do that, I did have to add a `remove_session_files()` step in stub_generator.py (because nidaqmx.proto contains `import "session.proto"`). Is that ok?

bkeryan: > However, to do that, I did have to add a remove_session_files() step in stub_generator.py (because nidaqmx.proto contains import "session.proto"). Is that ok?

I don't think you need this. protoc only generates stubs for files that you list on the command line (the `proto_files` variable), not files that you import. It generates references (imports) to stubs for files that you import.

I commented out the call to remove_session_files() and it doesn't create session_pb2.py.

pyproject.toml
```diff
@@ -64,6 +64,7 @@ distro = { version = ">=1.9.0", platform = "linux" }
 requests = ">=2.25.0"
 typing_extensions = { version = ">=4.0.0" }
 nitypes = {version=">=0.1.0dev10", allow-prereleases=true}
+ni-grpcdevice-v1-proto = { version = "^1.0.0" }
```
bkeryan: This is pulling in gRPC dependencies with Python interpreters for which there are no gRPC binary wheels, which is causing the tests to spend 45 minutes trying to rebuild `grpcio` and then fail:

```
PEP517 build of a dependency failed

Backend subprocess exited when trying to invoke build_wheel

    | Command '['C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp0z4tw_m7\\.venv\\Scripts\\python.exe', 'D:\\a\\_temp\\poetry\\home\\Lib\\site-packages\\pyproject_hooks\\_in_process\\_in_process.py', 'build_wheel', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpd8d0vqt6']' returned non-zero exit status 1.
    | 
    | Found cython-generated files...
...
```

Please add new gRPC dependencies to the `grpc` extra, not the main package:

```
[project.optional-dependencies]
grpc = ['grpcio (>=1.49.0,<2.0)', 'protobuf (>=4.21)']
```

generated/nidaqmx/_stubs/session_pb2.pyi
mikeprosserni: Ah, I must have gotten confused by some stale files or something. I've removed remove_session_files.

pyproject.toml
mikeprosserni: moved

PR #870, opened 2025-10-24
nidaqmx: Remove support for Python 3.9

bkeryan:

Remove support for Python 3.9

NI's Python support policy says:

> NI products that support or integrate with Python will drop official support for all Python versions that have reached their end of life [EOL](https://devguide.python.org/devcycle/#end-of-life-branches) . This should be done at the next release of the product following the EOL date for a given Python version

PR build

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,25 +2,11 @@
 
 [[package]]
 name = "alabaster"
-version = "0.7.16"
-description = "A light, configurable Sphinx theme"
-optional = false
-python-versions = ">=3.9"
-groups = ["docs"]
-markers = "python_version == \"3.9\""
-files = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
bkeryan, line 1: @hongloonwong @maxxboehme @zhindes For internal test suites that specify `python = "^3.8"` or `python = "^3.9"`, Poetry will continue using an older version of `nidaqmx`. Is updating these on the Venus team's radar?

zhindes: FYI Maxx & Hong Loon - keeping up with Python versions is more than just an internal package concern... keeping our ATSes using supported versions of Python is a good idea generally. If for no other reason than any random packages we use (not just ours) might drop support like we are doing here. Being proactive can prevent downtime.

bkeryan: I created an AzDO PR updating the one test suite that I'm certain needs to be updated.

PR #874, opened 2025-10-28
Create a synchronization example for TDMS logging

### LINKED ISSUE #820: Create a synchronization example for TDMS logging
The examples in synchronisation (required for multiple task types eg. analogue in and counter in) do not show how to log all tasks to a single TDMS file. It would be useful to have all synchronised tasks in one place as the NI hardware supports multiple tasks.

Doing something like this raises the error "File specified is already opened for output. NI-DAQmx requires exclusive write access. 

`ai_task.in_stream.configure_logging(
        filepath, 
        LoggingMode.LOG,
        operation=LoggingOperation.CREATE_OR_REPLACE
        )
    ci1_task.in_stream.configure_logging(
        filepath, 
        LoggingMode.LOG,
        operation=LoggingOperation.OPEN
        )`

[AB#3252273](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3252273)

lau-yeexuan:

Create an example of logging multiple synchronized tasks to a single TDMS file using a queue.

Fixes #820 
[AB#3252273](https://dev.azure.com/ni/DevCentral/_workitems/edit/3252273)

multi_task_data.tdms generated 

image

image

### DIFF
--- a/examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
+++ b/examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
@@ -12,7 +12,6 @@ import threading
 import time
 from collections.abc import Sequence
 from concurrent.futures import ThreadPoolExecutor
-from datetime import datetime
 from pathlib import Path
 
 import numpy as np
@@ -133,11 +132,9 @@ def consumer(
                         "wf_increment": waveform.timing.sample_interval.total_seconds(),
                         "wf_samples": len(waveform.raw_data),
                         "wf_start_offset": 0.0,
-                        "wf_start_time": (
-                            waveform.timing.start_time
-                            if isinstance(waveform.timing.start_time, datetime)
-                            else datetime.now()
-                        ).strftime("%Y-%m-%d %H:%M:%S"),
+                        "wf_start_time": waveform.timing.to_datetime().start_time.strftime(
+                            "%Y-%m-%d %H:%M:%S"
+                        ),
                     },
                 )
                 objects_to_write.append(channel)
@@ -154,11 +151,9 @@ def consumer(
                     "wf_increment": counter_waveform.timing.sample_interval.total_seconds(),
                     "wf_samples": len(counter_waveform.raw_data),
                     "wf_start_offset": 0.0,
-                    "wf_start_time": (
-                        counter_waveform.timing.start_time
-                        if isinstance(counter_waveform.timing.start_time, datetime)
-                        else datetime.now()
-                    ).strftime("%Y-%m-%d %H:%M:%S"),
+                    "wf_start_time": counter_waveform.timing.to_datetime().start_time.strftime(
+                        "%Y-%m-%d %H:%M:%S"
+                    ),
                 },
             )
             objects_to_write.append(channel)


examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+from nidaqmx.constants import (
+    AcquisitionType,
+    Edge,
+)
+
+# Configuration
+SAMPLE_RATE = 1000
+SAMPLES_PER_CHANNEL = 1000
+TIMEOUT = 10.0
+
```
zhindes, line 45: parameterizing the device might be nice, and then use f-strings with this later

```suggestion
DEVICE_NAME = "Dev1"
```

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+def producer(
+    tasks: List[nidaqmx.Task], data_queue: queue.Queue, stop_event: threading.Event
+) -> None:
+    """Producer function that reads data from DAQmx tasks and puts it in the queue."""
+    try:
+        while not stop_event.is_set():
+            data = []
+            for task in tasks:
+                task_data = task.read(
```
zhindes: by the nature of what this example is doing (multi-task synchronization using queues and threads), it seems performance would be a concern. given that, I would recommend this example use the Readers instead of `task.read`, which has a lot of overhead to figure out how to read. However, the interface for reading analog vs counters is different, naturally, so it would make it a lot more complicated. So, I think its OK to do this, but maybe some comments here explaining how you could improve performance by doing that. Something like this:

> Note: The Task read interface is less performant than stream_readers, but the interface is less generic. You could refactor this code to use the readers to improve performance, if necessary.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+# Configuration
+SAMPLE_RATE = 1000
+SAMPLES_PER_CHANNEL = 1000
+TIMEOUT = 10.0
+
+
+def producer(
+    tasks: List[nidaqmx.Task], data_queue: queue.Queue, stop_event: threading.Event
+) -> None:
+    """Producer function that reads data from DAQmx tasks and puts it in the queue."""
```
zhindes, line 52: I would add here the data the queue expects, it looks like an array of task read data, which is double arrays?

zhindes: Oh, no, the counter data is array of tuples

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+                            for chan_idx, chan_data in enumerate(task_data):
+                                chan_data = np.array(chan_data).flatten()
+                                channel = ChannelObject(
+                                    group_names[task_idx],
+                                    channel_names[task_idx][chan_idx],
+                                    chan_data,
+                                    properties={"Sample Rate": SAMPLE_RATE},
+                                )
+                                objects_to_write.append(channel)
+                        else:
```
zhindes: I would make this explicit; check for the data type you expect and make an else clause that throws an exception

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
zhindes: Oh, this is multi-device. Why does it need to be multi-device? That is more complicated, and you aren't quite doing it right. I would think you make this a single-device example and add a comment for how to make it multiple devices. Maybe add a comment above this device-name, something like this:

> Note: To make this example work across devices you would need to export a Sample Clock from one of the devices to the other.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+    try:
+        clk_task.co_channels.add_co_pulse_chan_freq(
+            counter="Dev1/ctr1",
+            freq=SAMPLE_RATE,
+        )
+        clk_task.timing.cfg_implicit_timing(sample_mode=AcquisitionType.CONTINUOUS)
+
+        ai_task.ai_channels.add_ai_voltage_chan("Dev2/ai0", "Torque01")
+        ai_task.ai_channels.add_ai_voltage_chan("Dev2/ai1", "Torque02")
+        ai_task.timing.cfg_samp_clk_timing(
```
zhindes, line 189: This task isn't actually synchronized to Dev1. Like I put above, I would just make this 1 device. However, you still need to synchronize them. I would do that by changing ci1 task. I'll comment below.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+            SAMPLE_RATE, sample_mode=AcquisitionType.CONTINUOUS, samps_per_chan=SAMPLES_PER_CHANNEL
+        )
+
+        ci1_chan = ci1_task.ci_channels.add_ci_count_edges_chan(
+            "Dev1/ctr0", "Rotations01", edge=Edge.RISING, initial_count=0
+        )
+        ci1_chan.ci_count_edges_term = "/Dev1/PFI0"
+        ci1_task.timing.cfg_samp_clk_timing(
```
zhindes: I would remove clk_task and simply make this is "/Dev1/ai/SampleClock"

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+                tdms_filepath,
+                ["AI_Task", "CI_Task"],
+                [["Torque01", "Torque02"], ["Rotations01"]],
+                stop_event,
+            ),
+        )
+
+        clk_task.start()
+        ci1_task.start()
+        ai_task.start()
```
zhindes: If we remove clk_task and make ci1_task use "ai/SampleClock" make sure ci1 starts before ai. You're already doing that, but its critical. I would add a comment.

> Note: Ensure you start the Analog task last since the Counter task is using its clock

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+    - A counter input task counting edges
+
+    Data is acquired continuously until user presses Enter, then saved to a TDMS file
+    using a producer-consumer pattern with a queue for thread-safe data transfer.
+    """
+    data_queue = queue.Queue(maxsize=10)
+    stop_event = threading.Event()
+
+    ai_task = nidaqmx.Task()
+    ci1_task = nidaqmx.Task()
```
zhindes: why ci1? just ci sounds better?

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+    group_names: List[str],
+    channel_names: List[List[str]],
+    stop_event: threading.Event,
+) -> None:
+    """Consumer function that writes data from the queue to a TDMS file."""
+    try:
+        with TdmsWriter(tdms_path) as tdms_writer:
+            while not stop_event.is_set():
+                try:
+                    data = data_queue.get(timeout=TIMEOUT)
```
bkeryan: This is the only thing that raises queue.Empty, right? If so, move the corresponding except up here to make the function easier to follow.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+
+                    if data is None:
+                        break
```
bkeryan: Why would data be None? If the queue was empty, you would get a queue.Empty exception. Is the producer putting None into the queue?

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+1. Synchronize multiple DAQmx tasks using a shared clock
+2. Use a producer-consumer pattern with a queue
+3. Log data from multiple tasks to a single TDMS file
+"""
+
+import os
+import queue
+import threading
+import time
+from typing import List
```
bkeryan: typing.List has been deprecated since we dropped Python 3.8 support. Use `list[T]`, or even better, use `collections.abc.Sequence[T]`.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+)
+
+# Configuration
+SAMPLE_RATE = 1000
+SAMPLES_PER_CHANNEL = 1000
+TIMEOUT = 10.0
+
+
+def producer(
+    tasks: List[nidaqmx.Task], data_queue: queue.Queue, stop_event: threading.Event
```
bkeryan: Queue is generic: https://github.com/python/typeshed/blob/main/stdlib/queue.pyi

Using it in type hints without a typevar is equivalent to Queue[Any] which disables type checking for queue elements.

What are you putting in it? A collections.abc.Sequence? That is also a generic type.

If you want the queue elements to be dynamically typed, use Queue[object] to indicate that the element type could be anything and that the consumer should inspect its type using isinstance() or whatever.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
bkeryan: IMO, the big problem with task.read() is that it converts the data from an NumPy array to a list[list[float]], which is an extremely expensive conversion, and then the consumer thread converts it back to a NumPy array. 

I don't think adding a comment is sufficient. You should update this to use the reader API.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+                        group = GroupObject(
+                            group_names[task_idx], properties={"Sample Rate": SAMPLE_RATE}
+                        )
+                        objects_to_write.append(group)
+
+                        if isinstance(task_data, (list, tuple)) and isinstance(
+                            task_data[0], (list, tuple, np.ndarray)
+                        ):
+                            for chan_idx, chan_data in enumerate(task_data):
+                                chan_data = np.array(chan_data).flatten()
```
bkeryan: You're already iterating over the channel data. Why would it not be flat? Is this related to tuples for counters?

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+    using a producer-consumer pattern with a queue for thread-safe data transfer.
+    """
+    data_queue = queue.Queue(maxsize=10)
+    stop_event = threading.Event()
+
+    ai_task = nidaqmx.Task()
+    ci1_task = nidaqmx.Task()
+    clk_task = nidaqmx.Task()
+
+    script_dir = os.path.dirname(os.path.abspath(__file__))
```
bkeryan: Use pathlib

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+        ci1_chan.ci_count_edges_term = "/Dev1/PFI0"
+        ci1_task.timing.cfg_samp_clk_timing(
+            SAMPLE_RATE,
+            "/Dev1/Ctr3InternalOutput",
+            sample_mode=AcquisitionType.CONTINUOUS,
+            samps_per_chan=SAMPLES_PER_CHANNEL,
+        )
+
+        producer_thread = threading.Thread(
```
bkeryan: Consider using `with ThreadPoolExecutor() as executor` rather than managing OS threads directly.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+            target=producer, args=([ai_task, ci1_task], data_queue, stop_event)
+        )
+
+        consumer_thread = threading.Thread(
+            target=consumer,
+            args=(
+                data_queue,
+                tdms_filepath,
+                ["AI_Task", "CI_Task"],
+                [["Torque01", "Torque02"], ["Rotations01"]],
```
bkeryan: "torque" is oddly specific considering that the rest of this example says nothing about what it's measuring.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+                                    properties={"Sample Rate": SAMPLE_RATE},
+                                )
+                                objects_to_write.append(channel)
+                        else:
+                            task_data = np.array(task_data).flatten()
+                            channel = ChannelObject(
+                                group_names[task_idx],
+                                channel_names[task_idx][0],
+                                task_data,
+                                properties={"Sample Rate": SAMPLE_RATE},
```
bkeryan: Using a global to pass a hardcoded sample rate to both the producer and consumer works in an example but may not work as well in a real application. 

It would be better to pass the sample rate along with the queued data. Good news: we have a data type for that: https://nitypes.readthedocs.io/en/latest/autoapi/nitypes/waveform/AnalogWaveform.html

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,208 @@
...
+                            objects_to_write.append(channel)
+
+                    tdms_writer.write_segment(objects_to_write)
+
+                except queue.Empty:
+                    continue
+
+    except Exception as e:
+        print(f"Error in consumer: {e}")
+        stop_event.set()
```
bkeryan: Futures automatically propagate exceptions between threads, so you may be able to remove this or move it to the main thread.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: I don't see sample_rate argument in AnalogWaveform, I can see that there is sample_count. Could you advise further?

bkeryan: It's like the LabVIEW and .NET waveform data types. `waveform.timing.sample_interval` is equivalent to the `dt` field in LabVIEW and to `waveform.Timing.SampleInterval` in .NET.

`sample_rate = 1.0 / waveform.timing.sample_interval.total_seconds()`

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Made it a single device example and added comment

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Using Reader for analog and counter

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: added comment for the queue

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Implemented

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Added comment

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Modified to ci

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: yes, modified it

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Producer put none when task ends

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Implemented

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Implemented queue.Queue[List[TaskData]]

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: this line is not needed, removed it

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Implemented

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Modified name to generic Channel and Counter

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Implemented

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Implemented

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Removed it

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Implemented

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,252 @@
...
+        while not stop_event.is_set():
+            data = []
+
+            ai_reader.read_many_sample(
+                ai_data, number_of_samples_per_channel=SAMPLES_PER_CHANNEL, timeout=TIMEOUT
+            )
+
+            ai_reader.read_waveforms(
+                ai_waveforms, number_of_samples_per_channel=SAMPLES_PER_CHANNEL, timeout=TIMEOUT
+            )
```
bkeryan: Don't read the data twice. read_waveforms() gives you both the data and the timing.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,252 @@
...
+TIMEOUT = 10.0
+# Note: This example currently assumes all tasks run on a single synchronized device.
+# To make this example work across multiple devices, export
+# a shared sample clock from one device to the other via a PFI line.
+DEVICE_NAME = "Dev1"
+
+TaskData = Union[
+    List[np.ndarray],  # Analog input: list of arrays
+    np.ndarray,  # Counter input: array
+    float,  # Sample rate
```
bkeryan: Why are you putting the sample rate into the queue separately? The point of a waveform object is to bundle the data and the timing into a single object. Put a waveform or a list of waveforms into the queue. 

The counter channels don't support waveforms, but you can create a waveform and call read_many_sample with the waveform's data buffer.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
bkeryan: The point of the waveform is to bundle data and timing into a single object. This is like the waveforms in LabVIEW, where you can connect a single wire from DAQmx Read to a graph indicator and it shows the data with appropriate timing on the X-axis.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,252 @@
...
+    # ]
+    ai_reader = AnalogMultiChannelReader(tasks[0].in_stream)
+    counter_reader = CounterReader(tasks[1].in_stream)
+
+    num_ai_channels = len(tasks[0].ai_channels.all)
+
+    ai_data = np.zeros((num_ai_channels, SAMPLES_PER_CHANNEL), dtype=np.float64)
+    counter_data = np.zeros(SAMPLES_PER_CHANNEL, dtype=np.float64)
+
+    ai_waveforms = [
```
bkeryan: Allocate new waveform objects for each read so that you can put them in the queue and you won't modify them during the next read.

It is possible to optimize this application by having the consumer put the used waveforms into a shared pool so the producer can reuse them, but I don't think it's necessary for an example program to demonstrate this.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Modified to only use read_waveforms()

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Implemented putting waveform or list of waveform

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Allocated new waveform objects for each read

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,240 @@
...
+            ]
+            counter_waveform = AnalogWaveform(sample_count=SAMPLES_PER_CHANNEL)
+
+            ai_reader.read_waveforms(
+                ai_waveforms, number_of_samples_per_channel=SAMPLES_PER_CHANNEL, timeout=TIMEOUT
+            )
+            data.append(ai_waveforms)
+
+            counter_reader.read_many_sample_double(
+                counter_waveform._data,
```
bkeryan: `_data` is private, so it's off-limits. Use the public `raw_data` property, if nidaqmx accepts a view. 
```suggestion
                counter_waveform.raw_data,
```

If not, then allocate the ndarray first and then pass it into the AnalogWaveform constructor.
```
ci_data = np.zeros(...)
read_many_sample_double(ci_data, ...)
counter_waveform = AnalogWaveform(raw_data=ci_data)
```

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,240 @@
...
+                number_of_samples_per_channel=SAMPLES_PER_CHANNEL,
+                timeout=TIMEOUT,
+            )
+            data.append(counter_waveform)
```
bkeryan: Consider filling in the counter waveform's timing and extended properties.

```suggestion
            counter_waveform.timing = ai_waveforms[0].timing
            counter_waveform.channel_name = tasks[1].channel_names[0]
            data.append(counter_waveform)
```

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,240 @@
...
+SAMPLES_PER_CHANNEL = 1000
+TIMEOUT = 10.0
+# Note: This example currently assumes all tasks run on a single synchronized device.
+# To make this example work across multiple devices, export
+# a shared sample clock from one device to the other via a PFI line.
+DEVICE_NAME = "Dev1"
+
+TaskData = Union[
+    List[AnalogWaveform],  # Analog input: list of waveforms
+    AnalogWaveform,  # Counter input: waveform
```
bkeryan: This doesn't match what you're putting in the queue. You're adding a `list[list[AnalogWaveform], AnalogWaveform]`. The outer list really ought to be a tuple.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
bkeryan: You added it back. Remove it again.

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,240 @@
...
+                # Case 1: Multi-channel analog input data
+                if isinstance(task_data, list) and all(
+                    isinstance(x, AnalogWaveform) for x in task_data
+                ):
+                    for chan_idx, waveform in enumerate(task_data):
+                        channel = ChannelObject(
+                            group_names[task_idx],
+                            channel_names[task_idx][chan_idx],
+                            waveform._data,
+                            properties={"Sample Rate": sample_rate},
```
bkeryan: How about the timestamp? Doesn't the TDMS viewer example in LV or the TDMS viewer included with FlexLogger show timestamps?

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,240 @@
...
+            objects_to_write = [root_object]
+
+            for task_idx, task_data in enumerate(data):
+                group = GroupObject(group_names[task_idx], properties={"Sample Rate": sample_rate})
+                objects_to_write.append(group)
+
+                # Case 1: Multi-channel analog input data
+                if isinstance(task_data, list) and all(
+                    isinstance(x, AnalogWaveform) for x in task_data
+                ):
```
bkeryan: It would be simpler to use tuple and assume 0=ai 1=ci

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Using tuple now

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: added several timestamps info referring tdms file generated from [voltage_acq_int_clk_tdms_logging.py](https://github.com/ni/nidaqmx-python/blob/master/examples/analog_in/voltage_acq_int_clk_tdms_logging.py)

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: implemented `raw_data`

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -0,0 +1,245 @@
...
+
+os.environ["NIDAQMX_ENABLE_WAVEFORM_SUPPORT"] = "1"
+
+# Configuration
+SAMPLE_RATE = 1000
+SAMPLES_PER_CHANNEL = 1000
+TIMEOUT = 10.0
+# Note: This example currently assumes all tasks run on a single synchronized device.
+# To make this example work across multiple devices, export
+# a shared sample clock from one device to the other via a PFI line.
```
bkeryan:
```suggestion
# To make this example work across multiple devices, share the sample clock
# between devices using the PXI trigger bus or a PFI line.
```

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
bkeryan, line 1: No wonder you're having trouble with typing: the PR build doesn't run mypy on the examples. 

Let's fix that: https://github.com/ni/nidaqmx-python/pull/879

Here are the errors:

```
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:71: error: Argument "timeout" to "read_waveforms" of "AnalogMultiChannelReader" has incompatible type "float"; expected "int"  [arg-type]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:82: error: Argument 1 to "put" of "Queue" has incompatible type "tuple[list[AnalogWaveform[float64]], AnalogWaveform[float64]]"; expected "Sequence[tuple[Sequence[AnalogWaveform[Any]], AnalogWaveform[Any]]]"  [arg-type]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:88: error: Argument 1 to "put" of "Queue" has incompatible type "None"; expected "Sequence[tuple[Sequence[AnalogWaveform[Any]], AnalogWaveform[Any]]]"  [arg-type]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:108: error: Statement is unreachable  [unreachable]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:112: error: "Sequence[AnalogWaveform[Any]]" has no attribute "timing"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:128: error: "object" has no attribute "raw_data"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:131: error: "object" has no attribute "timing"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:132: error: "object" has no attribute "raw_data"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:134: error: "object" has no attribute "timing"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:145: error: "tuple[Sequence[AnalogWaveform[Any]], AnalogWaveform[Any]]" has no attribute "raw_data"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:148: error: "tuple[Sequence[AnalogWaveform[Any]], AnalogWaveform[Any]]" has no attribute "timing"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:149: error: "tuple[Sequence[AnalogWaveform[Any]], AnalogWaveform[Any]]" has no attribute "raw_data"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:151: error: "tuple[Sequence[AnalogWaveform[Any]], AnalogWaveform[Any]]" has no attribute "timing"  [attr-defined]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:173: error: Need type annotation for "data_queue"  [var-annotated]
examples\synchronization\multi_function\cont_ai_ci_tdms_sync.py:210: error: Argument 3 to "submit" of "Executor" has incompatible type "Path"; expected "str"  [arg-type]
Found 15 errors in 1 file (checked 268 source files)
```

The timeout type is bug AB#3424636, cc: @mikeprosserni

lau-yeexuan: Thanks for the fix

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
lau-yeexuan: Modified

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
mikeprosserni: I've posted a PR to fix the timeout type: https://github.com/ni/nidaqmx-python/pull/880

bkeryan: Next: rebase/merge main into your branch and fix the errors.

lau-yeexuan: Modified. 

image

examples/synchronization/multi_function/cont_ai_ci_tdms_sync.py
```diff
@@ -131,7 +133,11 @@ def consumer(
                         "wf_increment": waveform.timing.sample_interval.total_seconds(),
                         "wf_samples": len(waveform.raw_data),
                         "wf_start_offset": 0.0,
-                        "wf_start_time": waveform.timing.start_time.strftime("%Y-%m-%d %H:%M:%S"),
+                        "wf_start_time": (
+                            waveform.timing.start_time
+                            if isinstance(waveform.timing.start_time, datetime)
+                            else datetime.now()
+                        ).strftime("%Y-%m-%d %H:%M:%S"),
```
bkeryan: You changed this because the timestamp might be an `nitypes.bintime.DateTime`, which doesn't support `strftime`. The correct solution is to convert it to a time type that does, not to record the wrong time.

```
"wf_start_time": waveform.timing.to_datetime().start_time.strftime("%Y-%m-%d %H:%M:%S"),
```

Same for the counter waveform.

PR #878, opened 2025-11-03
Implement gRPC Waveform Support

mikeprosserni:

For each of the nine interpreter methods in _grpc_interpreter.py:
```
read_analog_waveform()
read_analog_waveforms()
read_digital_waveform()
read_digital_waveforms()
read_new_digital_waveforms()
write_analog_waveform()
write_analog_waveforms()
write_digital_waveform()
write_digital_waveforms()
```
This PR implements the interpreter method by calling self._client.<RPC>

[AB#3424636](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3424636)

All of the waveform tests that had been disabled for grpc have now been enabled.

### DIFF
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -6,7 +6,9 @@ import threading
 import typing
 import warnings
 from nitypes.waveform import AnalogWaveform, DigitalWaveform
-from typing import Any, Callable, Generic, Sequence, TypeVar
+from typing import Any, Generic, TypeVar
+
+from collections.abc import Callable, Sequence
 
 import google.protobuf.message
 import grpc
@@ -23,7 +25,6 @@ from ni.protobuf.types.waveform_conversion import (
     float64_analog_waveform_to_protobuf
 )
 from nidaqmx.constants import WaveformAttributeMode
-from nidaqmx.errors import DaqError
 from nidaqmx.error_codes import DAQmxErrors
 from nidaqmx._grpc_time import convert_time_to_timestamp, convert_timestamp_to_time
 from nidaqmx._waveform_utils import get_num_samps_per_chan
--- a/src/codegen/templates/_grpc_interpreter.py.mako
+++ b/src/codegen/templates/_grpc_interpreter.py.mako
@@ -23,7 +23,9 @@ import threading
 import typing
 import warnings
 from nitypes.waveform import AnalogWaveform, DigitalWaveform
-from typing import Any, Callable, Generic, Sequence, TypeVar
+from typing import Any, Generic, TypeVar
+
+from collections.abc import Callable, Sequence
 
 import google.protobuf.message
 import grpc
@@ -40,7 +42,6 @@ from ni.protobuf.types.waveform_conversion import (
     float64_analog_waveform_to_protobuf
 )
 from nidaqmx.constants import WaveformAttributeMode
-from nidaqmx.errors import DaqError
 from nidaqmx.error_codes import DAQmxErrors
 from nidaqmx._grpc_time import convert_time_to_timestamp, convert_timestamp_to_time
 from nidaqmx._waveform_utils import get_num_samps_per_chan


generated/nidaqmx/_base_interpreter.py
```diff
@@ -1954,3 +1954,21 @@ def write_digital_waveforms(
         timeout: float,
     ) -> int:
         raise NotImplementedError
+
+    def _get_num_samps_per_chan(self, waveforms: Sequence[Any]) -> int:
```
bkeryan:
```suggestion
    def _get_num_samps_per_chan(self, waveforms: Sequence[AnalogWaveform[Any] | DigitalWaveform[Any]]) -> int:
```

bkeryan: I don't think there is a good reason to define this as an interpreter method. Please change it to a free function in a `_waveform_utils` submodule.

generated/nidaqmx/_stubs/ni/protobuf/types/precision_timestamp_pb2.pyi
bkeryan, line 1: Duplicating protobuf codegen is extremely brittle. Do not add any more duplicate protobuf codegen.

Instead, add https://pypi.org/project/ni.protobuf.types/ to the grpc extra.

src/codegen/stub_generator.py
bkeryan, line 1: Revert this file. We created https://pypi.org/project/ni.protobuf.types/ so we wouldn't have to do stuff like this.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -145,7 +147,11 @@ class GrpcStubInterpreter(BaseInterpreter):
                 'This operation is not supported by the NI gRPC Device Server being used. Upgrade NI gRPC Device Server.'
             )
         if error_code is None:
-            raise errors.RpcError(grpc_error, error_message) from None
+            # Convert certain gRPC validation errors to DaqError for consistency with library version
```
bkeryan: Which errors are these? Perhaps grpc-device should return ni-error metadata in these cases.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -265,7 +277,21 @@ class GrpcStubInterpreter(BaseInterpreter):
         waveforms: Sequence[AnalogWaveform[numpy.float64]],
         waveform_attribute_mode: WaveformAttributeMode
     ) -> int:
-        raise NotImplementedError
+        response = self._invoke(
+            self._client.ReadAnalogWaveforms,
+            grpc_types.ReadAnalogWaveformsRequest(
+                task=typing.cast(typing.Any, task_handle),
```
bkeryan: Narrow the type the same way you did for LibraryInterpreter

```
assert isinstance(task_handle, str)
```

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -265,7 +277,21 @@ class GrpcStubInterpreter(BaseInterpreter):
...
+                num_samps_per_chan=number_of_samples_per_channel,
+                timeout=timeout,
+                waveform_attribute_mode_raw=waveform_attribute_mode.value
+            ))
+
+        for i, grpc_waveform in enumerate(response.waveforms):
+            if i < len(waveforms):
```
bkeryan: Ignoring extra waveforms may mask other channel-swapping issues. I think it's better to raise an exception.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -325,16 +411,105 @@ class GrpcStubInterpreter(BaseInterpreter):
...
+        data_array = numpy.frombuffer(grpc_waveform.y_data, dtype=numpy.uint8)
+        data_array = data_array.reshape(samples_received, signal_count)
+        target_waveform.data[:samples_received, :signal_count] = data_array
+    _copy_timing_information(grpc_waveform, target_waveform, waveform_attribute_mode)
+    _copy_waveform_attributes(grpc_waveform, target_waveform)
+
+def _copy_analog_waveform_to_protobuf_waveform(waveform, grpc_waveform):
+    scaled_data = waveform.scaled_data
+    if not scaled_data.flags.c_contiguous:
+        scaled_data = scaled_data.copy(order="C")
```
bkeryan: Don't reinvent the wheel. There are conversion functions for protobuf <-> nitypes: https://github.com/ni/ni-apis-python/blob/main/packages/ni.protobuf.types/src/ni/protobuf/types/waveform_conversion.py

They don't support copying to an existing waveform yet. I think we should add that capability to ni.protobuf.types. 

For now, use the existing ni.protobuf.types conversion functions to return a temp waveform and then copy its data, timing, attributes, and scale mode into the passed-in waveform. We should consider adding this copy-into capability to nitypes.

Do not assume that the passed-in waveform has empty data/timing/attributes.

You don't have to check WaveformAttributeMode because the grpc-device server did that for you.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -298,7 +348,25 @@ class GrpcStubInterpreter(BaseInterpreter):
...
+                waveform_attribute_mode_raw=waveform_attribute_mode.value
+            ))
+
+        waveforms = []
+        for grpc_waveform in response.waveforms:
+            signal_count = grpc_waveform.signal_count
+            samples_received = len(grpc_waveform.y_data) // signal_count if signal_count > 0 else 0
+            waveform = DigitalWaveform(samples_received, signal_count=signal_count)
+            _copy_protobuf_waveform_to_digital_waveform(grpc_waveform, waveform, waveform_attribute_mode)
+            waveforms.append(waveform)
```
bkeryan:
```suggestion
        waveforms = [digital_waveform_from_protobuf(grpc_waveform) for grpc_waveform in response.waveforms]
```

src/codegen/templates/_grpc_interpreter.py.mako
bkeryan: BTW, use load_data to overwrite the data.

generated/nidaqmx/_base_interpreter.py
mikeprosserni: updated

mikeprosserni: moved to _waveform_utils.py

src/codegen/stub_generator.py
mikeprosserni: I reverted everything in this file except this:

`f"--proto_path={str(NI_APIS_PATH)}",  # ni-apis root path for ni.protobuf.types import resolution`

generated/nidaqmx/_stubs/ni/protobuf/types/precision_timestamp_pb2.pyi
mikeprosserni: this has been reverted, in favor of ni.protobuf.types

src/codegen/templates/_grpc_interpreter.py.mako
mikeprosserni: updated to use asserts intead of casts

src/codegen/templates/_grpc_interpreter.py.mako
mikeprosserni: changed to raise an exception if len(response.waveforms) != len(waveforms)

src/codegen/templates/_grpc_interpreter.py.mako
mikeprosserni: I've rewritten this to use the waveform_conversion methods (and load_data). Notably, the new code handles converting from bintime (but gets hightime for free).

src/codegen/templates/_grpc_interpreter.py.mako
mikeprosserni: updated

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -7,8 +7,9 @@
 import warnings
 from datetime import timezone
 from hightime import datetime as ht_datetime, timedelta as ht_timedelta
+from nitypes import bintime
 from nitypes.waveform import AnalogWaveform, DigitalWaveform, SampleIntervalMode, Timing
-from typing import Any, Callable, Generic, Sequence, TypeVar
+from typing import Any, Callable, Generic, Optional, Sequence, TypeVar
```
bkeryan: This project no longer supports Python 3.9, so use `T | None` instead of `Optional[T]`.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3682,20 +3692,23 @@ def read_digital_waveforms(
...
 
+        if len(response.waveforms) != len(waveforms):
+            raise ValueError(f"Expected {len(waveforms)} waveforms but received {len(response.waveforms)} from server")
+
         for i, grpc_waveform in enumerate(response.waveforms):
-            if i < len(waveforms):
-                if waveforms[i].data.shape[1] != grpc_waveform.signal_count:
-                    raise ValueError(f"waveforms[{i}].data has {waveforms[i].data.shape[1]} signals, but expected {grpc_waveform.signal_count}")
-                _copy_protobuf_waveform_to_digital_waveform(grpc_waveform, waveforms[i], waveform_attribute_mode)
+            if waveforms[i].data.shape[1] != grpc_waveform.signal_count:
```
bkeryan: load_data already validates the signal count, so this check should not be necessary. 

BTW, there is an accessor for `signal_count`.

src/codegen/templates/_grpc_interpreter.py.mako
bkeryan: That's 2 out of 4 wheels. Now do the other 2.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -18,11 +19,17 @@
 from nidaqmx._base_interpreter import BaseEventHandler, BaseInterpreter
 from nidaqmx._stubs import nidaqmx_pb2 as grpc_types
 from nidaqmx._stubs import nidaqmx_pb2_grpc as nidaqmx_grpc
-from nidaqmx._stubs.ni.protobuf.types import waveform_pb2
+from ni.protobuf.types import waveform_pb2
+from ni.protobuf.types.waveform_conversion import (
+    digital_waveform_from_protobuf,
+    float64_analog_waveform_from_protobuf
```
bkeryan: Use the to_protobuf functions too.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3801,61 +3811,54 @@ def write_digital_waveforms(
...
-    _setup_waveform_capacity_and_samples(target_waveform, samples_received)
-    if samples_received > 0 and signal_count > 0:
-        data_array = numpy.frombuffer(grpc_waveform.y_data, dtype=numpy.uint8)
-        data_array = data_array.reshape(samples_received, signal_count)
-        target_waveform.data[:samples_received, :signal_count] = data_array
-    _copy_timing_information(grpc_waveform, target_waveform, waveform_attribute_mode)
-    _copy_waveform_attributes(grpc_waveform, target_waveform)
+    _copy_timing_and_properties_from_temp_waveform(temp_waveform, target_waveform)
 
 def _copy_analog_waveform_to_protobuf_waveform(waveform, grpc_waveform):
```
bkeryan: Delete these copy_..._to_protobuf functions.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3801,61 +3811,54 @@ def write_digital_waveforms(
...
+    except RuntimeError:
+        sample_interval = None
+
+    if timestamp is not None or sample_interval is not None:
+        target_waveform.timing = Timing(
+            temp_waveform.timing.sample_interval_mode,
+            timestamp=timestamp,
+            sample_interval=sample_interval
+        )
+    else:
```
bkeryan: Why are you doing any of this? The Timing object was already converted to the correct format. `target_waveform.timing = temp_waveform.timing` should be all you need.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -3801,61 +3811,54 @@ def write_digital_waveforms(
...
+        target_waveform.timing = temp_waveform.timing
+
+    target_waveform.extended_properties.clear()
+    target_waveform.extended_properties.update(temp_waveform.extended_properties)
+
+
+def _copy_protobuf_waveform_to_analog_waveform(grpc_waveform, target_waveform):
+    temp_waveform = float64_analog_waveform_from_protobuf(grpc_waveform)
+    target_waveform.load_data(temp_waveform.scaled_data)
+    _copy_timing_and_properties_from_temp_waveform(temp_waveform, target_waveform)
```
bkeryan: Also assign scale_mode

generated/nidaqmx/_waveform_utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+    Args:
+        waveforms: Sequence of analog or digital waveforms to validate.
+
+    Returns:
+        int: The number of samples per channel.
+
+    Raises:
+        ValueError: If no waveforms are provided.
+        DaqError: If waveforms have different sample counts.
```
bkeryan: Why two different exception types?

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
bkeryan: BTW, you may be able to use a list comprehension not a for-loop.

generated/nidaqmx/_waveform_utils.py
bkeryan: Wait. This is a public function, so nps will require a docstring, but it's in an internal module, so it doesn't need a detailed docstring that lists what exceptions it might throw. Try deleting everything but the 1st line summary.

generated/nidaqmx/_grpc_interpreter.py
mikeprosserni: removed Optional

generated/nidaqmx/_grpc_interpreter.py
mikeprosserni: I've removed this check. (I also had to update the asserts in the tests that were expecting a specific error string, and are now getting a different one.)

generated/nidaqmx/_grpc_interpreter.py
mikeprosserni: updated to use digital_waveform_to_protobuf and float64_analog_waveform_to_protobuf (using list comprehensions)

generated/nidaqmx/_grpc_interpreter.py
mikeprosserni: deleted

generated/nidaqmx/_grpc_interpreter.py
mikeprosserni: I was doing all that because the tests were written to expect hightime datatimes specifically, and the waveform_converter methods are giving us bintime timestamps. 

But, I've removed all this in favor of simply `target_waveform.timing = temp_waveform.timing` and updated the tests no longer assert that the timestamps are hightime. 

(The tests still all have `assert _is_timestamp_close_to_now(waveform.timing.timestamp)`, which work fine with the bintime timestamps)

src/codegen/templates/_grpc_interpreter.py.mako
mikeprosserni: I've done all four now.

generated/nidaqmx/_waveform_utils.py
mikeprosserni: I changed the DaqError to ValueError, so they are both now ValueError. (let me know if DaqError would actually be more appropriate for both)

I originally went with the more verbose comments because the utility methods in `_bitfield_utils.py` have verbose comments, but I've now removed all but the first line summary here.

generated/nidaqmx/_grpc_interpreter.py
mikeprosserni: added `waveforms[i].scale_mode = temp_waveform.scale_mode` in `read_analog_waveforms()`

src/codegen/templates/_grpc_interpreter.py.mako
mikeprosserni: After changing those asserts from expecting DaqError to expecting ValueError, this is no longer needed. I've reverted `_handle_rpc_error()`.

generated/nidaqmx/_grpc_interpreter.py
bkeryan: This is a reasonable consequence of our choice to serialize timestamps as NI-BTF. Clients who want timestamps in a specific format should call `.to_hightime()` or whatever.

PR #880, opened 2025-11-06
Fix timeout type in read_waveform() methods

mikeprosserni:

In some of the read_waveform() methods (in AnalogMultiChannelReader, AnalogSingleChannelReader, and DigitalMultiChannelReader), the timeout parameter was marked as 'int', when they should all be 'float'.

(Also,`/nidaqmx/system/__init__.py` was missing an import, which was causing unrelated mypy errors, which I have fixed.)

[AB#3462409](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3462409)

Ran lint, mypy, and all the autotests


generated/nidaqmx/system/__init__.py
```diff
@@ -1,5 +1,6 @@
 """NI-DAQmx system classes."""
 
+from nidaqmx.system import storage
```
bkeryan, line 3: Why did this start causing problems now and not previously?

Issue #812, opened 2025-08-22
`READ_ALL_AVAILABLE` should use `DefaultNumberOfSamplesToRead` on DAQmx versions that support it


bkeryan:
## Problem to Solve

Specifying `samples_to_read=READ_ALL_AVAILABLE` (-1) tells DAQmx to automatically query `task.in_stream.avail_samp_per_chan`, but there are additional corner cases:

- For finite acquisitions:
  - Use `task.timing.samp_quant_samp_per_chan` instead.
  - ...Unless `task.in_stream.read_all_avail_samp` is set to False.
- Reading before the task is verified

## Proposed Solution

NI-DAQmx 24.5 or 24.8 added support for a new internal attribute, `DefaultNumberOfSamplesToRead`, which indicates the raw data size in bits. I think the corresponding C API function name is `DAQmxGetDefaultNumberOfSamplesToRead`. If this function is available, nidaqmx-python should use it instead of calculating the value based on multiple attributes.

[AB#3240032](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3240032)

PR #842, opened 2025-09-23
Add benchmarks for stream readers and stream writers

mikeprosserni:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).
- [ ] ~~I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~
- [x] I've added tests applicable for this pull request

### What does this Pull Request accomplish?

This PR adds benchmark tests for `AnalogSingleChannelReader`, `AnalogMultiChannelReader`, `DigitalSingleChannelReader`, `DigitalMultiChannelReader`, `AnalogSingleChannelWriter`, `AnalogMultiChannelWriter`, `DigitalSingleChannelWriter`, and `DigitalMultiChannelWriter`. 

### Why should this Pull Request be merged?

[AB#3309084](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3309084)

### What testing has been done?

See benchmark results in comment below.

PR #881, opened 2025-11-06
Add support for read_waveform(READ_ALL_AVAILABLE)

mikeprosserni:

This adds support for READ_ALL_AVAILABLE to these methods in _library_interpreter.py:

```
read_analog_waveform()
read_analog_waveforms()
read_digital_waveform()
read_digital_waveforms()
read_new_digital_waveforms()
```

Also, I discovered that the return type annotation in DigitalMultiChannelReader.read_waveforms() was incorrect. `_interpreter.read_digital_waveforms()` returns the samples read (an int), not the list of waveforms. I updated the annotation to be an int (which matches the documentation, which was correct).

[AB#3460923](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3460923)

https://github.com/ni/nidaqmx-python/issues/812

Added tests:
```
test___analog_multi_channel_reader___read_waveforms_read_all_available___returns_valid_waveforms
test___analog_single_channel_reader___read_waveform_read_all_available___returns_valid_waveform
test___digital_multi_channel_multi_line_reader___read_waveforms_read_all_available___returns_valid_waveforms
test___digital_single_line_reader___read_waveform_read_all_available___returns_valid_waveform
test___analog_single_channel___read_waveform_read_all_available___returns_valid_waveform
test___digital_single_channel___read_waveform_read_all_available___returns_valid_waveform
```


src/codegen/templates/_library_interpreter.py.mako
```diff
@@ -197,6 +211,9 @@ class LibraryInterpreter(BaseInterpreter):
         waveform_attribute_mode: WaveformAttributeMode
     ) -> int:
         """Read an analog waveform with timing and attributes."""
+        if number_of_samples_per_channel == READ_ALL_AVAILABLE:
+            number_of_samples_per_channel = self.get_default_number_of_samples_to_read(task_handle)
```
bkeryan: This does not belong in the interpreter. All the other read methods do it in the reader:

https://github.com/ni/nidaqmx-python/blob/master/src/handwritten/stream_readers/_analog_multi_channel_reader.py#L88

https://github.com/ni/nidaqmx-python/blob/master/src/handwritten/task/_task.py#L282

src/codegen/templates/_library_interpreter.py.mako
bkeryan, line 1: There is a bug about fixing this for all the other read methods: https://github.com/ni/nidaqmx-python/issues/812

The existing logic in Task._calculate_num_samps_per_chan should be updated to use get_default_number_of_samples_to_read when it is available (library && >=24.8?) and fallback to the old code otherwise.
https://github.com/ni/nidaqmx-python/blob/master/src/handwritten/task/_task.py#L282

I know you probably didn't set out to fix that bug, but I don't think it makes sense to solve the same problem in a completely different way for waveforms.

mikeprosserni: Is this a good way to use get_default_number_of_samples_to_read when it is available, and fall back to the old code otherwise? (Right now we get an AttributeError when trying to access `grpc_types.GetDefaultNumberOfSamplesToReadRequest(task=task)`, but I assume that will start working after grpc-device is updated.)

```
        elif num_samps_per_chan == READ_ALL_AVAILABLE:
            if hasattr(self._interpreter, 'get_default_number_of_samples_to_read'):
                try:
                    return self._interpreter.get_default_number_of_samples_to_read(self._handle)
                except AttributeError:
                    pass
            acq_type = self.timing.samp_quant_samp_mode

            if acq_type == AcquisitionType.FINITE and not self.in_stream.read_all_avail_samp:
                return self.timing.samp_quant_samp_per_chan
            else:
                return self.in_stream.avail_samp_per_chan
```

bkeryan: Using hasattr to check whether a method exists is a reasonable and legit thing to do. Protocols are a formalization of this: 
https://typing.python.org/en/latest/spec/protocol.html

However, that won't address compatibility with NI-DAQmx 23.0. The try/except could address this, but:
- DaqFunctionNotSupportedError isn't a subclass of AttributeError
- Raising and catching exceptions is apparently much more expensive than a conditional, but I don't know how much

mikeprosserni: I've changed everything to use GetReadAttributeUInt32(DAQmx_DefaultNumberOfSamplesToRead) instead of GetDefaultNumberOfSamplesToRead(). Now my code looks like this:

```
        elif num_samps_per_chan == READ_ALL_AVAILABLE:
            try:
                # DAQmx_DefaultNumberOfSamplesToRead is 0x31E8
                return self._interpreter.get_read_attribute_uint32(self._handle, 0x31E8)
            except DaqError:
                acq_type = self.timing.samp_quant_samp_mode

                if acq_type == AcquisitionType.FINITE and not self.in_stream.read_all_avail_samp:
                    return self.timing.samp_quant_samp_per_chan
                else:
                    return self.in_stream.avail_samp_per_chan
```

`get_read_attribute_uint32` throws a `DaqError('Property specified is not valid for this function.\n\nStatus Code: -200233')` if DAQmx_DefaultNumberOfSamplesToRead is not found.

On my dev machine, in the `test___persisted_task___load_and_read___returns_persisted_sample_count` test:
- The library version goes through the `return self._interpreter.get_read_attribute_uint32(self._handle, 0x31E8)` branch succesfully, in 0.47 seconds.
- The grpc version goes throws the exception and uses the fallback code, in 0.66 seconds.

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: Oh, right, yeah, the read_waveform methods were also already doing it in the reader. These in the interpreter were redundant, and I've removed them.

src/codegen/templates/_library_interpreter.py.mako
bkeryan: 470 ms is huge compared to your benchmark results from https://github.com/ni/nidaqmx-python/pull/842

I'm writing a READ_ALL_AVAILABLE benchmark case but in the process I rediscovered https://github.com/ni/nidaqmx-python/issues/421

bkeryan: also https://github.com/ni/nidaqmx-python/issues/528

bkeryan: *sigh* I just noticed something else: AB#3465696 DAQmx read benchmarks allocate a much larger buffer than necessary for >1 channel

bkeryan: I ran my benchmark with your change, and then I hacked it to generate an error by using the wrong attribute, and it went from 40-60 us to 3 ms.

https://github.com/ni/nidaqmx-python/pull/883#issuecomment-3505393230

I don't think that's acceptable performance. Ideas:
- Check DAQmx version
- Use error fallback, but cache the result and go straight to the fallback next time

bkeryan: `if System.local().driver_version >= (24, 8):` actually seems to have ok performance, at least for library interpreter.

But we should reuse the task's interpreter instead of calling local(). And maybe cache the system object or the version anyway, since this is used in read methods.

bkeryan: Querying driver version over gRPC would add 3 network round trips. I think it would make sense to move the implementation of System.driver_version to an interpreter method and cache the result for the lifetime of the interpreter.

mikeprosserni: I've updated the check to use the driver version:
```
        elif num_samps_per_chan == READ_ALL_AVAILABLE:
            if self._interpreter.driver_version >= (24, 8):
                # DAQmx_DefaultNumberOfSamplesToRead is 0x31E8
                return self._interpreter.get_read_attribute_uint32(self._handle, 0x31E8)
            else:
                acq_type = self.timing.samp_quant_samp_mode

                if acq_type == AcquisitionType.FINITE and not self.in_stream.read_all_avail_samp:
                    return self.timing.samp_quant_samp_per_chan
                else:
                    return self.in_stream.avail_samp_per_chan
```

But despite the fact that the driver_version for both the library interpreter and the grpc interpreter are "25.3.1", I'm still seeing the -200233 DaqError on the grpc version of the test___persisted_task___load_and_read___returns_persisted_sample_count test. 

If the library interpreter and the grpc interpreter are using the same version of the driver, why would GetReadAttributeUInt32(DAQmx_DefaultNumberOfSamplesToRead) work for one but not the other?

mikeprosserni: I've updated the check to use the driver version, and to also handle the DaqError, and also skip the try if it has failed in the past

```
        elif num_samps_per_chan == READ_ALL_AVAILABLE:
            if (
                self._interpreter.driver_version >= (24, 8)
                and self._default_number_of_samples_attribute_available
            ):
                try:
                    # DAQmx_DefaultNumberOfSamplesToRead is 0x31E8
                    return self._interpreter.get_read_attribute_uint32(self._handle, 0x31E8)
                except DaqError:
                    self._default_number_of_samples_attribute_available = False

            acq_type = self.timing.samp_quant_samp_mode

```
All the tests pass like this, and by skipping to the fallback if there's ever been an error, it should be pretty performant.

But, if we're checking for _default_number_of_samples_attribute_available anyway, is it still worth having the driver version check also?

mikeprosserni: Here's the current performance on my dev machine:
```
PS C:\dev\fireserp\nidaqmx-python> poetry run pytest -v .\tests\benchmark\test_analog_stream_readers.py -k "library and (read_all_available or read_many_sample)"
================================================================================== test session starts ===================================================================================
platform win32 -- Python 3.11.9, pytest-8.4.2, pluggy-1.6.0 -- C:\dev\fireserp\nidaqmx-python\.venv\Scripts\python.exe
cachedir: .pytest_cache
benchmark: 5.2.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: C:\dev\fireserp\nidaqmx-python
configfile: pyproject.toml
plugins: benchmark-5.2.0, cov-7.0.0, mock-3.15.1
collected 88 items / 72 deselected / 16 selected

tests/benchmark/test_analog_stream_readers.py::test___analog_single_channel_reader___read_many_sample[library_init_kwargs-1] PASSED                                                 [  6%]
tests/benchmark/test_analog_stream_readers.py::test___analog_single_channel_reader___read_many_sample[library_init_kwargs-1000] PASSED                                              [ 12%]
tests/benchmark/test_analog_stream_readers.py::test___analog_single_channel_reader___read_all_available[library_init_kwargs-1] PASSED                                               [ 18%]
tests/benchmark/test_analog_stream_readers.py::test___analog_single_channel_reader___read_all_available[library_init_kwargs-1000] PASSED                                            [ 25%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1-1] PASSED                                                [ 31%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1-2] PASSED                                                [ 37%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1-8] PASSED                                                [ 43%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1000-1] PASSED                                             [ 50%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1000-2] PASSED                                             [ 56%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1000-8] PASSED                                             [ 62%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1-1] PASSED                                              [ 68%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1-2] PASSED                                              [ 75%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1-8] PASSED                                              [ 81%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1000-1] PASSED                                           [ 87%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1000-2] PASSED                                           [ 93%]
tests/benchmark/test_analog_stream_readers.py::test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1000-8] PASSED                                           [100%]


-------------------------------------------------------------------------------------------------------------- benchmark 'analog_readers': 16 tests --------------------------------------------------------------------------------------------------------------
Name (time in us)                                                                           Min                    Max               Mean              StdDev             Median           
    IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1-1]           9.5000 (1.0)         659.0000 (5.26)     12.7003 (1.12)      11.8116 (2.94)     10.8000 (1.02)     2.2000 (4.40)      135;438       78.7380 (0.89)       4931           1
test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1-2]           9.6000 (1.01)     10,500.7000 (83.80)    12.6446 (1.12)     105.7712 (26.33)    10.6000 (1.0)      0.6001 (1.20)       6;1045       79.0850 (0.89)       9882           1
test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1-8]           9.7000 (1.02)        436.4001 (3.48)     11.2946 (1.0)        7.4047 (1.84)     10.7001 (1.01)     0.5000 (1.0)       153;640       88.5383 (1.0)        9672           1
test___analog_single_channel_reader___read_many_sample[library_init_kwargs-1]           10.3000 (1.08)        125.3000 (1.0)      16.9517 (1.50)      13.3386 (3.32)     12.0500 (1.14)     6.7001 (13.40)       24;27       58.9913 (0.67)        362           1
test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1000-1]       10.8000 (1.14)        165.3000 (1.32)     12.1714 (1.08)       4.0172 (1.0)      11.7000 (1.10)     0.5000 (1.0)       207;623       82.1599 (0.93)       9209           1
test___analog_single_channel_reader___read_many_sample[library_init_kwargs-1000]        10.9000 (1.15)        499.1001 (3.98)     13.8771 (1.23)      15.1680 (3.78)     11.8000 (1.11)     1.0000 (2.00)      126;831       72.0613 (0.81)       6342           1
test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1000-2]       11.4000 (1.20)        325.0999 (2.59)     13.0805 (1.16)       7.4220 (1.85)     12.3000 (1.16)     0.7000 (1.40)       94;859       76.4494 (0.86)       9133           1
test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1-2]        11.5000 (1.21)        278.6999 (2.22)     13.5120 (1.20)       6.7911 (1.69)     12.6000 (1.19)     0.7000 (1.40)      194;850       74.0083 (0.84)       8749           1
test___analog_single_channel_reader___read_all_available[library_init_kwargs-1]         11.5000 (1.21)        268.2001 (2.14)     13.7730 (1.22)       6.3511 (1.58)     12.9001 (1.22)     1.0000 (2.00)      235;592       72.6058 (0.82)       9108           1
test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1-1]        11.6000 (1.22)        534.6000 (4.27)     14.1798 (1.26)      10.7445 (2.67)     12.7000 (1.20)     0.9000 (1.80)     225;1085       70.5227 (0.80)       9719           1
test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1-8]        11.6000 (1.22)        308.6000 (2.46)     13.8991 (1.23)       6.5943 (1.64)     12.9000 (1.22)     0.7000 (1.40)     369;1074       71.9473 (0.81)       9738           1
test___analog_single_channel_reader___read_all_available[library_init_kwargs-1000]      13.9000 (1.46)        275.8000 (2.20)     17.7544 (1.57)       8.8697 (2.21)     15.7000 (1.48)     1.6000 (3.20)     529;1322       56.3240 (0.64)       9662           1
test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1000-1]     14.0000 (1.47)        245.0000 (1.96)     17.2464 (1.53)       8.7839 (2.19)     15.0000 (1.42)     0.8000 (1.60)      170;615       57.9832 (0.65)       3265           1
test___analog_multi_channel_reader___read_many_sample[library_init_kwargs-1000-8]       14.2000 (1.49)        334.9000 (2.67)     16.9008 (1.50)       8.9739 (2.23)     15.4000 (1.45)     0.9000 (1.80)     346;1067       59.1688 (0.67)      10257           1
test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1000-2]     14.9000 (1.57)        218.8000 (1.75)     17.3020 (1.53)       6.1209 (1.52)     16.4000 (1.55)     1.0000 (2.00)      288;757       57.7968 (0.65)       9719           1
test___analog_multi_channel_reader___read_all_available[library_init_kwargs-1000-8]     20.6999 (2.18)        317.0000 (2.53)     24.8101 (2.20)      13.3018 (3.31)     22.1999 (2.09)     1.2001 (2.40)      216;493       40.3062 (0.46)       4155           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```

bkeryan: > If the library interpreter and the grpc interpreter are using the same version of the driver, why would GetReadAttributeUInt32(DAQmx_DefaultNumberOfSamplesToRead) work for one but not the other?

I'll send you some debugging info offline.

> But, if we're checking for _default_number_of_samples_attribute_available anyway, is it still worth having the driver version check also?

No, you don't need both checks. 

Delete the try/except. If querying default number of samples returns an error for another reason, such as an invalid task or the task is in the wrong state, your fallback code will disable it for the wrong reason.

src/handwritten/task/_task.py
```diff
@@ -294,6 +296,16 @@ def _calculate_num_samps_per_chan(self, num_samps_per_chan):
         if num_samps_per_chan is NUM_SAMPLES_UNSET:
             return 1
         elif num_samps_per_chan == READ_ALL_AVAILABLE:
+            if (
+                self._interpreter.driver_version >= (24, 8)
```
bkeryan: I checked dmxa and we added this property in 24.5

mikeprosserni: I've changed the check to `if self._interpreter.driver_version >= (24, 5):`

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -86,11 +87,17 @@ class GrpcStubInterpreter(BaseInterpreter):
     __slots__ = [
         '_grpc_options',
         '_client',
+        '_driver_version',
     ]
 
     def __init__(self, grpc_options):
         self._grpc_options = grpc_options
         self._client = nidaqmx_grpc.NiDAQmxStub(grpc_options.grpc_channel)
+        DriverVersion = collections.namedtuple('DriverVersion', ['major_version', 'minor_version', 'update_version'])
```
bkeryan: Don't duplicate the named tuple definition. Move it to nidaqmx.types and replace System.DriverVersion with a TypeAlias.

mikeprosserni: done

src/codegen/templates/_library_interpreter.py.mako
mikeprosserni: It turns out that nidaqmx.proto was just missing `READ_ATTRIBUTE_DEFAULT_NUMBER_OF_SAMPLES_TO_READ` as a valid value for `ReadUInt32Attribute`. Once I added it, then all of the tests passed just fine without any try/except block.

PR #885, opened 2025-11-13
nitypes = {version=">=1.0.0"}

mikeprosserni:
- [] ~~I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~
- [] ~~I've added tests applicable for this pull request~~

This is just a small detail, but now that there's a released 1.0 version of nitypes, that's what we should specify in the pyproject.toml. (Note that the version in poetry.lock was already `1.1.0.dev0`, so this PR doesn't actually change the version of nitypes in use.)

[AB#3229809](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3229809)

```
poetry lock
poetry install --all-extras --with examples
poetry run pytest
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1790,14 +1790,14 @@ toml = ">=0.10.1"
 
 [[package]]
 name = "nitypes"
-version = "1.1.0.dev0"
+version = "1.0.0"
 description = "Data types for NI Python APIs"
 optional = false
 python-versions = "<4.0,>=3.9"
 groups = ["main"]
 files = [
-    {file = "nitypes-1.1.0.dev0-py3-none-any.whl", hash = "sha256:1dd0d15bb741294672ea7d61d98e951c5802eec0130d0162f4f19008dd74c0f6"},
-    {file = "nitypes-1.1.0.dev0.tar.gz", hash = "sha256:a82d2f518f514a24bfb3b9c4e023db23097a786ca40155693610d09578d86f10"},
+    {file = "nitypes-1.0.0-py3-none-any.whl", hash = "sha256:545dff978040fd92a68876023b8b6ff666c1d861ead1b3e7771108754bd57345"},
+    {file = "nitypes-1.0.0.tar.gz", hash = "sha256:dc5ecb8b5f52570c2311daaa60ce6dca9712d7151236ef1f5314ad6ac02e0bdc"},
 ]
 
 [package.dependencies]


pyproject.toml
```diff
@@ -62,7 +62,7 @@ click = ">=8.0.0"
 distro = { version = ">=1.9.0", platform = "linux" }
 requests = ">=2.25.0"
 typing_extensions = { version = ">=4.0.0" }
-nitypes = {version=">=0.1.0dev10", allow-prereleases=true}
```
bkeryan, line 65: Removing allow-prereleases will downgrade to 1.0.0 exactly when we update poetry.lock. Is that what you want?

bkeryan: Or at least that's the behavior with poetry 1.x. I'm not sure about 2.x.

mikeprosserni: I ran `poetry lock` and it didn't downgrade to 1.0.0. I think the indirect dependency through ni-protobuf-types is what's putting it at 1.1.0.dev0.

Actually, if we're getting nitypes through ni-protobuf-types, should we even specify this direct nitypes dependency at all, or should I just remove it completely instead of updating it?

mikeprosserni: (Well, I tried removing it completely, and that made nitypes be marked as optional=true in poetry.lock, which is probably not right)

bkeryan: > I ran poetry lock and it didn't downgrade to 1.0.0. I think the indirect dependency through ni-protobuf-types is what's putting it at 1.1.0.dev0.

I doubt that this is the case. I think this is the new Poetry 2.x lock behavior.

nidaqmx depends on ni-protobuf-types >=1.0.0 and does not allow prereleases of ni-protobuf-types, and ni-protobuf-types 1.0.0 has `Requires-Dist: nitypes (>=0.1.0dev8)`.

I think the monthly Renovate lock file maintenance deletes the lock file and regenerates the whole thing from scratch, which will update all versions according to pyproject.toml.

> Actually, if we're getting nitypes through ni-protobuf-types, should we even specify this direct nitypes dependency at all, or should I just remove it completely instead of updating it?

nidaqmx with no extras doesn't depend on ni-protobuf-types. nidaqmx[grpc] depends on ni-protobuf-types. But nidaqmx with no extras needs nitypes too.

bkeryan: Anyway, allow-prereleases doesn't affect the dependencies in the published package, so this is a question of whether we want to test nidaqmx against the latest nitypes prerelease or not.

bkeryan: I suggest doing this, so we don't get surprised when Renovate tries to upgrade nitypes but actually downgrades it:
```
HEAD is now at a4ea32d nitypes = {version=">=1.0.0"}
PS C:\dev\nidaqmx-python> poetry update nitypes
Updating dependencies
Resolving dependencies... (5.6s)

Package operations: 0 installs, 1 update, 0 removals

  - Downgrading nitypes (1.1.0.dev0 -> 1.0.0)

Writing lock file
```

mikeprosserni: I've done a `poetry update nitypes` and committed the poetry.lock file with the downgraded nitypes version. All the tests are still passing.

PR #886, opened 2025-11-13
Remove WAVEFORM_SUPPORT feature toggle

mikeprosserni:

Now that all the read waveform and write waveform methods have full support for grpc-device, we can remove the WAVEFORM_SUPPORT feature toggle.

[AB#3424643](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3424643)

Ran all the auto tests

Ran all the affected input examples

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@ All notable changes to this project will be documented in this file.
 
 * ### Major Changes
     * Removed support for Python 3.9.
-    * (IN PROGRESS behind "WAVEFORM_SUPPORT" feature toggle) Added support for reading and writing Waveform data through gRPC using [NI gRPC Device Server](https://github.com/ni/grpc-device).
+    * Completed support for reading and writing Waveforms, and removed the WAVEFORM_SUPPORT feature toggle.
+        * Added support for reading and writing Waveform data through gRPC using [NI gRPC Device Server](https://github.com/ni/grpc-device).
 
 * ### Known Issues
     * ...
--- a/src/handwritten/_feature_toggles.py
+++ b/src/handwritten/_feature_toggles.py
@@ -156,5 +156,3 @@ def requires_feature(
 # --------------------------------------
 # Define feature toggle constants here:
 # --------------------------------------
-
-WAVEFORM_SUPPORT = FeatureToggle("WAVEFORM_SUPPORT", CodeReadiness.INCOMPLETE)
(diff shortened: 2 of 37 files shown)


src/handwritten/_feature_toggles.py
bkeryan, line 1: It's too early for this. Read https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/900/Feature-Toggles

> Feature toggles should only be removed once the code protected behind the feature toggle ships in the enabled state. Do not remove a feature toggle before the code ships.

tests/component/stream_readers/test_digital_multi_channel_reader.py
bkeryan, line 1: I don't want to break compatibility after release. If AB#3178052 is not fixed by release, then I think we should disable the waveform feature toggle.

```
    # Note, the data on the port's waveform is MSB instead of LSB because of bug AB#3178052
    # When that bug is fixed, these asserts should be updated
```

CHANGELOG.md
```diff
@@ -34,7 +34,8 @@ All notable changes to this project will be documented in this file.
 
 * ### Major Changes
     * Removed support for Python 3.9.
-    * (IN PROGRESS behind "WAVEFORM_SUPPORT" feature toggle) Added support for reading and writing Waveform data through gRPC using [NI gRPC Device Server](https://github.com/ni/grpc-device).
+    * Completed support for reading and writing Waveforms, and removed the WAVEFORM_SUPPORT feature toggle.
+        * Added support for reading and writing Waveform data through gRPC using [NI gRPC Device Server](https://github.com/ni/grpc-device).
```
bkeryan, line 38: The parent bullet is unnecessary and only matters to developers, not users.

src/handwritten/_feature_toggles.py
mikeprosserni: Ah, I didn't know that policy was the same for GitHub repos, but it makes sense. I'll abandon this PR for now. Also, I'll change the work item from "remove the feature toggle" to "enable the feature toggle by default".

tests/component/stream_readers/test_digital_multi_channel_reader.py
mikeprosserni: I've added AB#3178052 as a prerequisite for the "enable the waveform feature toggle" work item AB#3367327.

PR #894, opened 2025-11-24
Validate Waveform Signals in Tests

mikeprosserni:

~~Investigating [AB#3178052](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3178052) revealed that the signal indices are supposed to be reversed compared to the data array indices. There were several  `# TODO: [AB#3178052](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3178052)` comments in the nidaqmx-python digital waveform tests that were expecting this signal ordering behavior to be changed. But it turned out that the existing behavior was actually correct.~~

~~This PR overhauls the digital waveform tests to use and expect the correct signal ordering (MSB first, with line numbers counting down instead of up) throughout.~~

- ~~In `conftest.py`, all the channel strings used with `CHAN_FOR_ALL_LINES` in `add_di_chan()` have been reversed (using `[::-1]`), so they count down instead of up. (Note that this matches the behavior of port expansions)~~
- ~~In `_digital_utils.py`, I updated `_get_expected_data_for_lines()` and `_get_waveform_data()` to interpret data as MSB (big-endian), to match the standard line/signal ordering. Also, I added `_get_waveform_bitstrings()` for tests where this is the best way to see the the expected values.~~
- Renamed `_get_waveform_data_msb` to `_get_waveform_port_data`, since MSB is standard and correct for ports in waveforms.
  - Addressed/removed all `TODO: AB# 3178052` comments
- Added `_validate_waveform_signals()` to all the tests that assert waveform data. This method verifies the signals' names and data. 

~~Our tests should use industry standard line/signal ordering, for clarity.~~

We need to test the waveform signals.

All the tests pass.


tests/component/task/test_task_read_waveform_di.py
```diff
@@ -61,28 +71,26 @@ def test___digital_single_channel___read_waveform_lines_and_port___returns_valid
...
+        "10000000100",
+        "10100000101",
+        "11000000110",
+        "11100000111",
+        "00000001000",
+        "00100001001",
+    ]
     assert waveform.sample_count == samples_to_read
     assert waveform.channel_name == di_single_chan_lines_and_port_task.di_channels[0].name
-    assert waveform._get_signal_names() == [
```
bkeryan, line 69: I assume this removes all references to internal methods like `_get_signal_names`.

tests/component/conftest.py
```diff
@@ -156,7 +156,7 @@ def di_single_channel_multi_line_task(
 ) -> nidaqmx.Task:
     """Configure a single-channel multi-line digital input task."""
     task.di_channels.add_di_chan(
-        flatten_channel_string(sim_6363_device.di_lines.channel_names[:8]),
+        flatten_channel_string(sim_6363_device.di_lines.channel_names[:8][::-1]),
```
bkeryan: > In conftest.py, all the channel strings used with CHAN_FOR_ALL_LINES in add_di_chan() have been reversed (using [::-1]), so they count down instead of up. (Note that this matches the behavior of port expansions)

I disagree with reversing the line order for the fixtures.

- The non-waveform tests were correct before.
- Users typically specify lines in forward order
- The port expansion only counts down for waveforms. It counts up for read_digital_lines().

If it's really necessary to use reverse order for waveform tests, then use separate fixtures or parameterize them.

tests/component/task/test_task_read_waveform_di.py
```diff
@@ -61,28 +71,26 @@ def test___digital_single_channel___read_waveform_lines_and_port___returns_valid
 
     waveform = di_single_chan_lines_and_port_task.read_waveform(samples_to_read)
 
-    # Note, the data on the port's waveform is MSB instead of LSB because of bug AB#3178052
-    # When that bug is fixed, these asserts should be updated
-    assert _get_waveform_data(waveform) == [0, 1025, 514, 1539, 260, 1285, 774, 1799, 128, 1153]
+    assert _get_waveform_bitstrings(waveform) == [
+        "00000000000",
```
bkeryan: Why not use binary integer literals? Do the bitstrings produce better failure messages?

```python
>>> 0b10000
16
```

bkeryan (changes requested):

tests/component/task/test_task_read_waveform_di.py
mikeprosserni: Yes, I think it does. I don't know if there's a better systematic way to check, but I get no results when I search for `_get_signal_names` or `waveform._`

tests/component/conftest.py
mikeprosserni: I reverted all changes to conftest.py. All the data asserts have been reverted (except that `_get_waveform_data_msb()` has been renamed to `_get_waveform_port_data()`). Now this PR is basically just about adding `_validate_waveform_signals()` to the tests.

tests/component/task/test_task_read_waveform_di.py
mikeprosserni: I've changed these to binary integer literals. (Bitstrings don't have any advantages over them)

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.1.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.2.1 and should not be changed by hand.
```
bkeryan: FYI, I recommend using 2.1.4 like the build workflows do, but it doesn't look like 2.2.x changed the poetry.lock schema so the result won't be any different except for this comment.

bkeryan (approved):
I think this looks correct. I would appreciate if @zhindes would take another look at the latest version before we merge this.

poetry.lock
bkeryan, line 1: > Note that until https://github.com/ni/nitypes-python/pull/222 goes in, and nidaqmx takes up the new version of niypes, the signal order here will be wrong.
> Note that there's also a bug where the signal names are cached and not updated, so some of the _validate_waveform_signals() checks have to be disabled until that is fixed.

Please update this part of the PR description

mikeprosserni: I've updated the title and description of this PR

poetry.lock
mikeprosserni: I've ensured both of my dev machines now have pip 2.1.4.

zhindes (approved):

PR #902, opened 2025-12-17
examples: Revise cont dig port output examples

bkeryan:

cont_dig_port_int_clk.py
- Add a comment about the walking 1's pattern
- Generate data based on `chan.do_num_lines` instead of hardcoding it
- Print the data.

cont_dig_port_int_clk_wfm.py
- Use a walking 1's pattern like the other example.
- Use `DigitalWaveform.from_port` instead of setting the data for each line.
- Save the result of `add_do_chan` instead of hardcoding the channel index.

I think a walking 1's pattern is a better example and connectivity test than binary counting because it toggles each line in an unambiguous order. Binary counting with a small number of samples does not toggle every line. The current implementation duplicates the count for each byte to cover more lines, but this would hide byte swapping.

Ran both examples on a PCIe-6353 connected to two BNC-2120s.

### DIFF
--- a/examples/digital_out/cont_gen_dig_port_int_clk_wfm.py
+++ b/examples/digital_out/cont_gen_dig_port_int_clk_wfm.py
@@ -4,10 +4,11 @@ This example demonstrates how to output a continuous digital
 pattern using the DAQ device's clock.
 """
 
-import nidaqmx
 import numpy as np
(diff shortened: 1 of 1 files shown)


bkeryan:
@zhindes @mikeprosserni Both examples look like this:

https://github.com/user-attachments/assets/96e25bc1-392f-4cc7-a028-089da2b67b2b

PR #934, opened 2026-03-06
Introducing spell checking and fixing several typos

### LINKED ISSUE #937: Change cspell.json location
## Tech Debt
cspell.json is a non-dotfile in the root directory of the repo, so everyone who looks at the repo on GitHub will see in the directory listing.

Suggestion:

- Rename this file to .config/cspell.json
- Rename .cspell/* to .config/cspell/* so you don't add two dot-directories
- Add dictionaries to cspell
- Remove cspell from each of the individual dictionary names

Before doing this, explore how this may impact our ability to use real words as suggestions within VSCode.

[AB#3777366](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3777366)

KostaIlic2:
**I have not run tests locally**

I introduced configuration and dictionaries for cspell spell checker. I fixed several typos.

To eliminate typos present now and reduce probability of typos in the future.

I ran all static checks and generated documentation. And I ran the spell checker.
I skipped regression tests. I hope PR CI covers that.

### Resolved Issues

Closes #937

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       codegen-python-version: "3.9"
     steps:
       - name: Check out repo
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           submodules: true
       - name: Set up Python
--- a/.github/workflows/generate_docs.yml
+++ b/.github/workflows/generate_docs.yml
@@ -10,14 +10,14 @@ jobs:
       - ubuntu-latest
     steps:
       - name: Check out repo
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Set up Python
         uses: ni/python-actions/setup-python@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
         id: setup-python
       - name: Set up Poetry
         uses: ni/python-actions/setup-poetry@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
       - name: Cache virtualenvs
-        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: |
             .venv
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -43,7 +43,7 @@ jobs:
     needs: [check_nidaqmx, check_docs]
     steps:
       - name: Check out repo
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Set up Python
         uses: ni/python-actions/setup-python@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
       - name: Set up Poetry
@@ -54,7 +54,7 @@ jobs:
       - name: Build distribution packages
         run: poetry build
       - name: Upload build artifacts
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: ${{ env.dist-artifact-name }}
           path: dist/*
@@ -71,7 +71,7 @@ jobs:
       id-token: write
     steps:
     - name: Download build artifacts
-      uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+      uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
       with:
         name: ${{ env.dist-artifact-name }}
         path: dist/
@@ -89,7 +89,7 @@ jobs:
       pull-requests: write
     steps:
       - name: Check out repo
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Set up Python
         uses: ni/python-actions/setup-python@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
       - name: Set up Poetry
--- a/.github/workflows/report_test_results.yml
+++ b/.github/workflows/report_test_results.yml
@@ -14,9 +14,9 @@ jobs:
       pull-requests: write
     steps:
       - name: Check out repo
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Download test results
-        uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
         with:
           path: test_results
           pattern: test_results_*
@@ -24,6 +24,6 @@ jobs:
       - name: List downloaded files
         run: ls -lR
       - name: Publish test results
-        uses: EnricoMi/publish-unit-test-result-action@34d7c956a59aed1bfebf31df77b8de55db9bbaaf # v2.21.0
+        uses: EnricoMi/publish-unit-test-result-action@27d65e188ec43221b20d26de30f4892fad91df2f # v2.22.0
         with:
           files: "test_results/*.xml"
--- a/.github/workflows/run_system_tests.yml
+++ b/.github/workflows/run_system_tests.yml
@@ -20,7 +20,7 @@ jobs:
     timeout-minutes: 90
     steps:
       - name: Check out repo
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Import DAQmx config
         run: C:\nidaqmxconfig\targets\win64U\x64\msvc-14.0\release\nidaqmxconfig.exe --eraseconfig --import tests\max_config\nidaqmxMaxConfig.ini
       - name: Set up Python
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Poetry
         uses: ni/python-actions/setup-poetry@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
       - name: Cache virtualenvs
-        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: |
             .venv
@@ -49,7 +49,7 @@ jobs:
           done
         if: always() && runner.os != 'Windows'
       - name: Upload test results
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: test_results_system_${{ matrix.configuration }}
           path: test_results/*.xml
--- a/.github/workflows/run_unit_tests.yml
+++ b/.github/workflows/run_unit_tests.yml
@@ -21,7 +21,7 @@ jobs:
       pytest-opts: ${{ (startsWith(matrix.python-version, 'pypy') || matrix.python-version == '3.14t') && '-k "not grpc"' || '' }}
     steps:
       - name: Check out repo
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Set up Python ${{ matrix.python-version }}
         uses: ni/python-actions/setup-python@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
         id: setup-python
@@ -30,7 +30,7 @@ jobs:
       - name: Set up Poetry
         uses: ni/python-actions/setup-poetry@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
       - name: Cache virtualenv (with test dependencies)
-        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: .venv
           key: nidaqmx-with-test-${{ runner.os }}-py${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('poetry.lock') }}
@@ -39,7 +39,7 @@ jobs:
       - name: Run unit tests
         run: poetry run pytest -v --cov=generated/nidaqmx --junitxml=test_results/unit-${{ matrix.os }}-py${{ matrix.python-version }}.xml tests/unit ${{ env.pytest-opts }}
       - name: Upload test results
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: test_results_unit_${{ matrix.os }}_py${{ matrix.python-version }}
           path: test_results/*.xml
@@ -54,7 +54,7 @@ jobs:
         python-version: ["3.10", "3.13", "pypy3.11"]
     steps:
       - name: Check out repo
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Set up Python ${{ matrix.python-version }}
         uses: ni/python-actions/setup-python@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
         id: setup-python
@@ -63,7 +63,7 @@ jobs:
       - name: Set up Poetry
         uses: ni/python-actions/setup-poetry@aa64e60612cb078b0c2ada666becbd70d4817d55 # v0.7.1
       - name: Cache virtualenv (main only)
-        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         id: cache-nidaqmx-main-only
         with:
           path: .venv
--- a/examples/analog_in/calculated_power_acq_int_clk.py
+++ b/examples/analog_in/calculated_power_acq_int_clk.py
@@ -1,0 +1,28 @@
+﻿"""Example of analog input calculated power acquisition.
+
+This example demonstrates how to acquire a finite amount
+of calculated power data using the DAQ device's internal clock.
+
+Supported Devices:
+
+- PXIe-4311 (DAQmx 26.3.0 or later)
+"""
+
+import nidaqmx
+from nidaqmx.constants import READ_ALL_AVAILABLE, AcquisitionType
+
+with nidaqmx.Task() as task:
+    task.ai_channels.add_ai_calculated_power_chan(
+        voltage_physical_channel="Dev1/ai0",
+        current_physical_channel="Dev1/ai1",
+        voltage_min_val=0.0,
+        voltage_max_val=5.0,
+        current_min_val=0.0,
+        current_max_val=0.02,
+        ext_shunt_resistor_val=249.0,
+    )
+
+    task.timing.cfg_samp_clk_timing(1000.0, sample_mode=AcquisitionType.FINITE, samps_per_chan=50)
+
+    data = task.read(READ_ALL_AVAILABLE)
+    print("Acquired data: [" + ", ".join(f"{value:f}" for value in data) + "]")
--- a/examples/analog_in/cont_calculated_power_acq_int_clk.py
+++ b/examples/analog_in/cont_calculated_power_acq_int_clk.py
@@ -1,0 +1,39 @@
+﻿"""Example of analog input calculated power acquisition.
+
+This example demonstrates how to acquire a continuous amount of
+calculated power data using the DAQ device's internal clock.
+
+Supported Devices:
+
+- PXIe-4311 (DAQmx 26.3.0 or later)
+"""
+
+import nidaqmx
+from nidaqmx.constants import AcquisitionType
+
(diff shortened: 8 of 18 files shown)


KostaIlic2:
CONTRIBUTING.md will need to be updated to cover spell checking expectations.

.cspell/cspell-dict.txt
```diff
@@ -0,0 +1,168 @@
+# Real words
+nidaqmx
```
zhindes: heck ya nidaqmx is a word!

.github/workflows/build.yml
```diff
@@ -32,6 +32,8 @@ jobs:
           install-args: "--all-extras --with examples"
       - name: Run Bandit security checks
         run: poetry run bandit -c pyproject.toml -r generated/nidaqmx
+      - name: Run cspell
+        run: npx cspell "**" --no-progress --quiet
```
zhindes: Does this pipeline not need to install cspell or something? I'm not familiar with node. Also, I don't love using node for this. But... some light googling isn't giving me clear alternatives.

CONTRIBUTING.md
```diff
@@ -34,7 +34,8 @@ pass.
 conventions. If this reports errors, first run `poetry run ni-python-styleguide fix` in order to sort
 imports and format the code with Black, then manually fix any remaining errors.
 10. Run `poetry run mypy` to statically type-check the updated code.
-11. Send a GitHub Pull Request to the main repository's master branch. GitHub Pull Requests are the
+11. Run `npx cspell "**" --no-progress` to check for spelling errors. This requires [Node.js](https://nodejs.org/) to be installed.
```
zhindes, line 39: How bad is it if engineers just wait for the PR to do it for them? At least for a while people will stumble into it. Does the error reporting look good?

generated/nidaqmx/_linux_installation_commands.py
```diff
@@ -66,7 +66,7 @@ class DistroInfo:  # noqa: D101 - Missing docstring in public class (auto-genera
     install_commands: list[list[str]]
 
 
-# Mapping of distros to their command templates and version handlers
```
mshafer-NI, line 69: "distro" and "distros" are used often enough that I think we should consider them to be words

generated/nidaqmx/types.py
```diff
@@ -3,7 +3,7 @@
 import collections
 import typing
 
-# region Task Counter IO namedtuples
+# region Task Counter IO named tuples
```
mshafer-NI, line 6: `namedtuple` is a type word, I think we should allow `namedtuple` and `namedtuples`

.github/workflows/build.yml
bkeryan: There are well-known problems with configuring build workflows to use the latest version of a package:
- Build reproducibility: the latest version of the package may introduce incompatible behavior or bugs.
- Security: the latest version of the package or one of its dependencies may be compromised. This is a problem for any package registry but it seems like an even bigger problem for Npm.js.

We try to use these techniques to mitigate these problems:
- Pinning: specifying a version like `cspell@9.7.0` would prevent us from uptaking new cspell versions until we're ready, but it would still use the latest compatible versions of cspell's dependencies
- Locking: generating a `package-lock.json`, checking it into version control, and using `npm ci` to install cspell should ensure that we use known versions of cspell and its dependencies
- Automatic updates: we use Renovate to update our Poetry lock files and pinned GitHub Actions
- Dependency cooldowns: we have Renovate configured to use minimumReleaseAge=14 days, so that we don't update to the latest version of a dependency until at least 2 weeks after it has been released. If an upstream package is compromised, this significantly increases the likelihood that this will be detected before we update to it.

bkeryan: npx is like pipx. It downloads the latest version of the package that has been published to npm.js and runs it.

We are already using Node.js in many of our Python repos to run Pyright.
- There is a Python wrapper package: https://pypi.org/project/pyright/
- Apparently the wrapper uses the Python `nodeenv` package to create a Node.js virtual environment: https://pypi.org/project/nodeenv/
- Then it installs a pinned version of Pyright: https://github.com/RobertCraigie/pyright-python/blob/main/src/pyright/_utils.py#L69
- The pinned version is hardcoded in the Python package: https://github.com/RobertCraigie/pyright-python/blob/main/src/pyright/_version.py

I guess pinning without locking is "Not great, not terrible."

.cspell/cspell-dict.txt
```diff
@@ -0,0 +1,168 @@
...
+rtol
+
+# nitypes
+nitypes
+
+# hightime
+hightime
+
+# nptdms
+bnptdms
```
bkeryan: `bnptdms` is not a word. It's `\b` concatenated with `nptdms`. Why are we spell checking regexps?

https://github.com/ni/nidaqmx-python/blob/ac17727a34db3d34705d40283093c51aa6674086/tests/acceptance/test_examples.py#L42

.cspell/cspell-dict.txt
```diff
@@ -0,0 +1,168 @@
...
+# pytest
+pytest
+addoption
+callspec
+caplog
+fixturenames
+getfixturevalue
+metafunc
+modifyitems
+parametrizing
```
bkeryan: This one (parametrizing) hurts because I accidentally spell it correctly and then my test doesn't work.

.cspell/cspell-dict.txt
```diff
@@ -0,0 +1,168 @@
...
+# nidaqmx C library / header filenames (external)
+nicaiu
+nicai
+
+# NI-DAQmx C API identifier fragments (surface in tests/examples but defined in excluded metadata)
+outp
+lvls
+nums
+CAPI
+descr
```
bkeryan: lvl, num, and descr are in the DAQmx abbreviation list, which is used for DAQmx attribute names.

outp is not. It's only used in the C API, which I guess does not enforce use of the abbreviation list.

.cspell/cspell-dict.txt
bkeryan, line 1: Please alphabetize each group in order to reduce the cognitive overhead of maintaining this file and prevent merge conflicts caused by people adding new words to the end.

.config/cspell.json
bkeryan, line 1: This is a non-dotfile in the root directory of the repo, so everyone who looks at the repo on GitHub will see in the directory listing.

How about:
- Rename this file to `.config/cspell.json`
- Rename `.cspell/*` to `.config/cspell/*` so you don't add two dot-directories

I am assuming that the `.cspell/` directory is not handled specially by cspell. (The docs say that `.config/` is handled specially.)

cspell.json
```diff
@@ -0,0 +1,55 @@
...
+    "docs/img/**",
+    "docs/conf.py",
+    "docs/make.bat",
+    "docs/Makefile",
+    "generated/**",
+    "src/codegen/functions/**",
+    "src/codegen/metadata/**",
+    "src/codegen/protos/**",
+    "src/codegen/properties/**",
+    "src/codegen/*.py",
```
bkeryan: Why exclude so many specific directories in src/codegen? Why not exclude the whole thing? Spell-checking generated/** will catch any errors in the codegen templates that are visible to users.

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
```diff
@@ -13,14 +13,14 @@
     # or samps_per_chan_written=)
     samps_per_chan_param = get_samps_per_chan_read_or_write_param(function.base_parameters)
 %>\
-        cfunc = lib_importer.${'windll' if function.calling_convention == 'StdCall' else 'cdll'}.DAQmx${function.c_function_name}
-        if cfunc.argtypes is None:
-            with cfunc.arglock:
-                if cfunc.argtypes is None:
-                    cfunc.argtypes = [
+        c_func = lib_importer.${'windll' if function.calling_convention == 'StdCall' else 'cdll'}.DAQmx${function.c_function_name}
```
bkeryan, line 16: Please revert cfunc->c_func and arglock->arg_lock. I think they introduce a lot of churn without improving clarity. 

PEP 8 naming conventions say "with words separated by underscores as necessary to improve readability". IMO "cfunc" and "arglock" are sufficiently readable.
https://peps.python.org/pep-0008/#function-and-variable-names

Also, you got by without renaming cfunc in _lib.py.

src/handwritten/_dotenv_path.py
bkeryan, line 1: https://peps.python.org/pep-0008/#package-and-module-names

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. 

IMO "_dotenvpath.py" is sufficiently readable. 

Also, "dot" and "env" are separate words. The fact that there is a deprecated package we are not using called https://pypi.org/project/dotenv/ is a result of the Python naming conventions saying that separating words with underscores is optional.

.github/workflows/build.yml
bkeryan: The authors of cspell publish a GitHub Action for running it: https://cspell.org/docs/Configuration/githubaction

Let's use cspell-action in our GitHub Actions workflow instead of npx. Renovate will keep the action up to date. I assume the action installs the dependency versions listed in its lock file.

The example in CONTRIBUTING.md should still use npx.

bkeryan: Actually, the situation with Pyright is probably fine. In GitHub Actions workflows we use https://github.com/jakebailey/pyright-action/tree/main to run Pyright. 

We pass it the Pyright version from our pyproject.toml, but the action downloads & caches the tarball: https://github.com/jakebailey/pyright-action/blob/main/src/helpers.ts#L234

Pyright has only 1 non-dev dependency, and it's optional: https://www.npmjs.com/package/pyright?activeTab=code

It looks like all of Pyright's other dependencies are bundled into the tarball, so specifying a Pyright version should be sufficient for reproducible builds (assuming it doesn't install the optional fsevents dependency).

.cspell/cspell-dict.txt
KostaIlic2: Thank you for catching that.

We are spell checking regexps because the spell checker can't tell regex from other strings on its own.

I see that I added a word to the dictionary incorrectly. There are several alternative approaches. I will give them some thought.

.cspell/cspell-dict.txt
KostaIlic2: Thank you for bringing this up.

The word is not in built-in dictionaries. The form you pointed out came from the code. I added it to the dictionary because I believe it is OK. For example, refer to [https://en.wikipedia.org/wiki/Parametrization_(geometry)](https://en.wikipedia.org/wiki/Parametrization_(geometry))

What would you like me to do here?

Some options that come to mind are:

1. Add the 'e' to the dictionary, then change the code.
2. Add parameterizing to the dictionary so if someone spells it differently they don't get an error
3. On top of the prior option, introduce two more with 's' instead of 'z'.

We can consider capturing decisions about these types of questions in a separate document.

.cspell/cspell-dict.txt
KostaIlic2: I don't see `descr` in the list. Do you think `descr` should be added to that list?

While `lvl` and `num` are in the list, plural forms are not. Since they are used in the API, I would feel better adding them to the list of abbreviations, next to the singular forms, than having them in a separate file.

I don't know what to do about `outp`.  One thought is to grandfather it into the list.

Thoughts?

CONTRIBUTING.md
KostaIlic2: I can see a workflow where engineers wait on the PR to do everything for them. Is there a reason to treat spell checking differently from other static analysis chores?

I think that reporting of "unknown words" is reasonable:
image

generated/nidaqmx/_linux_installation_commands.py
KostaIlic2: `distros` is used in exactly one location in this body of code.

I did not add `distro` to the list because the singular form is already present in [the software terms built-in dictionary](https://raw.githubusercontent.com/streetsidesoftware/cspell-dicts/main/dictionaries/software-terms/dict/softwareTerms.txt).

I believe that the authors of that dictionary have given thought to plural form of such industry specific abbreviations.

Do you think that the readers of the comment will be confused by the fully spelled out word?

This seems like another candidate for a document about spelling.

generated/nidaqmx/types.py
KostaIlic2: `namedtuples` is used in exactly one location in this body of code.

`namedtuple` is in [cspell built-in python dictionary](https://raw.githubusercontent.com/streetsidesoftware/cspell-dicts/main/dictionaries/python/dict/python.txt), and `namedtuples` is not.

I personally find the two real words with a space in between in the comment easier to understand. Do you think that this spelling will confuse any readers of the comment?

I prefer not to add plural form to our dictionary.

Objections?

.cspell/cspell-dict.txt
KostaIlic2: After some analysis, I realized that we could remove `bnptdms` from the dictionary. We already have `nptdms`, and cspell treats `\b` as a delimiter. `bnptdms` is gone!

.cspell/cspell-dict.txt
KostaIlic2: I am leaning toward 1, based on what I get when I ask MML which of the two spellings is more common.

.cspell/cspell-dict.txt
KostaIlic2: I hope that I addressed the feedback with a different change I made. I alphabetized the whole file and added comments to each of the entries. Given the number of groups, what I did works better for me. Let me know if you have any objections.

KostaIlic2: If you like this, we can invent a linter for dictionaries; but that would be outside the scope of this PR.

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
KostaIlic2: We have a balancing act between adding content to our dictionary, using other ways to pass spell check, and changing the code. I can address cfunc with a minor cspell configuration change. I lean toward including arglock to the dictionary if we decide to change it back, but I don't particularly like that.

The way I see it, the churn is cosmetic and low-risk.

I would like for us to consider this along with distros and namedtuples mentioned elsewhere in this PR.

src/handwritten/_dotenv_path.py
KostaIlic2: Interesting that the wording in PEP-8 is slightlty different for module names and variable/function names.

Given the wording for module names, I can argue that adding the underscore improves readability. The guide does not make inclusion of underscores conditional on insufficient readability without the underscores.

.cspell/cspell-dict.txt
KostaIlic2: After more research, I concluded that the word is based on pytest [parametrize (sic)](https://docs.pytest.org/en/stable/how-to/parametrize.html) concept. I am leaving the comment and the dictionary entry just the way they are.

.cspell/cspell-dict.txt
KostaIlic2: @bkeryan, Based on our live conversation, I will look into consolidating all DAQmx API elements in one dictionary.

.config/cspell.json
KostaIlic2: I appreciate the value of this idea. I would like to take this up as a separate PR. I would like to learn more about cspell integration into VSCode to provide suggestions, and make sure that this change doesn't impact our ability to do such a thing.

cspell.json
KostaIlic2: I reduced the number of excluded directories a lot.

I am still excluding  generated/nidaqmx/_stubs/**" because those are protobuf-generated stubs with many issues, and the file is not intended for human consumption.

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
KostaIlic2: Based on our earlier conversation, I renamed cfunc in _lib.py for consistency.

.github/workflows/build.yml
KostaIlic2: @bkeryan, I think I made the change you requested. I don't know how to test it. Please review my chnage and either resolve this conversation or let me know what else I need to do.

.cspell/cspell-dict.txt
KostaIlic2: @bkeryan, I consolidated all DAQmx API terminology in one dictionary.

.config/cspell.json
KostaIlic2: I created #937 to address this later.

src/handwritten/_dotenv_path.py
KostaIlic2: I created #938 to remember to rename the file in a separate PR.

.config/cspell.json
KostaIlic2: I got a chance to complete the research I had in mind. I implemented the suggestion as part of this PR.

If this PR is completed, we'll be able to close #937 without additional work.

src/codegen/metadata/attributes.py
bkeryan, line 1: This metadata comes from one of our internal Git repos. I will post a PR to fix it there so this doesn't get reverted the next time someone updates this file based on the internal metadata.

src/codegen/templates/library_interpreter/default_c_function_call.py.mako
bkeryan: I dealt with the churn by checking out your branch locally, changing the names back, and seeing what was left in the diff: just ptr -> pointer. I guess that's ok.

```
--- C:\Users\bkeryan\AppData\Local\Temp\git-blob-a28532\_library_interpreter.py	2026-03-12 20:48:32.000000000 -0500
+++ C:\Dev\nidaqmx-python\generated\nidaqmx\_library_interpreter.py	2026-03-12 20:48:25.000000000 -0500
@@ -6596,26 +6596,26 @@
                         ctypes.c_uint,
                         ctypes.c_uint,
                         ctypes.POINTER(ctypes.c_int),
                         ctypes.POINTER(c_bool32),
                     ]
 
-        read_array_ptrs = (ctypes.POINTER(ctypes.c_double) * channel_count)()
+        read_array_pointers = (ctypes.POINTER(ctypes.c_double) * channel_count)()
         for i, read_array in enumerate(read_arrays):
-            read_array_ptrs[i] = read_array.ctypes.data_as(ctypes.POINTER(ctypes.c_double))
+            read_array_pointers[i] = read_array.ctypes.data_as(ctypes.POINTER(ctypes.c_double))
 
         error_code = cfunc(
             task_handle,
             num_samps_per_chan,
             timeout,
             t0_array,
             dt_array,
             0 if t0_array is None else t0_array.size,
             self._get_wfm_attr_callback(properties),
             None,
-            read_array_ptrs,
+            read_array_pointers,
             channel_count,
             array_size,
             ctypes.byref(samps_per_chan_read),
             None,
         )
         self.check_for_error(error_code, samps_per_chan_read=samps_per_chan_read.value)
@@ -7096,22 +7096,22 @@
                         ctypes.POINTER(ctypes.POINTER(ctypes.c_double)),
                         ctypes.c_uint,
                         ctypes.POINTER(ctypes.c_int),
                         ctypes.POINTER(c_bool32),
                     ]
 
-        write_array_ptrs = (ctypes.POINTER(ctypes.c_double) * channel_count)()
+        write_array_pointers = (ctypes.POINTER(ctypes.c_double) * channel_count)()
         for i, write_array in enumerate(write_arrays):
-            write_array_ptrs[i] = write_array.ctypes.data_as(ctypes.POINTER(ctypes.c_double))
+            write_array_pointers[i] = write_array.ctypes.data_as(ctypes.POINTER(ctypes.c_double))
 
         error_code = cfunc(
             task_handle,
             num_samps_per_chan,
             auto_start,
             timeout,
-            write_array_ptrs,
+            write_array_pointers,
             channel_count,
             ctypes.byref(samps_per_chan_written),
             None,
         )
 
         return error_code, samps_per_chan_written.value
```

.github/workflows/build.yml
bkeryan: It ran successfully.

https://github.com/ni/nidaqmx-python/actions/runs/22983930545/job/66729633778?pr=934

```
Run streetsidesoftware/cspell-action@v8
  with:
    incremental_files_only: false
    check_dot_files: explicit
    inline: warning
    treat_flagged_words_as_errors: false
    strict: true
    suggestions: false
    verbose: false
    use_cspell_files: false
    summary: false
  env:
    codegen-python-version: 3.9
    pythonLocation: /opt/hostedtoolcache/Python/3.9.25/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.9.25/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.25/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.25/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.25/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.25/x64/lib
    pythonVersion: 3.9.25
    POETRY_BIN_DIR: /home/runner/work/_temp/poetry/bin
    POETRY_CONFIG_DIR: /home/runner/work/_temp/poetry/config
    POETRY_HOME: /home/runner/work/_temp/poetry/home
    POETRY_HOME_BIN: /home/runner/work/_temp/poetry/home/bin
cspell-action
Pull Request
Files checked: 428, Issues found: 0 in 0 files.
Done.
```

bkeryan: Installing cspell is now handled by cspell-action. Closing.

.config/cspell/daqmx-api-elements.txt
bkeryan, line 1: > This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).
> I have not run tests locally
> TODO: Check the above box with an 'x' indicating you've read and followed [CONTRIBUTING.md](https://github.com/ni/nidaqmx-python/blob/master/CONTRIBUTING.md).

That's okay. Everything passes in the PR workflow, except for "Run system tests (win-10-py32)" which is an intermittent issue with the self-hosted runner used for that job.

> I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.
> TODO: Check the above box with an 'x' if considered if there were any and then documented client facing changes in [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md). Strike through if this is not a relevant client facing change.

Please add a bullet point saying the project now uses cspell for spell checking.

We forgot to add the 1.5.0 section after releasing 1.4.0, so please add it to the top:
```
## 1.5.0
* ### Merged Pull Requests
    * [Full changelog: 1.4.0...1.5.0](https://github.com/ni/nidaqmx-python/compare/1.5.0...1.5.0)

* ### Resolved Issues
    * ...

* ### Major Changes
    * ...

* ### Known Issues
    * ...
```

> I've added tests applicable for this pull request
>TODO: Check the above box with an 'x' indicating you have considered and added any applicable system or unit tests. Strike through if you considered and decided additional tests were not warranted.

Strike through this, e.g. `- [ ] ~~blah blah~~`

CONTRIBUTING.md
bkeryan: FYI, Zach is OOO

generated/nidaqmx/types.py
bkeryan: @mshafer-NI Here's what I think:

In prose, it is acceptable to convert class names to prose. For example, "Returns a GrpcEventHandler" -> "Returns a gRPC event handler".

It is also acceptable to use the class name as a modifier for another noun (kind of like a trademark). For example, "Returns a GrpcEventHandler" -> "Returns a GrpcEventHandler object".

Either of these techniques can be used to avoid pluralizing a class name. For example, "Returns a collection of GrpcEventHandlers" -> "Returns a collection of gRPC event handlers" or "Returns a collection of GrpcEventHandler objects".

Also, I thought I converted all of these `namedtuple` classes to use `typing.NamedTuple`. 😅

.config/cspell/daqmx-api-elements.txt
bkeryan: > We forgot to add the 1.5.0 section after releasing 1.4.0, so please add it to the top:

Correction: this was done a month ago. Please rebase/merge against the latest master branch.

.config/cspell.json
KostaIlic2: I managed to complete the research I had in mind today, so I implemented the requested change as part of this PR.

.cspell/cspell-dict.txt
KostaIlic2: Marking resolved since there were no requests to change anything.

.github/workflows/build.yml
KostaIlic2: The issue is addressed.

.cspell/cspell-dict.txt
KostaIlic2: Addressed.

.config/cspell/daqmx-api-elements.txt
KostaIlic2: Done.

PR #963, opened 2026-04-22
Update changelog and pyproject for 1.6.0-dev0

TanYiQing:
~~- [ ] I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~

- Update version of master branch to 1.6.0-dev0.
- Add section for 1.6.0 in changelog.

Allow development for next version.

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Changelog
+* [1.6.0](#160)
 * [1.5.0](#150)
 * [1.4.1](#141)
 * [1.4.0](#140)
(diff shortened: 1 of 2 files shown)


zhindes (approved):

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Changelog
+* [1.6.0](#160)
```
zhindes, line 2: Actually, is there any reason we need to bump `MINOR` versions in the near future? I don't think we're doing any feature development. I think we should move to `1.5.1`, then `1.5.2`, for a bit.

zhindes (changes requested):
I think we probably want `1.5.1`

CHANGELOG.md
bkeryan: We should release 1.5.1 out of releases/1.5. IMO, main should still target 1.6.

bkeryan (approved):

CHANGELOG.md
TanYiQing: @zhindes Yes, I share the same view as Brad. Do you have any concerns about this?

zhindes: Nope, sound good.

zhindes (approved):

PR #979, opened 2026-06-01
Update ni-python-styleguide (2.1.0) and mypy (0.5.0)

zhindes:
- [x] ~~I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~
- [x] ~~I've added tests applicable for this pull request~~

Two commits (I'll rebase):
1. update to mypy 2.1.0 (from 1.20.2)
2. update to ni-python-styleguide 0.5.0 (from 0.4.9)

Breaking version updates.

Ran in my setup:
```sh
> poetry run ni-python-styleguide lint
> poetry run mypy
```


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.4.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.4 and should not be changed by hand.
```
bkeryan: Renovate uses this comment to know which version of Poetry to run, so downgrading it will break minimumReleaseAge because that feature was added in Poetry 2.4.0.

bkeryan (changes requested):

poetry.lock
zhindes: Fixed!
* updated poetry
* did a rebase
* re-ran `poetry update ni-python-styleguide` after my first commit
* re-ran `poetry update mypy` after the second
* squashed everything back down to 2 commits

bkeryan (approved):

PR #1000, opened 2026-08-03
add clear task method and made close an alias

zhindes:

I had some customer feedback that `clear` task was more familiar for people coming from other DAQmx APIs, which I agree with. In practice, tasks should almost always be used with Python context managers, but in the rare cases where they aren't, this method could be useful.

Simple rename of an existing method with an alias to the old name to not break compatibility.

Added a unit test, and ran all the existing ones. griffe says:

```
>poetry run griffe check -s generated\ -a 1.5.0 --verbose nidaqmx
generated\nidaqmx\task\_task.py:364: Task.close:
Public object points to a different kind of object:
  Old: function
  New: attribute
```

Their [documentation](https://mkdocstrings.github.io/griffe/guide/users/checking/#object-changed-kind) says:

> Changing the kind of an object is not really an API breakage, depending on our definition of API, since this won't always raise immediate errors like TypeError. The object is still here and accessed: the contract is fulfilled. But developers sometimes rely on the kind of an object, so changing it will lead to incorrect behavior, potentially making it difficult to detect, understand and fix the issue. That is why it is important to warn developers about such changes.

### DIFF
--- a/generated/nidaqmx/task/_task.py
+++ b/generated/nidaqmx/task/_task.py
@@ -322,7 +322,7 @@ class Task:
 
         self._interpreter.add_global_chans_to_task(self._handle, channels)
 
-    def close(self):
+    def clear(self):
         """Clears the task.
 
         Before clearing, this method aborts the task, if necessary,
@@ -360,6 +360,12 @@ class Task:
 
         if first_exception:
             raise first_exception
+
+    close = clear
+    """Clears the task.
+
+    :meth:`close` is an alias for :meth:`clear`.
+    """
 
     def control(self, action):
         """Alters the state of a task according to the action you specify.
--- a/tests/unit/test_task.py
+++ b/tests/unit/test_task.py
@@ -111,6 +111,20 @@ def test___call_alternate_constructor___create_task_not_called(
 
 
 @pytest.mark.parametrize("close_on_exit", [False, True])
+def test___varying_close_on_exit___clear___clear_task_called(
+    interpreter: Mock, close_on_exit: bool
+):
+    expect_create_task(interpreter, "MyTaskHandle", close_on_exit)
+    expect_get_task_name(interpreter, "MyTask")
+    task = Task("MyTask")
+    task_handle = task._handle
+
+    task.clear()
+
+    interpreter.clear_task.assert_called_with(task_handle)
+
+
+@pytest.mark.parametrize("close_on_exit", [False, True])
 def test___varying_close_on_exit___close___clear_task_called(
     interpreter: Mock, close_on_exit: bool
 ):
@@ -119,6 +133,7 @@ def test___varying_close_on_exit___close___clear_task_called(
     task = Task("MyTask")
     task_handle = task._handle
 
+    # close is an alias for clear
     task.close()
 
     interpreter.clear_task.assert_called_with(task_handle)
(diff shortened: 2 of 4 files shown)


generated/nidaqmx/task/_task.py
```diff
@@ -361,6 +361,19 @@ def close(self):
         if first_exception:
             raise first_exception
 
+    def clear(self):
```
bkeryan: You should be able to alias methods by writing

```
clear = close
```

I think Sphinx will display it as an alias.

Perhaps clear should be the canonical name and close should be the alias?

Use https://pypi.org/project/griffe/ to verify that this doesn't break compatibility. Install w/pipx. Here's the command line: https://github.com/ni/measurement-plugin-python/pull/1228#discussion_r2356591728
> griffe check -s packages\service ni_measurement_plugin_sdk_service -a 2.3.1 --verbose

bkeryan (changes requested):

generated/nidaqmx/task/_task.py
zhindes: griffe says this:

```
>poetry run griffe check -s generated\ -a 1.5.0 --verbose nidaqmx
generated\nidaqmx\task\_task.py:364: Task.close:
Public object points to a different kind of object:
  Old: function
  New: attribute
```

Their [documentation](https://mkdocstrings.github.io/griffe/guide/users/checking/#object-changed-kind) says:

> Changing the kind of an object is not really an API breakage, depending on our definition of API, since this won't always raise immediate errors like TypeError. The object is still here and accessed: the contract is fulfilled. But developers sometimes rely on the kind of an object, so changing it will lead to incorrect behavior, potentially making it difficult to detect, understand and fix the issue. That is why it is important to warn developers about such changes.

Documentation isn't explicit, but the lack of a source link is a hint, I guess.

image

Overall, I think this is much cleaner and good!

tests/unit/test_task.py
```diff
@@ -124,6 +124,21 @@ def test___varying_close_on_exit___close___clear_task_called(
...
+@pytest.mark.parametrize("close_on_exit", [False, True])
+def test___varying_close_on_exit___clear___clear_task_called(
+    interpreter: Mock, close_on_exit: bool
+):
+    expect_create_task(interpreter, "MyTaskHandle", close_on_exit)
+    expect_get_task_name(interpreter, "MyTask")
+    task = Task("MyTask")
+    task_handle = task._handle
+
+    # clear is an alias for close
```
zhindes: woops, need to move this comment. i'm going to keep all the other tests as-is, though.

generated/nidaqmx/task/_task.py
bkeryan: Ok, I guess Sphinx doesn't do this yet: https://github.com/sphinx-doc/sphinx/issues/13910

generated/nidaqmx/task/_task.py
```diff
@@ -361,6 +361,8 @@ def close(self):
         if first_exception:
             raise first_exception
 
+    close = clear
```
bkeryan, line 364: Try this:

```suggestion
    close = clear
    """Clears the task.

    :meth:`close` is an alias for :meth:`clear`.
    """
```

zhindes: Beautiful

image

bkeryan (approved):

Issue #917, opened 2026-01-14
Add gRPC examples


bkeryan:
## Problem to Solve

There are no example programs in this repo demonstrating how to use the `nidaqmx` package with gRPC.

The only example of `nidaqmx` with gRPC is in another repo:
- https://github.com/ni/measurement-plugin-python/tree/main/examples/nidaqmx_analog_input

## Proposed Solution

Add a small number of gRPC examples (1-3).

[AB#3709351](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3709351)

PR #1019, opened 2026-09-10
Enable the creation of mTLS encrypted gRPC channels leveraging NI TLS via nitlsconfig (Python)

alexdubois-ni:

- [X] I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.

- [X] I've added tests applicable for this pull request

Note: I've only added unit tests as of this time. We intend to add system tests in a follow-up PR.

We also intend to add documentation in a follow-up PR.

Leverage and integrate nitlsconfig (from the nitlsconfig-python repository). This will give us the capability to:
- Use nitlsconfig for integration with NI TLS workflows from Hardware Manager to produce secure gRPC channels.
- Perform integration to add other parts of nitlsconfig's Python support such as client side connection auditing and improved error elaborations to aid customers.

Without this integration, customers would need to know about and add nitlsconfig[grpc] themselves as a python package to install. This will add a unified platform method for mTLS encryption with gRPC.

Furthermore, we can also leverage our package and integrate it to add CRA required audit logging capabilities and also help us provide customers with additional setup debugging information telling them they need to go to Hardware Manager.

Lastly, this is all provided from a shared public github repo that we can modify to reiterate and share our code and improvements with nimi-python.

NOTE: nitlsconfig 1.0.0a4 is our most recent alpha version. We will publish a formal 1.0.0 after we've completed all of our testing. This will automatically be pulled in for customers with our existing >= 1.0.0a4. But, we will want to go back and update this to 1.0.0 most likely, especially if owners here would desire this :)

Unit tests have been added and succeeded locally. System tests are soon to follow.

We have integrated this very identically in nimi-python for all of our gRPC Python API supporting drivers and it has been unit tested and manually tested there as well (ATS tests pending in that repo, soon to follow for here too).

### DIFF
--- a/.config/cspell/project-software-terms.txt
+++ b/.config/cspell/project-software-terms.txt
@@ -37,6 +37,7 @@ NISHAREDDIR       # NI Linux install directory environment variable
 nicai             # NI-DAQmx C library filename (Windows DLL)
 nicaiu            # NI-DAQmx C library filename (Windows DLL, Unicode)
 nidevice          # NI package name; ni-device-grpc
+nitlsconfig       # Dependency: nitlsconfig
 nitypes           # Dependency: nitypes
 nptdms            # Dependency: nptdms
 numpy             # Dependency: numpy
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ All notable changes to this project will be documented in this file.
 
 * ### Major Changes
     * Add support for Analog Path Delay
-    * Added `nitlsconfig[grpc]>=1.0.0a4` to the optional gRPC dependencies.
+    * Added `nitlsconfig[grpc]` to the `grpc` extra to enable using NI TLS to create TLS-encrypted gRPC channels.
 
 * ### Known Issues
     * ...
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ All notable changes to this project will be documented in this file.
 
 * ### Major Changes
     * Add support for Analog Path Delay
+    * Added `nitlsconfig[grpc]>=1.0.0a4` to the optional gRPC dependencies.
 
 * ### Known Issues
     * ...
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ All notable changes to this project will be documented in this file.
 
 * ### Major Changes
     * Add support for Analog Path Delay
-    * Added `nitlsconfig[grpc]` to the `grpc` extra to enable using NI TLS to create TLS-encrypted gRPC channels.
+    * Added `nitlsconfig[grpc]` to the `grpc` extra to enable using NI TLS to create TLS-encrypted gRPC channels. `nitlsconfig` is the Python API for reading the NI TLS configuration installed on the system and creating NI gRPC Device client channels from it.
 
 * ### Known Issues
     * ...
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -115,6 +115,7 @@ class GrpcStubInterpreter(BaseInterpreter):
             minor_version = 0
             update_version = 0
         finally:
+            # Record the outcome of a driver's gRPC session initialize RPC.
             nitlsconfig.audit_session_connect('NI-DAQmx', grpc_options.grpc_channel, connected)
         self._driver_version = DriverVersion(major_version, minor_version, update_version)
 
--- a/generated/nidaqmx/_grpc_interpreter.py
+++ b/generated/nidaqmx/_grpc_interpreter.py
@@ -12,6 +12,7 @@ from collections.abc import Callable, Sequence
 
 import google.protobuf.message
 import grpc
+import nitlsconfig
 import numpy
 
 from . import errors as errors
@@ -101,14 +102,20 @@ class GrpcStubInterpreter(BaseInterpreter):
     def __init__(self, grpc_options):
         self._grpc_options = grpc_options
         self._client = nidaqmx_grpc.NiDAQmxStub(grpc_options.grpc_channel)
+        # Querying the driver version is the first contact with the server, so it decides
+        # whether this interpreter connected.
+        connected = False
         try:
             major_version = self.get_system_info_attribute_uint32(0x1272)
             minor_version = self.get_system_info_attribute_uint32(0x1923)
             update_version = self.get_system_info_attribute_uint32(0x2f22)
+            connected = True
         except Exception:
             major_version = 0
             minor_version = 0
             update_version = 0
+        finally:
+            nitlsconfig.audit_session_connect('NI-DAQmx', grpc_options.grpc_channel, connected)
         self._driver_version = DriverVersion(major_version, minor_version, update_version)
 
     def _invoke(self, func, request, metadata=None):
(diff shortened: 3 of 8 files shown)


generated/nidaqmx/_grpc_interpreter.py
bkeryan, line 1: This project uses CSpell. Add "nitlsconfig" to https://github.com/ni/nidaqmx-python/blob/master/.config/cspell/project-software-terms.txt

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ All notable changes to this project will be documented in this file.
 
 * ### Major Changes
     * Add support for Analog Path Delay
+    * Added `nitlsconfig[grpc]>=1.0.0a4` to the optional gRPC dependencies.
```
bkeryan: Please include a brief description of what nitlsconfig is and why nidaqmx-python is using it.

generated/nidaqmx/_grpc_interpreter.py
```diff
@@ -2427,7 +2438,7 @@ def read_analog_f64(
                 task=task, num_samps_per_chan=num_samps_per_chan,
                 timeout=timeout, fill_mode_raw=fill_mode,
                 array_size_in_samps=read_array.size))
-    
+
```
bkeryan: Why did this trailing whitespace change? I don't see any whitespace changes in _grpc_interpreter.py.mako that would cause this.

tests/unit/test_nitlsconfig_grpc_integration.py
```diff
@@ -0,0 +1,133 @@
...
+
+        def trailing_metadata(self):
+            return []
+
+except ImportError:
+    grpc = None  # type: ignore
+
+
+@pytest.fixture
+def nitls_tagged_channel():
```
bkeryan: CSpell is also flagging "nitls". Consider spelling out "nitlsconfig" or abbreviating further to "tls".

generated/nidaqmx/_grpc_interpreter.py
bkeryan, line 1: Will you be adding an example program? If so, that would also address https://github.com/ni/nidaqmx-python/issues/917

bkeryan, line 1: Almost all of the tests in https://github.com/ni/nidaqmx-python/tree/master/tests/component run over grpc, so we do have coverage of this code with real gRPC channels, but not tagged for nitlsconfig. Do you intend to add any coverage of real gRPC channels tagged for nitlsconfig, or should the new unit tests and nitlsconfig's own tests be enough?

bkeryan (changes requested):

generated/nidaqmx/_grpc_interpreter.py
bkeryan: nvm, I didn't read the full PR description

alexdubois-ni: Unfortunately, we will not. We intended to extend them for nitlsconfig if they existed already, but they were purposefully left out during the initial gRPC implementation it would seem and this is beyond the scope of what our team is trying to accomplish.

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -118,14 +119,20 @@ class GrpcStubInterpreter(BaseInterpreter):
...
             major_version = self.get_system_info_attribute_uint32(0x1272)
             minor_version = self.get_system_info_attribute_uint32(0x1923)
             update_version = self.get_system_info_attribute_uint32(0x2f22)
+            connected = True
         except Exception:
             major_version = 0
             minor_version = 0
             update_version = 0
+        finally:
+            nitlsconfig.audit_session_connect('NI-DAQmx', grpc_options.grpc_channel, connected)
```
zhindes, line 136: What does this do? It looks like the first server connection will raise if it fails, so what is an audit?

alexdubois-ni: As part of the CRA, we want to provide auditing information which we will be placing in windows event viewer or syslog that provides in this case:

1. Did the connection succeed or fail?
2. What is the endpoint you successfully or failed to connect to (hostname:port)

We have auditing that also exists for what the security transport expectations are as well such as:
1. Is this a proper mTLS connection? Or is it insecure?

This will allow customers to check servers and clients for suspicious or malicious behavior is the goal.

generated/nidaqmx/_grpc_interpreter.py
alexdubois-ni: I think I accidentally saved in VS code which trimmed them all. Added it back!

src/codegen/templates/_grpc_interpreter.py.mako
zhindes: Ah neat! I think a comment here would be good since its one-level removed. The one-liner from that method's docstring is nice.

```python
# Record the outcome of a driver's gRPC session initialize RPC.
```

src/codegen/templates/_grpc_interpreter.py.mako
```diff
@@ -118,14 +119,20 @@ class GrpcStubInterpreter(BaseInterpreter):
...
         self._grpc_options = grpc_options
         self._client = nidaqmx_grpc.NiDAQmxStub(grpc_options.grpc_channel)
+        # Querying the driver version is the first contact with the server, so it decides
+        # whether this interpreter connected.
+        connected = False
         try:
             major_version = self.get_system_info_attribute_uint32(0x1272)
             minor_version = self.get_system_info_attribute_uint32(0x1923)
             update_version = self.get_system_info_attribute_uint32(0x2f22)
+            connected = True
```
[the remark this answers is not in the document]
alexdubois-ni: If the first succeeds, I would expect the rest to succeed and vice versa. This would fail a session initialize and it's fine to overall fail the connection otherwise as well. It may not be 100% accurate, but 99% accurate is fine here.

zhindes (approved):
nothing more than Brad's comments!

.config/cspell/project-software-terms.txt
bkeryan, line 1: Does this require any new software to be installed for existing non-TLS use cases? I see that the nitlsconfig Python package shells out to a new `nitlsconfig` command: https://github.com/ni/nitlsconfig-python/blob/main/src/nitlsconfig/cli.py

I saw the comment about "create_grpc_device_channel needs the nitlsconfig CLI installed on the system." I just want to check that the auditing part works without the nitlsconfig CLI. Otherwise, this will break existing applications with older drivers that install the latest `nidaqmx` Python package.

bkeryan: I see the system tests passed, which is a good sign, but I'm not sure which version of NI-DAQmx we're using for testing or whether it includes the `nitlsconfig` command.

alexdubois-ni: nitlsconfig.audit_session_connect and nitlsconfig.get_tls_error_elaboration are both independent of the nitlsconfig CLI itself.

nitlsconfig itself as one of our platform libraries does not provide auditing capabilities. However, CRA does require that we provide auditing capabilities and we did not wish to provide duplicated code between all drivers in nimi-python and nidaqmx-python, so we instead opted to share the code in nitlsconfig's python API itself since all users of nitlsconfig will more than likely wish to integrate with the auditing capabilities.

Lastly, for the elaborations, that's a simple improvement we can share and provide with consistent error messaging for customers between our python driver APIs and since it's also NI TLS related, we decided to share it in nitlsconfig as well.

Should've made that clear in the description and understandable concern!

bkeryan (approved):

PR #1022, opened 2026-09-14
Add documentation for creating nitlsconfig gRPC channels

alexdubois-ni:

~~- [ ] I've updated [CHANGELOG.md](https://github.com/ni/nidaqmx-python/blob/master/CHANGELOG.md) if applicable.~~

Provides documentation for utilizing nitlsconfig with nidaqmx-python.

Since we're not opting to add examples, we should at the very least provide public documentation that highlights how to use nitlsconfig to produce mTLS encrypted channels using NI TLS, and walk customers through some of the required setup.

This matches the existing documentation that was added to nimi-python. However, the language did need to change because standard NI driver APIs ONLY have a session object with gRPC options, whereas NI-DAQmx has multitudes. I produced a short summary aligned with changelog mentions to at least make it clearer here.

N/A. This is documentation only. Full integration is done in this PR and ATS tests will soon follow: https://github.com/ni/nidaqmx-python/pull/1019


docs/grpc_session_options.rst
```diff
@@ -7,6 +7,56 @@ Support for using NI-DAQmx over gRPC
...
+:py:class:`nidaqmx.GrpcSessionOptions`. The constructors for :py:class:`nidaqmx.Task`,
+:py:class:`nidaqmx.Scale`, and other classes accept a ``grpc_options`` parameter, and
+:py:meth:`nidaqmx.system.System.remote` accepts one to access the remote DAQmx system. You own the
+channel, not the objects created from it, so you must close the gRPC channel only after every
+NI-DAQmx gRPC object using it is closed.
+
+The recommended way to create a gRPC channel to a remote system running NI gRPC Device Server is
+:py:func:`nitlsconfig.create_grpc_device_channel() <nitlsconfig.grpc_channel.create_grpc_device_channel>`
+from the `nitlsconfig <https://nitlsconfig-python.readthedocs.io/en/latest/>`_ package, which the
+``grpc`` extra installs for you. It reads the nitlsconfig client configuration installed with the
```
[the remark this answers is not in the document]
alexdubois-ni: Change is being made in the corresponding PR attached which will make this valid. Resolving.

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,56 @@ Support for using NI-DAQmx over gRPC
...
+NI-DAQmx runtime and by default will attempt to build an encrypted gRPC channel using mTLS.
+
+Before ``create_grpc_device_channel`` can succeed, you must use NI Hardware Manager to perform a
+certificate exchange with the remote system.
```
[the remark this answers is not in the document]
alexdubois-ni: TLS is enabled by default and I want this to be clear. This is the language we're already using in nimi-python.

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,56 @@ Support for using NI-DAQmx over gRPC
...
+    arguments such as ``grpc.ssl_target_name_override``, and a ``retry_policy`` parameter. Channel
+    arguments cannot be changed after the channel is built, so they must be supplied here.
+
+.. note:: NI gRPC Device Server must be configured to accept remote connections and to take its
+    TLS settings from nitlsconfig. See
+    `Bind Address Support <https://github.com/ni/grpc-device#bind-address-support>`_ and
+    `NI TLS Config Integration <https://github.com/ni/grpc-device#ni-tls-config-integration>`_ for details.
+
+You can also build the gRPC channel yourself with ``grpc.insecure_channel`` or ``grpc.secure_channel``
+if you need full control over how credentials are supplied.
```
bkeryan: There is an existing API for connecting to NI gRPC Device Server from a measurement service running on the same machine:

https://nimeasurementlinksessionmanagementclient.readthedocs.io/en/latest/autoapi/ni/measurementlink/sessionmanagement/v1/client/index.html#ni.measurementlink.sessionmanagement.v1.client.BaseReservation.create_nidaqmx_task

There is also a low-level helper object that we use to cache gRPC channel objects:

https://ni-grpc-extensions.readthedocs.io/en/latest/autoapi/ni_grpc_extensions/channelpool/index.html

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,56 @@ Support for using NI-DAQmx over gRPC
...
+  $ python -m pip install nidaqmx[grpc]
+
+Every NI-DAQmx gRPC object is created from a ``grpc.Channel`` that you build and pass to
+:py:class:`nidaqmx.GrpcSessionOptions`. The constructors for :py:class:`nidaqmx.Task`,
+:py:class:`nidaqmx.Scale`, and other classes accept a ``grpc_options`` parameter, and
+:py:meth:`nidaqmx.system.System.remote` accepts one to access the remote DAQmx system. You own the
+channel, not the objects created from it, so you must close the gRPC channel only after every
+NI-DAQmx gRPC object using it is closed.
+
+The recommended way to create a gRPC channel to a remote system running NI gRPC Device Server is
```
bkeryan: This only covers the remote system case. I think we should have separate sub-sections for remote systems with nitlsconfig and the local system with measurement services. This also goes for nimi-python.

bkeryan (changes requested):

docs/grpc_session_options.rst
bkeryan: BTW, in case you're unfamiliar:
- https://www.ni.com/docs/en-US/bundle/measurementplugins/page/measurement-plugins.html
- https://www.ni.com/docs/en-US/bundle/measurementplugins/page/session-manager-src.html
- https://github.com/ni/measurement-plugin-python/blob/main/examples/nidaqmx_analog_input/measurement.py

zhindes (approved):
Brad's on it; I trust y'all

docs/grpc_session_options.rst
alexdubois-ni: Created a new remote system and local system set of sections that tries to dive into more detail for considerations and recommendations for each. I will note to you upfront that NI Hardware manager doesn't really provide a good streamlined way to do local host connections with mTLS. This is why I make it clear that the initial recommendation is just a grpc.insecure_channel. But, I do know it is possible. Alex Hearn has at least mentioned it to me, and there are rough mentions on the mTLS management page I linked.

docs/grpc_session_options.rst
alexdubois-ni: I did try to make some mention of measurement plug-ins in the local system section I added. I will note that I am entirely unfamiliar with measurement plugins themselves and would definitely like for you to vet and make recommendations for what would be useful and not useful information in that section. I tried to keep it succinct :)

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,89 @@ Support for using NI-DAQmx over gRPC
...
+    `NI TLS Config Integration <https://github.com/ni/grpc-device#ni-tls-config-integration>`_ for details.
+
+If you are writing a
+`measurement plug-in <https://www.ni.com/docs/en-US/bundle/measurementplugins/page/measurement-plugins.html>`_,
+you do not create the channel at all. The
+`session manager <https://www.ni.com/docs/en-US/bundle/measurementplugins/page/session-manager-src.html>`_
+creates it for you and hands you a :py:class:`nidaqmx.Task`, so you do not create
+:py:class:`nidaqmx.GrpcSessionOptions` yourself. For working measurements that use NI-DAQmx this
+way, see the
+`measurement plug-in examples <https://github.com/ni/measurement-plugin-python/tree/main/examples>`_.
```
bkeryan:
```suggestion
`NI-DAQmx measurement plug-in example <https://github.com/ni/measurement-plugin-python/tree/main/examples/nidaqmx_analog_input>`_.
```

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,89 @@ Support for using NI-DAQmx over gRPC
...
+    nitlsconfig. If it is not configured this way, do not use ``create_grpc_device_channel`` for
+    the local system. See
+    `NI TLS Config Integration <https://github.com/ni/grpc-device#ni-tls-config-integration>`_ for details.
+
+If you are writing a
+`measurement plug-in <https://www.ni.com/docs/en-US/bundle/measurementplugins/page/measurement-plugins.html>`_,
+you do not create the channel at all. The
+`session manager <https://www.ni.com/docs/en-US/bundle/measurementplugins/page/session-manager-src.html>`_
+creates it for you and hands you a :py:class:`nidaqmx.Task`, so you do not create
+:py:class:`nidaqmx.GrpcSessionOptions` yourself. For working measurements that use NI-DAQmx this
```
bkeryan:
```suggestion
`session management service <https://www.ni.com/docs/en-US/bundle/measurementplugins/page/session-manager-src.html>`_ tracks the lifetimes of NI-DAQmx tasks on the NI gRPC Device Server and the `session management client <https://nimeasurementlinksessionmanagementclient.readthedocs.io/en/latest/autoapi/ni/measurementlink/sessionmanagement/v1/client/index.html#ni.measurementlink.sessionmanagement.v1.client.BaseReservation.create_nidaqmx_task>`_ creates a :py:class:`nidaqmx.Task` for you, so you do not create
:py:class:`nidaqmx.GrpcSessionOptions` yourself. For working measurements that use NI-DAQmx this
```

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,89 @@ Support for using NI-DAQmx over gRPC
...
+
+.. note:: NI gRPC Device Server must be configured to accept remote connections and to take its
+    TLS settings from nitlsconfig. See
+    `Bind Address Support <https://github.com/ni/grpc-device#bind-address-support>`_ and
+    `NI TLS Config Integration <https://github.com/ni/grpc-device#ni-tls-config-integration>`_ for details.
+
+The local system
+~~~~~~~~~~~~~~~~
+
+For a simple local system setup, build the channel yourself with ``grpc.insecure_channel``.
```
bkeryan: Let Intersphinx link to the gRPC docs.
```suggestion
For a simple local system setup, build the channel yourself with :py:func:`grpc.insecure_channel`.
```

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,89 @@ Support for using NI-DAQmx over gRPC
...
 .. py:currentmodule:: nidaqmx
 
+Creating a gRPC channel
+-----------------------
+
+Using NI-DAQmx over gRPC requires the ``grpc`` extra::
+
+  $ python -m pip install nidaqmx[grpc]
+
+Every NI-DAQmx gRPC object is created from a ``grpc.Channel`` that you build and pass to
```
bkeryan:
```suggestion
Every NI-DAQmx gRPC object is created from a :py:class:`grpc.Channel` that you build and pass to
```

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,89 @@ Support for using NI-DAQmx over gRPC
...
+:py:class:`nidaqmx.GrpcSessionOptions`. The constructors for :py:class:`nidaqmx.Task`,
+:py:class:`nidaqmx.Scale`, and other classes accept a ``grpc_options`` parameter, and
+:py:meth:`nidaqmx.system.System.remote` accepts one to access the remote DAQmx system. You own the
+channel, not the objects created from it, so you must close the gRPC channel only after every
+NI-DAQmx gRPC object using it is closed.
+
+The recommended way to create the channel depends on where NI gRPC Device Server runs. The sections
+below cover a remote system and the local system. In either case you can instead build the channel
+yourself, with ``grpc.insecure_channel`` for an insecure channel or ``grpc.secure_channel`` when you
```
bkeryan:
```suggestion
yourself, with :py:class:`grpc.insecure_channel` for an insecure channel or :py:class:`grpc.secure_channel` when you
```

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,89 @@ Support for using NI-DAQmx over gRPC
...
+need full control over how credentials are supplied.
+
+Remote systems
+~~~~~~~~~~~~~~
+
+For a remote system, the recommended way is
+:py:func:`nitlsconfig.create_grpc_device_channel() <nitlsconfig.grpc_channel.create_grpc_device_channel>`
```
bkeryan, line 33: Is it necessary to repeat the function name?
```suggestion
:py:func:`nitlsconfig.create_grpc_device_channel`
```

docs/grpc_session_options.rst
```diff
@@ -7,6 +7,89 @@ Support for using NI-DAQmx over gRPC
...
 
+Creating a gRPC channel
+-----------------------
+
+Using NI-DAQmx over gRPC requires the ``grpc`` extra::
+
+  $ python -m pip install nidaqmx[grpc]
+
+Every NI-DAQmx gRPC object is created from a ``grpc.Channel`` that you build and pass to
+:py:class:`nidaqmx.GrpcSessionOptions`. The constructors for :py:class:`nidaqmx.Task`,
```
bkeryan: Intersphinx is not linking to nidaqmx.Task and nidaqmx.Scale because those are aliases that Sphinx doesn't know about.

Easy fix: Use nidaqmx.task.Task and nidaqmx.scale.Scale here. 

Less easy fix: change `nidaqmx/__init__.py` to define Task and Scale as type aliases rather than imports,
e.g.
```
import nidaqmx.task

Task: TypeAlias = nidaqmx.task.Task
```

This is less easy because it could have unintended side effects and involves typing_extensions.

bkeryan (changes requested):

docs/grpc_session_options.rst
alexdubois-ni: In this case, yes.

It renders as just nitlsconfig.create_grpc_device_channel() still, and we intentionally name it twice so the link text shows the public name while the target points at the grpc_channel submodule where it's actually documented.

We can't link nitlsconfig.create_grpc_device_channel directly because nitlsconfig disables imported-members in its Sphinx config, so re-exported names aren't in its inventory. The top-level name is also resolved lazily via __getattr__ (only available with the [grpc] extra), so it isn't statically visible to autoapi either.

docs/grpc_session_options.rst
bkeryan: ok, specifying a different display name from the link seems like a good solution to this one too.

alexdubois-ni: I opted for the easy fix as you recommended, but followed a bit of what we did with nitlsconfig where we provide the expected usage instead of the full name from the submodule. Bit more verbose and odd in the rst file itself, but should yield a better customer experience!

bkeryan (approved):

